PR #3, opened 2017-01-04
LUAUnit and Toolchain documentation

salacryl:


salacryl:
mh I got lualint  only as plugin, cause it only calls luac -p. Other linters are not up to date

PR #17, opened 2017-01-10
talents addon_english

Yahnich:
to make it easier to change them in the future


chrisinajar:
We've since discovered techniques that work better for solving this problem

PR #22, opened 2017-01-12
"Infinite" levels

SquawkyArctangent:
Includes decreased stat gain for levels over 25, increasing the maximum level to 5000 ( might be able to go higher, but not by much ), and replacing barebone's dumb modified experience table for DOTA's default one, including a simple formula for extending it past level 25.


chrisinajar:
Can you set it to 10,000?

SquawkyArctangent:
Nope. Crashes at some level between 5001 and 6900.

PR #23, opened 2017-01-12
gold.lua + chatcommand.lua

Zarnotox:
gold.lua is a working version of AABS infinite gold system with some
bits added in to make it work
Also added in information to completely integrate infinite gold to AAA

with chatcommand.lua you can really easily add some commands in the game
from literally anywhere, also because it's dynamic it doesn't break
those commands when using script_reload
An example can be seen tester.lua


Zarnotox:
Updated chatcommand.lua to dynamically call functions in other classes too, gonna push that version

PR #24, opened 2017-01-12
gold.lua + chatcommand.lua

Zarnotox:
gold.lua is a working version of AABS infinite gold system with some
bits added in to make it work
Also added in information to completely integrate infinite gold to AAA

with chatcommand.lua you can really easily add some commands in the game
from literally anywhere, also because it's dynamic it doesn't break
these commands when using script_reload
An example can be seen in tester.lua


chrisinajar:
Wouldn't the AAR style infinite gold not require rewriting things like buying items? We can still overwrite the UI to show the right value -- what is the value in not using the built in gold system?

I really like chatcommands.lua, you should break that out into a separate pull request.

Zarnotox:
Okay I'll close this one, chatcommand.lua is even better now.

AAR won't require you to change the complete system.
2 small problems: 
- you can't buy items that cost more than 99999 gold
- if we limit the 'real' gold to 80000 you can't get more than 19999 gold within one tick

chrisinajar:
We can fix those special cases and leave the rest to work normally, as opposed to starting by replacing everything.

PR #27, opened 2017-01-12
Adjusted stat formula and set MAX_LEVEL to 6892

SquawkyArctangent:
Anything higher actually crashes the game on loading.


chrisinajar:
Change xp needed to `level * ((level - 25 -32084) / -32084) * 108` so that level 10000 is equal to level 6892. set the cap to 10000.

If this doesn't work then it's a problem of maximum total xp as opposed to max xp per level, i can come up with something else for that one.

SquawkyArctangent:
That change only allows a maximum of 6898, actually.

chrisinajar:
@SquawkyArctangent fix conflicts and this is good to go, we'll figure out what to do about 6k+ levels when someone decides that matters (hint: It doesn't)

PR #38, opened 2017-01-21
Adjusted stat formula past 25

SquawkyArctangent:
This truly is something to be prideful of, this truly complicated change.

### DIFF
--- a/game/scripts/vscripts/events.lua
+++ b/game/scripts/vscripts/events.lua
@@ -145,7 +145,7 @@ function GameMode:OnPlayerLevelUp(keys)
   
   -- reduce the stat gain past a certain level
   if level > 25 then
-    local div = level / 20
+    local div = div = (level - 25 + 12)/12
 	
 	-- get the hero's stat gain
 	local gainStr = hero:GetStrengthGain()
--- a/game/scripts/vscripts/events.lua
+++ b/game/scripts/vscripts/events.lua
@@ -145,7 +145,7 @@ function GameMode:OnPlayerLevelUp(keys)
   
   -- reduce the stat gain past a certain level
   if level > 25 then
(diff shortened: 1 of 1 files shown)


chrisinajar:
Monumental.

game/scripts/vscripts/events.lua
```diff
@@ -145,7 +145,7 @@ function GameMode:OnPlayerLevelUp(keys)
   
   -- reduce the stat gain past a certain level
   if level > 25 then
-    local div = level / 20
+    local div = (level - 25 + 12)/12
```
zelding: -25+12? why not just subract 13? :)

chrisinajar: so that the `n` value is more apparent. modern compilers precalculate math using constants so it ends up executing the same exact thing, just with slightly clearer source code.

zelding:
I don't know the base stat gain for the heroes, but after a while the divider will be way over 3. If you are using relatively sane values (assuming around 3) then the stat gain will be little over nothing after lvl 50 (give or take).

See plot [here](https://www.wolframalpha.com/input/?i=plot+(x-13)%2F12+x%3D%7B25,+100%7D)

If you do a simple logarithmic normalization then the values become more manageable
example [here](https://www.wolframalpha.com/input/?i=plot+ln((x-13)%2F12)+x%3D%7B25,+100%7D)  

just ideas ;) keep up the good work guys!

game/scripts/vscripts/events.lua
zelding: I didn't thought about that, cool ;)

Haganeko:
The idea is that total stats will behave logarithmically, enabling us to have infinite maximum level while preventing carry overflow. Stat gains go to 0 by design.

PR #39, opened 2017-01-21
More Links

Chronophylos:
~also added the secret link~


chrisinajar:
Discord link is still an easter egg, don't put it in readme. Also you messed up the second one.

I like the new contributor link though!

chrisinajar:
Remove the links entirely until it's public. linking not to the thing you're linking is weird.

PR #46, opened 2017-01-23
enable universal shop

thepacman98:


Chronophylos:
this is exactly what I did in #44 too.

PR #49, opened 2017-01-29
Adding new map with detailing

superadude:
Worked on this and fleshed out all the details of the map. Boss areas blank so we can flesh them out with the bosses


superadude:
This pull has map version 9.1

chrisinajar:
baumi has since modified and uploaded this map, it's in the game now. thanks a ton!

PR #50, opened 2017-01-29
Patch Infinite gold

Chronophylos:
I added the custom Gold API from Angel Angel Arena Blackstar as suggested in #9 by @Zarnotox.
We now have two Gold API at the same time:
* the vanilla Gold API called with `PlayerResource:Fnc()`
* the custom API called with `Gold:Fnc()`

To make the custom API a compatible as possible with the vanilla one I added a loop in `game/scripts/vscripts/gamemode.lua` to check for vanilla gold remove it and add it to the custom gold.
The HUD is updated automatically when the Gold API is used.
It can be formatted in differently by changing a variable in `game/scripts/vscripts/gold.lua`

# Usage
Pretty much the same as vanilla, all the functions are the same.
But I added a global containing player data like Angel Arena Blackstar did. This variable just holds all the gold values.

# ToDo
Every function that accesses the vanilla API

# Misc
I fixed a error in `game/scripts/custom_net_tables.txt` by removing a ';' at the end of the line
ignore https://github.com/AngelArenaAllstars/aaa/pull/50/commits/9134553eb0b41cec6ad1e49dd72e5b349bbeb016


game/scripts/vscripts/globals.lua
```diff
@@ -0,0 +1,13 @@
+-- I don't like this..
```
chrisinajar: I don't like this either. Make it a local inside of `gold.lua` since that's the only place it's ever used.

If another part of code needs to access this data it should be an API instead of accessing the data directly.

game/scripts/vscripts/internal/util.lua
```diff
@@ -106,8 +106,26 @@ function DebugAllCalls()
     end
 end
 
-
-
+--[[
+  Credits:
+    Angel Arena Blackstar
+  Description:
+    I don't really know what this does
```
chrisinajar: It looks like it returns the player id from a given unit / player / table. For example, you should be able to pass in a reference to a lycan wolf and get back the correct player's ID.

@Yahnich any insight on what this code does?

content/panorama/scripts/custom_game/gold.js
```diff
@@ -0,0 +1,27 @@
...
+    GoldLabel.text = GameUI.FormatGold(data.gold[playerID]);
+  } else if (useFormatting == "half") {
+    GoldLabel.text = GameUI.FormatComma(data.gold[playerID]);
+  } else {
+    GoldLabel.text = data.gold[playerID];
+  }
+}
+
+(function () {
+    GameEvents.Subscribe("aaa_update_gold", UpdateGoldHud);
```
chrisinajar: Why not use the net table update event? `PlayerTables.SubscribeNetTableListener`

game/scripts/vscripts/gamemode.lua
```diff
@@ -129,6 +132,9 @@ function GameMode:OnGameInProgress()
   -- initialize modules
   InitModule(PointsManager)
   InitModule(CreepCamps)
+
+  -- start think timer
+  Timers:CreateTimer(0, Dynamic_Wrap(GameMode, "GameModeGoldThink"))
```
chrisinajar: You should create an initialization method inside of gold.lua and put both this and the playertable definition in there. The table should also be named something like "gold" since there's nothing aaa specific about it.

game/scripts/vscripts/gamemode.lua
```diff
@@ -151,6 +162,29 @@ function GameMode:InitGameMode()
...
 
+--[[
+  Author:
+    Chronophylos
+  Credits:
+    Angel Arena Blackstar
+  Description:
+    Add Gold to all players via our custom Gold API
+]]
+function GameMode:GameModeGoldThink()
```
chrisinajar: This also belongs in gold.lua

PR #56, opened 2017-01-30
Added minimap boundaries

Rachol:
Added 4 boundaries to the map. These boundaries are used to generate the
minimap


chrisinajar:
Hammer does this automatically now, and you also only need 2 of them.

Rachol:
Hammer doesn't do it on my computer though.... and I got a brand new setup today

PR #60, opened 2017-02-07
Added Creeps Upgrade Logic

Rachol:
1. When Camp reaches maximum amount of creeps, creeps will be upgraded.
Health, Mana, Armor, Damage, Gold and Exp Bounty are evenly distributed
towards existing creeps.
2. Added creep power level control. It is possible to define creep
levels and change the creep levels when spawning. Currently there is a
small timer doing that. Current CreepPowerTable is just for testing and
values need to be adjusted for actual needs.

### DIFF
--- a/game/scripts/vscripts/components/creeps/spawner.lua
+++ b/game/scripts/vscripts/components/creeps/spawner.lua
@@ -5,28 +5,51 @@ if CreepCamps == nil then
     CreepCamps = class({})
 end
 
+--creep power level is from CREEP_POWER_LEVEL_MIN to CREEP_POWER_LEVEL_MAX
+local CreepPowerLevel = 1.0 
+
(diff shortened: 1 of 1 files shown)


Rachol:
Btw. I have the Mana included in creep upgrade logic, but it currently does nothing... since SetMana() doesn't modify the maximum mana for the creep. If someone knows how to set maximum mana for the unit, let me know.

On the other hand, creeps are not using mana in Angel Arenas, so perhaps, we could just remove the mana part?

Chronophylos:
I think you can change the `MODIFIER_PROPERTY_MANA_BONUS` modifier. See https://developer.valvesoftware.com/wiki/Dota_2_Workshop_Tools/Lua_Abilities_and_Modifiers

PR #67, opened 2017-02-11
Rework gold

Chronophylos:
Removed the global variable `PLAYER_GOLD`. Gold now uses only `PlayerTables`.
Added `game/panorama/images` to .gitignore
Fixed error spewing in `gold.js`


chrisinajar:
Great work!

PR #69, opened 2017-02-13
Added item drop system

Rachol:
Item drop system:
1. Possibility to set percentage rate for dropping items.
2. Possible to define for each item:
	*when the drop is enabled
	*when the drop stops
	*how rare drop it is comparing to other items.


Rachol:
Any issues with this or are you guys just not having time to review it?

Rachol:
??

chrisinajar:
Sorry for the delay, there was a bit of debate on if we'd include an item drop system or not, but last I recall we settled on *at least* minor consumables being dropped.

PR #74, opened 2017-02-19
Update npc_items_custom.txt

RamonNZ:
I made this Staquila about 2 weeks ago. Also includes an Aquila which can have stats changed from the default. First time trying github... so might mess this up. Needs my fix for addon_english to display correctly.


chrisinajar:
Thank you!

PR #75, opened 2017-02-19
added minimap to aaa map

zelding:
Also

Added netutral spawners so they will show icons on the minimap
I've set them to -1 number of times so they wont spawn anything


zelding:
closing, since the map had numerous updates since than

PR #80, opened 2017-02-21
Better Debugging Proposal

Chronophylos:
This is suppose to be a example for better debugging.
Reject it if you don't like it otherwise I'll do this to the rest.
I'm using Gold as an Example.

I added Wrapper Functions for Gold:
```Lua
function Gold:DebugPrint ( args )
  if Gold.debuggingEnabled then
    DebugPrint( args )
  end
end

function Gold:DebugPrintTable ( args )
  if Gold.debuggingEnabled then
    DebugPrintTable( args )
  end
end
```
The Debugging Functions `DebugPrint()` and `DebugPrintTable()` get only called when `Gold.debuggingEnabled` is `true`. It's meant to make debugging easy by just changing a single variable and not commenting and uncommenting various parts of code.

### DIFF
--- a/game/scripts/vscripts/components/gold/gold.lua
+++ b/game/scripts/vscripts/components/gold/gold.lua
@@ -8,38 +8,46 @@
 
 
 if Gold == nil then
-  DebugPrint ( '[gold/gold] creating new Gold object' )
+  DebugPrint( '[gold/gold] Creating new Gold object.' )
   _G.Gold = class({})
 end
 
-local GOLD_CAP = 50000
 
-function Gold:Init()
+function Gold:Init ( )
+  DebugPrint( '[gold/gold] Initializing..' )
+  Gold.GoldCap = 50000
   -- a table for every player
-  PlayerTables:CreateTable("gold", {
-    gold = {}
-  }, {0,1,2,3,4,5,6,7,8,9})
+  PlayerTables:CreateTable( 'gold',
+                            { gold = {} },
+                            {0,1,2,3,4,5,6,7,8,9} )
+
+  Gold.debuggingEnabled = false
 
     -- start think timer
-  Timers:CreateTimer(0, Dynamic_Wrap(Gold, "Think"))
+  Timers:CreateTimer( 0, Dynamic_Wrap( Gold, 'Think' ) )
 end
 
-function Gold:UpdatePlayerGold(unitvar, newGold)
-  local playerID = UnitVarToPlayerID(unitvar)
+function Gold:DebugPrint ( args )
+  if Gold.debuggingEnabled then
+    DebugPrint( args )
+  end
+end
+
+function Gold:DebugPrintTable ( args )
+  if Gold.debuggingEnabled then
+    DebugPrintTable( args )
+  end
+end
+
+function Gold:UpdatePlayerGold ( unitvar, newGold )
+  local playerID = UnitVarToPlayerID( unitvar )
+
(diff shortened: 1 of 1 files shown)


chrisinajar:
I like the idea, but I don't like having to put those lines into every file.

What about a central component factory,
```lua
function Components:CreateComponent (name)
  local component = class({})

  function printWithPrefix (msg)
    if component.debuggingEnabled then
      DebugPrint ('[' .. name .. '] ' .. msg)
    end
  end
  function printTableWithPrefix (msg)
    if component.debuggingEnabled then
      DebugPrint ('[' .. name .. '] ')
      DebugPrintTable(msg)
    end
  end

  -- i miss function hoisting
  component.DebugPrint = printWithPrefix
  component.DebugPrintTable = printTableWithPrefix

  return component
end
```

then where you'd normally do `Gold = class({})` you instead do `Gold = Components:CreateComponent("gold/gold")` and all your messages are auto-prefixed too!

yeahbuddy-dota:
This is my implementation of the code discussed in coding-help

cheats.lua (my cheats.lua, not aaa's)
~~~

function Cheats:EnableDebug(debugFile)
  local is = Cheats.Debug[debugFile]
  --can't be ~is because ~(nil) will throw an error
  if is then
    Cheats.Debug[debugFile] = false
  else
    Cheats.Debug[debugFile] = true
  end
  if debugFile == "all" then
    print("All debuggers " .. (Cheats.Debug[debugFile] and "activated" or "dormant (unless activated previously)"))
  else
    print("Debugging for file " .. debugFile .. (Cheats.Debug[debugFile] and " enabled" or " disabled"))
  end
end
~~~
util.lua
~~~

function GetCallingFile()
	local str = debug.traceback()
	function lines(str)
	  local t = {}
	  local function helper(line) table.insert(t, line) return "" end
	  helper((str:gsub("(.-)\r?\n", helper)))
	  return t
	end

	local linez = lines(str)
	local line = linez[4]
    local regex = string.match(line, "[/\\]([^\\/]+).lua:")

	return regex
end


function GetCallingFileWithDirectory()
	local str = debug.traceback()
	function lines(str)
	  local t = {}
	  local function helper(line) table.insert(t, line) return "" end
	  helper((str:gsub("(.-)\r?\n", helper)))
	  return t
	end

	local linez = lines(str)
	local line = linez[4]
    local regex , lineNo= string.match(line, "scripts[/\\]vscripts[/\\](.+).lua:([0-9]+):")
    
    if lineNo then
		return regex .. ":" .. lineNo
    else
    	return regex
    end
end

function DebugPrint(...)
	local trace = GetCallingFile()
	local dir = GetCallingFileWithDirectory()

	if Cheats then
		if Cheats.Debug[trace] or Cheats.Debug.all then
			if type(...) == "table" then
				print(dir,unpack(...))
			else
				print("[" .. dir .. "]",...)
			end
		end
	end

end

~~~
then when you call the console command `debug gamemode` (or `debug all`), stuff like the following will spew
~~~
[   VScript              ]: [barebones\internal\gamemode:54]	[BAREBONES] GameRules set
[   VScript              ]: [barebones\internal\gamemode:121]	[BAREBONES] Done loading Barebones gamemode!
~~~

chrisinajar:
I think if we clean up yeahbuddy's code and name things better that seems like a great way to do it. those debug API's are wonderful.

chrisinajar:
I used this basic technique but implemented it a little different, merged in with #98

PR #98, opened 2017-03-02
Add new debugging system with automatic reverse compatibility

chrisinajar:
Based on https://github.com/AngelArenaAllstars/aaa/pull/80

### DIFF
--- a/game/scripts/vscripts/internal/util.lua
+++ b/game/scripts/vscripts/internal/util.lua
@@ -1,3 +1,12 @@
+
+Debug = Debug or {
+  EnabledModules = {
+    ['internal:*'] = true,
+    ['gamemode:*'] = true
+  },
+  EnableAll = false
+}
+
 function split(s, delimiter)
     result = {};
     for match in (s..delimiter):gmatch("(.-)"..delimiter) do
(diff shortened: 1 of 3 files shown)


game/scripts/vscripts/internal/util.lua
```diff
@@ -6,14 +15,88 @@ function split(s, delimiter)
...
+        table.insert(result, match);
+    end
+    return result;
+end
+
+function TracesFromFilename (filename)
+  local traces = {}
+  local i = 1
+
+  local parts = regexsplit(filename, '%s/\\')
```
Chronophylos: This is not Cross-platform compatible

game/scripts/vscripts/internal/util.lua
```diff
@@ -6,14 +15,88 @@ function split(s, delimiter)
...
+
+-- written by yeahbuddy, taken from https://github.com/AngelArenaAllstars/aaa/pull/80
+-- modified for clarity
+function GetCallingFile (offset)
+  offset = offset or 4
+
+  local str = debug.traceback()
+  local lines = split(str, '\n')
+  local line = lines[offset]
+  local dirName , lineNo= string.match(line, "scripts[/\\]vscripts[/\\](.+).lua:([0-9]+):")
```
Chronophylos: Again this is (if I'm not missing something) not Crossplatform compatible.

game/scripts/vscripts/internal/util.lua
chrisinajar: I'm splitting on `%s`, `/`, and `\` here. The regex split method takes in the string and puts it inside a `[^ delim ]` segment, so here it'll grab whitespace or either platform's slash.

game/scripts/vscripts/internal/util.lua
chrisinajar: Also splitting on `/` and `\`.

`[/\\]` is a character group of / and \, so it'll grab either.

chrisinajar:
Patterns are confusing because they look like regex's but don't always perform like them. In the code here I use `[/\\]` a lot, which will match either platform's slash since `\` isn't the escape character, `%` is.

Chronophylos:
Sorry I forgot how to regex. :/

PR #99, opened 2017-03-02
Gonna start collecting and publishing /docs

warpdragon:
Docs collected so far.


chrisinajar:
You reference the "Moar-docs" branch... Would you like to merge it into a similarly named branch on this repo, or simply consider this a work in progress for the time being?

warpdragon:
what is the advantage of pulling it into its own branch on angelarenaallstars/oaa?

Chronophylos:
We could centralize the work to this docs branch on AngelArenaAllstars/oaa instead of the branch on your warpdragon/moar-docs.

warpdragon:
yeah i wasn't sure about that myself. i really wanted to put it on angelarenaallstars but i didn't know if it would be good to ask people to send PRs to someone else without their explicit permission (even though i know for a fact that chris wants documentation). So I guess it's ok and I really don't think people are going to send me anything (you're the only one so far. I guess we should merge it with chris. I'll update the documentation to reflect the new procedure.

Issue #90, opened 2017-02-26
Need Ability-UI for 6/5 lvl spells


warpdragon:
# Need Ability-UI for Six-Level abilities
By default, I believe, Dota only properly displays the ability-level indicator dots (you know the ones) beneath Hero-Abilities in *odd-numbers*. Currently if you were to, say, put a six-level hero ability in AAA, it would only show the first 5 indicator dots (the sixth being hidden/invisible/just not there at all). vOLvo Pls

Need a LUA-scripting/PNG-overlay solution for displaying an even number (6) of dots beneath a spell.

spar36:
I have dont know how to do this, but could you just display the next highest odd number, so 7, and then cover up the last dot.

yeahbuddy-dota:
Aren't all normal DotA skills 4 levels.. they all show up fine

warpdragon:
No ur right it does show 4 dots fine. But nevertheless it doesnt work when u give em 6

Trildar:
This can actually be fixed with just some CSS since the UI does actually insert classes to indicate abilities of arbitrary levels. The problem is that doing it that way would require overriding one of the built-in CSS files, meaning that stuff may break if they ever change that file.

It's probably also possible to add the required style by JavaScript, which may be more resistant to breaking from changes depending on how it's done, but it's also harder.

Also, just to note, the dots have class LevelPanel and setting the left and right margins to 1px seems to produce good results.

Trildar:
On further investigation, the JavaScript method might not be at all practical as it causes lag. If anyone can think of a way that doesn't involve looping through all the dots every time the selected unit is changed, but otherwise I'll probably put in a PR to add the CSS file tomorrow.

PR #100, opened 2017-03-02
Fix for 6 level abilities in HUD #90

Trildar:
* Add dota_hud_ability_panel.css
  Lines 1676-1679 are the important bit. The rest is extracted from vanilla Dota.

Note: If Valve decide to change this file (in vanilla Dota) for whatever reason, it will likely need to be updated with the changes.

### DIFF
--- a/content/panorama/styles/hud/dota_hud_ability_panel.css
+++ b/content/panorama/styles/hud/dota_hud_ability_panel.css
@@ -1675,7 +1675,8 @@ DOTAAbilityPanel.Hidden
 /* Added for AAA*/
 .Reborn .AbilityMaxLevel6 .LevelPanel
 {
-  margin: 3px 1px 3px 1px;
+  width: 7px;
+  margin: 3px 1.5px 3px 1.5px;
 }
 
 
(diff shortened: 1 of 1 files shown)


Chronophylos:
Do we really need the whole file?

Trildar:
Pretty sure it completely overrides rather than adding. Tried just putting the additional part and the UI gets messed up. You're welcome to double check though.

chrisinajar:
CSS files are scoped to the dom which includes them, so to override system level dom styling you need to entirely replace the CSS files.......... UNLESS you edit the styling from JavaScript, in which case you can overwrite any system level styling and it works!

If we edit the styles from code then we don't need to worry about merging non-conflicting changes from valve, which is an inconvenient long term maintinence item. I haven't reviewed if this technique works in this particular case, let me know your results!

Thanks for the work!

Trildar:
I did try the JavaScript method, which I mentioned briefly in #90. The issue is that the function that I had to change the styles caused lag, presumably due to DOM traversal. Adding to that, I think the styles don't reset when selecting another unit (haven't really tested though), which means the styling function needs to run on every selection change. We could just assume only the styling for the player's hero is important, but it could create minor visual glitches when viewing other units.

There might be some more efficient method, but I can't really think of one.

As a side note, I've completely forgotten to care about the separate unit query panel.

P.S. I'll probably put in the code I used later today when I'm actually on my computer.

PR #114, opened 2017-03-08
Charge_BKB

RamonNZ:
Hello Chris,

I hope this is the correct way to do a pull request with both the item and script files at once.


game/scripts/vscripts/items/charge_bkb.lua
```diff
@@ -0,0 +1,89 @@
...
+		
+		--RamonNZ: Increment the Magic Wand's current charges by 1, but only if CurrentCharges are less than MaxCharges
+		if oldest_unfilled_wand ~= nil then
+			if oldest_unfilled_wand:GetCurrentCharges() < keys.MaxCharges then
+				oldest_unfilled_wand:SetCurrentCharges(oldest_unfilled_wand:GetCurrentCharges() + 1)
+			end
+		end
+		--RamonNZ: start the charges decay timer & first remove it if there's already one in play (we don't want more than one in play removing charges)
+		--RamonNZ: this resets the decay timer every time it gains a charge. Seems fair. Otherwise you could theoretically gain a charge and lose it like 1 second later which would be unfortunate.
+		Timers:RemoveTimer("charges_decay_timer")
```
RamonNZ: 1) ah you were right. It's not that the others don't work, it's just that their charges don't degrade. 
2) yeah I guess...

RamonNZ: Hey chris, I'm looking at RemoveTimer, and there's no way to do it without a name. Have you got any ideas? I guess I could just leave the timer running from the start and not destroy it, but that could lead to some unfortunate situations where somebody gets a charge and it instantly disappears because of unlucky Timer timing.

Trildar: @RamonNZ Dota has `DoUniqueString` to do just that, but do you have a way to store the name so that you can use it to wipe the timer the next time `OnAbilityExecuted` is called? Lua abilities/items seem to be able to store state simply by putting it in properties in their class, but I don't know about DataDriven stuff.

chrisinajar: Just append the playerID to the end of the string. Then they're unique per player. Make sure the code degrades all instances of BKB charges for a given player, otherwise only 1 will decay. It's expected for a given player's BKB charges to decay in sync.

warpdragon:
so basically it's the new header for #base?
Ramon I believe this is due to you starting your item work before chris changed the format of item files to include an extra "DOTAItems" block. Conflicts are annoying. @chrisinajar should he 'git rebase' and add the missing block?

chrisinajar:
rebase or merge, merge would be slightly easier here.

PR #117, opened 2017-03-09
Remove compiled maps from versioncontrol.

Relacibo:
We don't need to have the map-vpk-files in git. This pullrequest removes them and adds the path game/maps to the gitignorefile.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,4 @@ game/panorama/styles
 game/panorama/scripts
 game/panorama/images
 panorama_debugger.cfg
+game/maps


Chronophylos:
I like the Idea of having no build files (compiled and builded files), but if we only have the source files, every time someone updates the map everyone has to re-build the map in Hammer. This can be a bit annoying.

Relacibo:
The aaa.vpk is in the git only since b27dd33476b98efe17f0a95b68b0b87a168cd0f8 and I'd rather not have it in git.

chrisinajar:
I think it's fine to have to build on update. Thanks Relacibo!

PR #122, opened 2017-03-11
fix major bugs in point system, add new event to gamemode

chrisinajar:
there was a race condition where if both teams won within the same 5 seconds then radiant would win, regardless of who actually won first
additionally, points were increasing based on assists as opposed to kills, so 1v1 kills didn't count, and group kills counted for extra


Chronophylos:
I didn't think of that.

PR #132, opened 2017-03-12
Cleaned up some files

Chronophylos:
Removed Precaching of random stuff

### DIFF
--- a/game/scripts/vscripts/internal/gamemode.lua
+++ b/game/scripts/vscripts/internal/gamemode.lua
@@ -161,7 +161,7 @@ function GameMode:_CaptureGameMode()
     mode:SetTopBarTeamValuesOverride ( USE_CUSTOM_TOP_BAR_VALUES )
     mode:SetTopBarTeamValuesVisible( TOP_BAR_VISIBLE )
     mode:SetUseCustomHeroLevels ( USE_CUSTOM_HERO_LEVELS )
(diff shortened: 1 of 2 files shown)


game/scripts/vscripts/internal/gamemode.lua
```diff
@@ -161,7 +161,7 @@ function GameMode:_CaptureGameMode()
     mode:SetTopBarTeamValuesOverride ( USE_CUSTOM_TOP_BAR_VALUES )
     mode:SetTopBarTeamValuesVisible( TOP_BAR_VISIBLE )
     mode:SetUseCustomHeroLevels ( USE_CUSTOM_HERO_LEVELS )
-    mode:SetCustomHeroMaxLevel ( MAX_LEVEL )
+    --mode:SetCustomHeroMaxLevel ( MAX_LEVEL )
```
chrisinajar: We actually use this. The value is set in `settings.lua`

Chronophylos: No we don't, we use the global to generate the table with the needed xp for each level. `SetCustomHeroMaxLevel` is deprecated and not needed when using `SetCustomXPRequiredToReachNextLevel`.

chrisinajar: Good to know :+1:

PR #134, opened 2017-03-13
Ability rewrite for Nature's Call

Trildar:
Some of the summon abilities need to be rewritten to summon stronger units rather than just more units at levels 5 and 6. I would like to do the other abilities as well, but I'm not sure how long that would take me so I'm putting in this PR first.

~~The KV definition for the Giant Treants are in @warpdragon's repo and I didn't include it here in any way. I'm not sure if this is how it should be done. Advice is welcome.~~ Missed the fact that that file was already merged into this repo.


game/scripts/vscripts/abilities/oaa_force_of_nature.lua
```diff
@@ -0,0 +1,116 @@
...
+--   local area_of_effect = self:GetSpecialValueFor( "area_of_effect" )
+
+--   if GridNav:IsNearbyTree( point, area_of_effect, true ) then
+--     --print(ability,"Trees found")
+--     return true
+--   else
+--     FireGameEvent( 'dota_hud_error_message', { reason = 80, message = "#dota_hud_error_must_target_tree" } )
+--     return false
+--   end
+-- end
```
Chronophylos: Why is this disabled function here?

Trildar: Derp. It's sort of an alternate way to do the "Ability Must Target Tree" cast error that's currently done using `CastFilterResultLocation` and the accompanying `GetCustomCastErrorLocation`. The commented function was what I used first cause it was adapted from the Spell Library.

Guess I should just remove that block?

Chronophylos: Yep.

chrisinajar:
I assume there's no need to precache values since you had to have selected natures profit in order to get the ability anyway. if this is incorrect open up a ticket (or a PR that fixes it if you have the time).

Thanks for this, it's also a great example to show others on how to implement somewhat complex abilities in a very simple way!

PR #139, opened 2017-03-14
Added Scatter Shot to Greater Power Treads

SquawkyArctangent:
Made functional Greater Power Treads with Switch Attribute and Scatter
Shot, a Cleave-esque ability that functions on both melee and ranged.


Trildar:
Supposedly Treads are intended to give a different effect from Phase which have split shot. I was thinking it could be splash since cleave doesn't really make sense with ranged, but there hasn't really been discussion on it. I'm not really sure what the rationale for having technically different effects is, but that's what I've heard.

chrisinajar:
Does scatter shot work like cleave or like drow ult? Phase boots are supposed to have the drow-style-cleave where things can proc on the additional hits, where power treads are supposed to be like magnus empower cleave. Both are creep only.

Thanks!

SquawkyArctangent:
It works like neither, although it's closest to splash damage and visually similar to Drow splinter shot. However, the damage dealt is directly based off of your attack damage, rather than the damage dealt with your attack - OnProjectileHit_ExtraData doesn't seem to want to work with this codebase, so I ended up throwing in the towel and dropping the whole crit-transfer thing.

game/scripts/vscripts/items/item_greater_power_treads.lua
```diff
@@ -0,0 +1,299 @@
...
+	
+	return 0
+end
+
+--------------------------------------------------------------------------------
+
+item_greater_power_treads_2 = item_greater_power_treads
+item_greater_power_treads_3 = item_greater_power_treads
+item_greater_power_treads_4 = item_greater_power_treads
+item_greater_power_treads_5 = item_greater_power_treads
```
chrisinajar: you should use an editor that supports editorconfig so that it automatically avoids formatting errors like this

chrisinajar:
That's too bad, so that hit handler in the `IsServer` block doesn't have any information about how much damage was dealt?

SquawkyArctangent:
That's basically what it does, yeah.

And no, the server is the side that deals with all the damage and stuff. I and everyone I checked in with has no idea _why_ ExtraData refuses to work.

Aside from ghosts, but that's only me.

PR #140, opened 2017-03-14
Fixed Decoy Courier KV File to allow game to load

spar36:
There was an extra } at the end which crashed the build in hammer.


chrisinajar:
You have random binary files in this PR, mind fixing those?

spar36:
removed the binaries

PR #143, opened 2017-03-14
Buff/Debuff tooltips and updated furion KV

warpdragon:
Updated KV works with Trildar's Force of Nature plus Buff/Debuff tooltips for items


Trildar:
Note about the Nature's Call (Force of Nature) KVs. When I tried ordering the AbilitySpecial like it is now, basically displacing the normal KVs, it seemed unable to read duration for some reason (making the Treants die instantly due to 0 duration. I could have messed it up at the time, but should be checked.

warpdragon:
When i added in some extra Keys I reordered them to be more organized. Duration was "03" and is now "04". Does this mean it's looking at the wrong key because they're in a different order? I guess the way to check would be to switch the "03" and the "04" and see if it fixes it.

warpdragon:
i assumed that since your code probably references the keys by name, and not number, then i could change their order. if your spell is indeed pulling from "03" in the new file, that would explain the "0" duration

warpdragon:
do you want me to switch the numbers in my repo so they match the old values, or are you going to test independently or what?

Trildar:
You can leave it as it is. I'll test myself when I have time but that might be another day or two, so if you or anyone else wants to do the testing in the meantime then feel free.

warpdragon:
I went ahead and put them back in their original order. with the new ones tacked on the end.

PR #149, opened 2017-03-17
Some Ability Repairs

warpdragon:
Leaving this open for a little bit cuz we noticed some bugs.
1st one: Lion was missing data value for lvl 5 finger (aghs)


chrisinajar:
instead of leaving this open, just open new PR's as you fix issues! :D

PR #156, opened 2017-03-17
You knew it was gonna happen eventually

warpdragon:
power tread damage adjustment


chrisinajar:
40 on all levels, harsh

PR #159, opened 2017-03-17
Fixed io's lvl 5 and 6 abilities

Honeth:
made the custom levels of io work instead of being locked at lvl 4


warpdragon:
i assume u tested it

Honeth:
yeah this should make his abilities have a lvl 5 and 6

PR #160, opened 2017-03-17
Fixed Treads Recipe - Now requires 10 rapiers

warpdragon:
kappa


game/scripts/npc/items/item_power_treads.txt
```diff
@@ -1,5 +1,5 @@
-"DOTAItems"
-{
+"DOTAItems"
+{
```
chrisinajar: get good

chrisinajar:
the recipe is still there, you just made it free. make it *exactly and literally* base dota. Perhaps entirely remove the entry if you can't get it to work

game/scripts/npc/items/item_power_treads.txt
```diff
@@ -1,77 +1,50 @@
...
+			{
+				"var_type"				"FIELD_INTEGER"
+				"bonus_attack_speed"	"25"
+			}		
+			"04"
+			{
+				"var_type"				"FIELD_INTEGER"
+				"bonus_damage"			"0"
+			}		
+		}
```
chrisinajar: don't you need an `ItemRequirements` entry to create it, or will dota just know how because of the base item?

chrisinajar: it looks like the recipe needs to exist but cost 0 and not be purchasable? not sure. test the different things and see

spar36: If you just set the cost to 0, then it doesn't even show up in the shop as required, but i believe you still need the recipe because it tells dota what else needs to be combined.

Issue #146, opened 2017-03-15
Infinite death


chrisinajar:
During duels we disable respawning. For some reason sometimes you die forever even though we re-enable respawn at the end of the duel. Dying forever kind of sucks, so please fix.

spar36:
Question. Why do you re-enable respawn in two locations in the same function?
Lines 255 and 268 of duels.lua

And why do you instantly respawn the units on line 269, but don't on 256

chrisinajar:
it was an attempt to fix the bug, players respawn from the duel fine

RamonNZ:
The above fix has been tested single player, then multi-player with bots. All respawning is working.

RamonNZ:
Hi chrisinajar:

regarding:
"  -- why dont these run?" in duels.lua

The answer is that all of your code fired, and units still didn't respawn regardless.

[VScript] [components\duels\duels:253]	Duel has ended
[VScript] [components\duels\duels:277]	Is this a player id? 5
[VScript] [components\duels\duels:277]	Is this a player id? 6
[VScript] [components\duels\duels:277]	Is this a player id? 7
[VScript] [components\duels\duels:277]	Is this a player id? 8
[VScript] [components\duels\duels:277]	Is this a player id? 9
[VScript] [components\duels\duels:277]	Is this a player id? 0
[VScript] [components\duels\duels:277]	Is this a player id? 1
[VScript] [components\duels\duels:277]	Is this a player id? 2
[VScript] [components\duels\duels:277]	Is this a player id? 3
[VScript] [components\duels\duels:277]	Is this a player id? 4

Progress: Units who die before the duel starts get permadeath. They actually respawn during the duel if they die again or if their first respawn timer hits 0, then after the duel ends they get permadeath.

RamonNZ:
Problem #1: If they die before the duel starts, and have a respawn timer counting down, the timer will still cause a respawn during the duel, whether they're dead or alive. Then if you die during the duel you'll respawn again. Then...

Problem #2: Permadeath happens partly because you save the state of a dead unit, then respawn the unit, then restore it's state (0 HP/dead) with RestorePlayerState, which kills it, but perhaps its respawn timer is 0/-1.

Problem #3: Minor - when you respawn after a duel you respawn where you died instead of at your fountain, meaning possibly right in the middle of creeps or enemy heroes.

Solutions #1: hero:SetTimeUntilRespawn(-1) after the instances of 
                    hero:RespawnUnit()
[...]

PR #162, opened 2017-03-18
Update duels.lua - fix for issue: Infinite death #146

RamonNZ:
Edit: The first fix made single player respawn work, but then I added bots to the game and spotted my coding error, so fixed that.


RamonNZ:
Nope, That fixes single player, but breaks duels. I have to do more testing.

chrisinajar:
This will still have errors when there's an un-even number of heroes trying to duel. I have a similar fix that disables respawn when they're actually added to the duel instead -- with that change, there is still infinite death.

PR #178, opened 2017-03-19
no purchase

spar36:
Makes it so that Boss items aren't purchasable from the shop. 
Just a Fix for Sunday Stream.

When Azazel's shop gets implemented, we can make them purchasable there.


chrisinajar:
Conflicts from warpdragon, womp womp

chrisinajar:
that's how you disable purchasing? I thought there was a whitelist somewhere....

PR #179, opened 2017-03-19
duels.lua - fix for players respawning during duels.

RamonNZ:
This is a fix for players respawning during a duel, regardless of SetRespawnsDisabled(true)

It is loosely linked to the infinite death #146 bug.

RespawnUnit apparently doesn't remove the respawn timer. RespawnHero apparently does.

From the testing I've been able to do, it does actually **stop the respawning**. I haven't noticed any bugs from changing it.

void RespawnHero(bool buyback, bool IsActuallyBeingSpawnedForTheFirstTime, bool RespawnPenalty) 
so RespawnHero(false,false,false)


chrisinajar:
GREAT change, this is something i was planning on doing if my COMPUTER DIDN'T EXPLODE INTO THAT TERRIBLE FIRE

PR #184, opened 2017-03-19
Update item drop table

Rachol:
Removed bottle drop.

Items are dropped about every 3rd creep. 

When it drops, it drops with following rate: 
-clarity with 25.7% rate
-tango with 64.3% rate
-mango with 10% rate


chrisinajar:
i don't know if i'm ready to get rid of bottles, i've grown very attached

Rachol:
:)

There are plenty of problems with bottles currently:
*They start dropping after 130 seconds.
*They make all the gpm calculations by Haganeko go to thrash, so we can't really see the real thing, how the power levels and so on develop.... GPM is through the roof because of them in the early game.
*Creeps are super weak because of them

Rachol:
It would be also interesting if removing that, would make Baumi reconsider drops being a thing...

Chronophylos:
We could make special bottles that can't be selled.-

RamonNZ:
If there absolutely have to be bottles make them very unlikely to drop... watching the stream it is ridiculous. Like seriously.

VoidsKeeper:
Make 6 bottles be crafted into portable-weaker-type-Shrines. You just place them and they heal your team for a certain amount. Maybe also make these "heals" not stack. That at least would give some meaning for the supports to collect them.
Or maybe allow three bottles combine into something (upgraded bottle?) that then can be combined into the portable shrine  (3 upgraded bottles become the Shrine ). 
Or any other variation of this... (I'm sure there's better ways to do it , but just thought this would be of use to supports)

chrisinajar:
🎼I'm in love with a bottle 🎶

Issue #82, opened 2017-02-22
Add 20 player support for the whole project


Chronophylos:
## Sometimes we're assuming we just have 10 player instead of possibly 20.
* Go through the whole code and find these occasions.
* Make sure we're using 20 players.

e.g. change this
```Lua
for playerId = 0,9 do
```
to this
```Lua
for playerId = 0,19 do
```
or look for arrays like this
```Lua
{ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9 }
```
and add the other 10 players

### Places where this needs to be done:
- [ ] gold.lua

yeahbuddy-dota:
or just normalize it. You can attach data to existing structs like so

~~~
for playerId in PlayerResource:GetAllPlayers() do
...
end
~~~
than make your own PlayerResources.GetAllPlayers(self) function.

Trildar:
Is there any reason to not use `DOTA_MAX_TEAM_PLAYERS` (which is currently 24) instead? Even if there are no plans for 12v12.

Chronophylos:
No not really. But @yeahbuddy-dota 's solution is the more elegant solution.

Trildar:
Where should a class extension like that go?

yeahbuddy-dota:
Doesn't matter. In my code I put functions like that in the libraries they interface with, e.g. i have a unit:AddBonus function thats extended in my bonus.lua file. For this my best recommendation would be to make some playerhandler.lua or etc. file that has this function in addition to some others that may be helpful like GetConnectedPlayers(), GetAllRadiantPlayers(), or whatever or stuff like PlayerResource:MakePlayerTable() which will make that {1 = {}, 2 = {}, 3 = {}, ...} table as shown in chris's code.

Trildar:
Why was this closed? It's still relevant though it's not a priority. Most of the code is even assuming 20 players. I've even started trying to implement yeahbuddy's solution. :p

Chronophylos:
That'd be great.

PR #187, opened 2017-03-19
change only available map to "aaa"

ProstoSanja:


game/addoninfo.txt
```diff
@@ -1,3 +1,10 @@
 "AddonInfo"
 {
+	"TeamCount" 	"2"
+	"maps"	"aaa"
+	"IsPlayable"	"1"
+	"aaa"
+	{
+		"MaxPlayers"	"20"
```
chrisinajar: should that be 10?

ProstoSanja:
what about issue #82 ??

Chronophylos:
This map is only playable with 10 player.

PR #216, opened 2017-03-20
Added game length result

ProstoSanja:
It is displayed under time i hero picking phase
hopefully no merging conflicts @chrisinajar

### DIFF
--- a/game/scripts/vscripts/components/points/game_length_vote.lua
+++ b/game/scripts/vscripts/components/points/game_length_vote.lua
@@ -35,19 +35,23 @@ function GameLengthVotes:SetGameLength ()
     end
   end
 
-  local length = 'normal'
+  local length
   if votes.long > votes.normal then
     if votes.long > votes.short then
       length = 'long'
+      CustomNetTables:SetTableValue( 'team_scores', 'limit', { value = 200, name = length } )
     else
       length = 'short'
+      CustomNetTables:SetTableValue( 'team_scores', 'limit', { value = 50, name = length } )
     end
   elseif votes.short > votes.normal then
     length = 'short'
+    CustomNetTables:SetTableValue( 'team_scores', 'limit', { value = 50, name = length } )
+  else
+    length = 'normal'
+    CustomNetTables:SetTableValue( 'team_scores', 'limit', { value = 100, name = length } )
   end
   DebugPrint ( 'votes ' .. votes.short .. ', ' .. votes.normal .. ', ' .. votes.long .. ' result: ' .. length)
   GameRules.GameLength = length
-  CustomGameEventManager:Send_ServerToAllClients("gamelength_vote_confirmed", {
-    length = length
-  })
+
 end
--- a/game/scripts/vscripts/components/points/game_length_vote.lua
+++ b/game/scripts/vscripts/components/points/game_length_vote.lua
@@ -47,4 +47,7 @@ function GameLengthVotes:SetGameLength ()
   end
   DebugPrint ( 'votes ' .. votes.short .. ', ' .. votes.normal .. ', ' .. votes.long .. ' result: ' .. length)
   GameRules.GameLength = length
+  CustomGameEventManager:Send_ServerToAllClients("gamelength_vote_confirmed", {
+    length = length
+  })
 end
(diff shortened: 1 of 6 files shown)


game/scripts/vscripts/components/points/game_length_vote.lua
```diff
@@ -47,4 +47,7 @@ function GameLengthVotes:SetGameLength ()
   end
   DebugPrint ( 'votes ' .. votes.short .. ', ' .. votes.normal .. ', ' .. votes.long .. ' result: ' .. length)
   GameRules.GameLength = length
+  CustomGameEventManager:Send_ServerToAllClients("gamelength_vote_confirmed", {
```
Chronophylos: You should use `CustomNetTables.GetTableValue('team_scores', 'limit')['value'];` instead of a event.

ProstoSanja: why is an event a bad variant for you? it was okay for you couple of hours ago

Chronophylos: I simply forgot we have a NetTable with current scores and the limit.
NetTables are better because if someone reconnects the NetTable repopulates. Via events the information is lost.

ProstoSanja: can you give any advise on implementing? or example?

Chronophylos: The NetTable is already there. You can access the limit with `CustomNetTables.GetTableValue('team_scores', 'limit')['value'];`.
Implementation is [here](https://github.com/OpenAngelArena/oaa/blob/master/game/scripts/vscripts/components/points/points.lua#L19-L31)

ProstoSanja: i would like to stick to this one, as 
1) it is only needed during pick phase, 
2)nettables do not emit pushes, so it is hard to catch them
3) there is a failover, which in case of pick time reconnect will display "open angel arena"

Chronophylos: 2: [They do](https://github.com/OpenAngelArena/oaa/blob/master/content/panorama/scripts/custom_game/pointsmanager.js#L8)

ProstoSanja: okay, but other thing is that this nettable you referenced is emitted only when game is loaded(after pregame hero pick screen), while event is done as soon as chosen

Chronophylos: You could move [this](https://github.com/OpenAngelArena/oaa/blob/master/game/scripts/vscripts/components/points/points.lua#L19-L31) to [here](https://github.com/OpenAngelArena/oaa/blob/master/game/scripts/vscripts/components/points/game_length_vote.lua#L49). This would be better either way.

ProstoSanja: 1) this will break hierarchy
2) it is still not that, it outputs numbers, and note text.
Let it be like so, what dose @chrisinajar think?

Trildar: Not too sure what hierarchy you mean. I think it makes more sense to set table values relating to length in the `GameLengthVotes` module anyway, though the score initialisation should probably stay in int `PointsManager`.

Numbers can be very easily converted to text using `switch` or some other solution. I'm guessing you meant that it would require changing if the length and kill limit relation was ever changed.

chrisinajar: There is already an event for this from the net table when the value changes. It's basically the exact same implementation on the JS side with less code on the lua side. It also keeps the management of score and score cap scoped to exactly 1 object who's modifications are tracked by the dota engine.

chrono is also right about moving that code, i'm pretty sure it's needed to make it work right. also that's really where it should be, since that's the chunk of code with authority over the `limit`.

points manager should still initialize the value for `score`, just not `limit`

also lua doesn't have a switch statement, you have to use a table to do something like that. it's really dumb.

hierarchy?? 😂

Trildar:
I'm aware Lua doesn't have a switch. I ran into that when writing the Demonic Conversion rewrite. Panorama is JavaScript though, so unless you intend to put the length as a string in the net table as well, you'd be using switch in JavaScript.

ProstoSanja:
I'll deal with it

ProstoSanja:
Evtything done and tested, game works and finishes successfully,
technical description under commit,
ready to merge

Issue #201, opened 2017-03-20
Wraith King's Ult does not respawn him during duels


KarmaticNeutral:
Supposedly the duel always treats WK's first death as duel ending and prevents him from repsawning.

chrisinajar:
There's some sort of isReincarnating flag on heroes, when a hero dies you should check that flag and allow them to respawn if it is set, otherwise count the point and move on.

Additionally, it's possible that killing a reincarnating wraith king falsely counts for a point, that should also be fixed.

Chronophylos:
Why did you reopen this?

KarmaticNeutral:
because we had 4 reports of it occuring today during the open session

Chronophylos:
Ah ok

KarmaticNeutral:
definitely broken. Also, it tries to respawn him, but the duel is over so he gets stuck outside.

Chronophylos:
So WK trys to respawn but the duel is over?

KarmaticNeutral:
yes

warpdragon:
This is resolved?

PR #217, opened 2017-03-20
Fixed Wraith King reincarnation (#201)

Chronophylos:
Fixed points increasing during reincarnation
Fixed WK not respawning during duel

###### Note: I haven't tested this.

### DIFF
--- a/game/scripts/vscripts/components/duels/duels.lua
+++ b/game/scripts/vscripts/components/duels/duels.lua
@@ -44,7 +44,7 @@ end
 local DUEL_IS_STARTING = 21
 
 function Duels:CheckDuelStatus (keys)
-  if not Duels.currentDuel or Duels.currentDuel == DUEL_IS_STARTING then
+  if not Duels.currentDuel or Duels.currentDuel == DUEL_IS_STARTING or keys.killed:IsReincarnating() then
     return
   end
 
--- a/game/scripts/vscripts/components/points/points.lua
+++ b/game/scripts/vscripts/components/points/points.lua
@@ -32,7 +32,7 @@ function PointsManager:Init ()
 
   GameEvents:OnHeroKilled(function (keys)
     -- increment points
-    if keys.killer:GetTeam() ~= keys.killed:GetTeam() then
+    if keys.killer:GetTeam() ~= keys.killed:GetTeam() and not keys.killed.IsReincarnating() then
       PointsManager:AddPoints(keys.killer:GetTeam())
     end
   end)


ProstoSanja:
Hello, this ruined duels, isreincarnating should be called with arguments, now they dont end

[ W VScript                ]: Script Runtime Error: scripts/vscripts/components\points\points.lua:35: IsReincarnating called with 0 arguments - expected 1
[ W VScript                ]: stack traceback:
[ W VScript                ]: 	[C]: in function 'IsReincarnating'
[ W VScript                ]: 	scripts/vscripts/components\points\points.lua:35: in function 'fn'
[ W VScript                ]: 	scripts/vscripts/libraries\event.lua:31: in function 'OnHeroKilledEvent'
[ W VScript                ]: 	scripts/vscripts/events.lua:331: in function 'OnHeroKilled'
[ W VScript                ]: 	scripts/vscripts/internal\events.lua:92: in function <scripts/vscripts/internal\events.lua:56>

Chronophylos:
There is no argument in the docs.

ProstoSanja:
But still script crashes at this point with logs I sent above
And removing this line fixed it, so we need to find g a sample somewhere

ProstoSanja:
ive found the fix, it should be called like this
`keys.killed.IsReincarnating`
no : or ()
but now another problem appears, because of dul not allowing to respawn, he reincarnates, but stays dead at the same time, and duel doesnt end

Chronophylos:
Enabling respawn and disableing respawn could fix it.

ProstoSanja:
well , i am not into this mechanics, and dont even know where it is, so good luck further 🥇

Chronophylos:
I should test things before I push them..
Thanks @ProstoSanja!

Trildar:
I'm pretty sure it should be `keys.killed:IsReincarnating()`. Without colon or brackets would give you the function handle. The colon makes it call the function with a reference to its `self` object as a first argument. Hence why it's complaining of a lack of arguments.

ProstoSanja:
i tested 
`keys.killed.IsReincarnating` and it worked, duel didnt end when he was reincarnating

Chronophylos:
Ohh thats what `:` and `.` mean!

Trildar:
@ProstoSanja Just FYI, if you didn't test if duels could still end at all, probably what happened is that `keys.killed.IsReincarnating` gives the function handle and in Lua, a function handle would be considered truthy and so that block just always runs (so long as `keys.killed` actually has some value or method called `IsReincarnating`).

chrisinajar:
: passes a self context to the method, . does not. Functions defined with a : have a "secret" first parameter for self built in.

function ASDF:Add(b)
  return self + b
end

is the same as

function ASDF.Add(a, b)
  return a + b
end

and can be run like

ASDF.Add(1, 3) -- returns 4


All valve functions expect a context.

Chronophylos:
Ah, thanks.

SphereKatzen:
It's a method call, so there is an argument, it must not be being called on the correct object. Try storing the hero in variable first maybe?

PR #222, opened 2017-03-21
Switch to tables for level requirements

SphereKatzen:
This is more adaptable to future changes in level requirements, also clearer.

### DIFF
--- a/game/scripts/vscripts/components/abilities/level.lua
+++ b/game/scripts/vscripts/components/abilities/level.lua
@@ -16,8 +16,8 @@ function AbilityLevels:FilterAbilityUpgradeOrder (keys)
   end
 
   -- Ability hero level requirements
-  local basic_reqs = {0, 0, 0, 0, 28, 40}
-  local ultimate_reqs = {0, 0, 0, 37, 49}
+  local basicReqs = {0, 0, 0, 0, 28, 40}
+  local ultimateReqs = {0, 0, 0, 37, 49}
 
   local ability = EntIndexToHScript(keys.entindex_ability)
   local player = PlayerResource:GetPlayer(keys.issuer_player_id_const)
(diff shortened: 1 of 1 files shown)


Trildar:
You didn't declare `req_table` and `requirement`. Does Lua default to function scope? I'm not that familiar with details like that when it comes to Lua. I also don't know why you wouldn't just stick with `requiredHeroLevel` instead of `requirement`, but that's partially me being petty about my own code I suppose.

game/scripts/vscripts/components/abilities/level.lua
```diff
@@ -16,43 +16,33 @@ function AbilityLevels:FilterAbilityUpgradeOrder (keys)
   end
 
   -- Ability hero level requirements
-  local basicLevel5Req = 28
-  local basicLevel6Req = 40
-  local ultimateLevel4Req = 37
-  local ultimateLevel5Req = 49
+  local basic_reqs = {0, 0, 0, 0, 28, 40}
+  local ultimate_reqs = {0, 0, 0, 0, 37, 49}
```
Trildar: Ultimates have 5 levels, not 6.

game/scripts/vscripts/components/abilities/level.lua
```diff
@@ -16,43 +16,33 @@ function AbilityLevels:FilterAbilityUpgradeOrder (keys)
...
-    end
+    req_table = ultimate_reqs
   end
-
-  if abilityCanUpgrade then
+  
+  if abilityLevel > #req_table then
+    requirement = req_table[#req_table]
+  else
+    requirement = req_table[abilityLevel]
```
Trildar: Should be `abilityLevel + 1` because the function acts before the ability upgrade, meaning that when trying to go from level 4 to 5 of a basic ability, `abilityLevel` would be 4

SphereKatzen:
This is actually my first time writing lua, I just assumed it would default to function scope, which seems to be wrong! I've fixed the issues now, it should work fine now.

SphereKatzen:
Done!

Issue #227, opened 2017-03-21
Boss item distribution ui is not intuitive


ProstoSanja:
![image](https://cloud.githubusercontent.com/assets/26527928/24141680/db8c9b38-0e2c-11e7-8aae-f2d68c83542a.png)
these buttons on the right of ui are not intuitively understandable, we need to add tooltips, similar to those when you hover items on them

PR #228, opened 2017-03-21
Reenabled tooltips, tested works ok

ProstoSanja:
image

reenabled them, clicking works okay
also a reference for issue #227


chrisinajar:
Art team wanted the text removed, but until those lazy bums give us real icons this is much better. I forgot I ever had these!

Description needs work wrap too, if you want to dig into that next :)

Issue #229, opened 2017-03-21
Yesterday wrait king fix ruined duels!


ProstoSanja:
problem is available at #217

PR #226, opened 2017-03-21
Refactored duel mechanic, hopefully fixed a few issues in the process

SphereKatzen:
I've refactored the duel mechanic to have one "duels" handler that maintains the timers and generates teams for duels. It then instantiates duel objects which take care of player state and movement into/out of the duel. I think there was some kind of problem somewhere in the team allocation code that was causing a lot of the issues with people not being sent into duels properly etc. Hopefully this fix the issue, and if not it'll be a lot quicker to fix now that the functions are much simplified.

PR #239, opened 2017-03-21
Fix WK duel reincarnation fix

Chronophylos:
Fixed the fix ( #229 )


ProstoSanja:
also a question, did you count with #226 ?

Chronophylos:
No, I'll resolve the conflicts.

Chronophylos:
As soon as #226 has the fix I'll close this.

ProstoSanja:
Also a question, shouldnt you do the same thing at game/scripts/vscripts/components/points/points.lua, because it has same error i think

Chronophylos:
I used `.` instead of `:`. In points.lua I used the right one.

game/scripts/vscripts/components/duels/duels.lua
```diff
@@ -47,6 +47,12 @@ function Duels:CheckDuelStatus (keys)
   if not Duels.currentDuel or Duels.currentDuel == DUEL_IS_STARTING or keys.killed:IsReincarnating() then
     return
   end
+  if keys.killed:IsReincarnating() then
+    keys.killed:SetRespawnsDisabled(false)
+  end
+  Timers:CreateTimer(1, function ()
+    keys.killed:SetRespawnsDisabled(true)
```
chrisinajar: this will race with the end of a duel and cause infinite death.

last kero is killed
timer is created
end duel handler runs
all heroes have respawn enabled
timer runs, disables respawn

chrisinajar: this code will also never run since there's a check to `or keys.killed:IsReincarnating() then` right before it

PR #250, opened 2017-03-21
Working preemptive core 1, but building it crashes the game

chrisinajar:
why does it crash the game when you build this item?

if you add it to your inventory with -item, it works perfectly fine.


chrisinajar:
i think the error is
[ W Entity System  ]: Cannot create an entity because entity class is NULL -1

PR #254, opened 2017-03-21
Complete nuke gulp + redundant russian file

ProstoSanja:
Remove all gulp mentions
there are two files with exactly same content in russian, i deleted one with english title and remainde one with russian


package.json
```diff
@@ -1,8 +1,8 @@
 {
-  "name": "angel_arena_allstars",
-  "version": "0.0.0",
+  "name": "open_angel_arena",
+  "version": "0.0.1",
```
chrisinajar: we're not at 0.1 yet, and we'd use npm cli to do it anyway (for auto-tagging and all that fanciness)

package.json
```diff
@@ -1,8 +1,8 @@
...
   "description": "A really great Dota 2 game mode.",
-  "main": "index.js",
+ // "main": "index.js",
```
chrisinajar: .json doesn't support comments, this will throw errors later

package.json
ProstoSanja: do you want me to revert version AND name?

Chronophylos: Version only

ProstoSanja:
Ready for merge

ProstoSanja:
@chrisinajar

PR #275, opened 2017-03-22
Update settings to allow bot AI.

RamonNZ:
To allow AI bots to work properly: This change must be made or bot AI won't do much at all.

I have a basic bot script which I'll add soon.

Bot requested by chrisinajar & Warpdragon.

### DIFF
--- a/game/scripts/vscripts/settings.lua
+++ b/game/scripts/vscripts/settings.lua
@@ -29,7 +29,7 @@ BUYBACK_ENABLED = false                 -- Should we allow people to buyback whe
 DISABLE_FOG_OF_WAR_ENTIRELY = false     -- Should we disable fog of war entirely for both teams?
 USE_UNSEEN_FOG_OF_WAR = false           -- Should we make unseen and fogged areas of the map completely black until uncovered by each team?
                                             -- Note: DISABLE_FOG_OF_WAR_ENTIRELY must be false for USE_UNSEEN_FOG_OF_WAR to work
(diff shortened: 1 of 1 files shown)


Honeth:
if they act like normal dota wouldn't that mean they buy items corresponding to their hero? if so that would make them really bad at OAA (I'm thinking in a perspective for when they substitute for a disconnected player)

Alpha-Omega1996:
@Honeth agreed...

Chronophylos:
We can write the Bots AI, so we decide if and what they buy. Or am I wrong?

Honeth:
we can change what they do but i'm not sure if this value would need to be true for that would require some minor testing but i was looking at a possible future fix for disconnected players this should be fine for testing as is

Alpha-Omega1996:
either i'm overthinking it or it is hard to write one...

Honeth:
writing a completely new bot would certainly test your capabilities (if you want it to feel like a player) although it isn't impossible using base dota bots can work but the code itself even says in the commented part that for it to be true the game needs 3 lanes, normal items , etc. so it might be a bit buggy for OAA if we don't write our own basic bots

Alpha-Omega1996:
as expected...........

Haganeko:
I say we leave the ambition of substituting disconnected players for now and just make some decent bots available for testing and perhaps noobs learning the game, later.

chrisinajar:
bots are great for testing. there's no way to spawn them without cheats on.

also most complex change eu

PR #280, opened 2017-03-23
Added extra install instructions if the map is not compiled

darklordabc:
I runned into the "map not specified by sever" problem, so added some extra info to help any future devs who find themselves in the same sticky situation.


chrisinajar:
super psyched to see stuff coming in from you darklord! :D :D

PR #285, opened 2017-03-23
Recoloured the extra upgrade cores

darklordabc:
I've got no idea the discussion that went on before this or if there was a reason they are all green, so feel free to close this PR, I wont be offended it only took 10 minutes. 
image


chrisinajar:
Strict improvement.

PR #286, opened 2017-03-23
Made Skip Phase Faster

darklordabc:
Made it so skip phase is always enabled in tools mode, also skips the selection phase (about 1 second), and also removed the showcase phase. 

Enabling it always in toolsmode will cause problems if you are trying to test the Panorama menu at the start, but having it enabled, with these modifications will save like 4 or 5 seconds each time you restart the map, which will add up over time.


chrisinajar:
is `SKIP_TEAM_SETUP` from settings or something?

darklordabc:
SKIP_TEAM_SETUP is the constant already. I've removed the IsToolsMode condition, so that the skipping only happens if players specify to use the SKIP_TEAM_SETUP mode. SKIP_TEAM_SETUP already skips the panorama start screen, so it doesn't seem sensible to make any extra skip constants.

PR #293, opened 2017-03-23
Pull translations from transifex

chrisinajar:
Basic scripts for pulling in translations from transifex. Also built files for each language.


Chronophylos:
Cool.

Issue #295, opened 2017-03-24
logging system fails when path to an executed lua script is too long


mokonaDesu:
The reason is that lua's debug.traceback() function truncates very long paths and our internal logger code fails when trying to match regular expression agains those.

PR #296, opened 2017-03-24
Fixed logging failing on long script file paths

mokonaDesu:
A fix for #295. Should also make logging a tiny bit faster :D


chrisinajar:
cool, hope it works :D

PR #297, opened 2017-03-24
Hero progression component

### LINKED ISSUE #261: Fix rate at which heroes get ability points passed level 25
After level 25, heroes should not get any new ability points until level 28. At 28 they get 1 point, and receive 1 more point each 3rd level thereafter. So ability points would be gained at 28, 31, 34, 37, 40, 43, 46, and 49.

mokonaDesu:
Implemented the hero progression component - register all 'levelup' and exp gain handlers there.
Resolved #261


mokonaDesu:
Shit. A accidently included the logging fix in the pull request as well. It should be all good after this one is accepted though: #296

game/scripts/vscripts/components/progression/hero_progression.lua
```diff
@@ -0,0 +1,48 @@
...
+function HeroProgression:ShouldGetAnAbilityPoint(level)
+    -- After level 25 hero gets an additional skill point every 3 levels
+    return level < 25 or math.fmod(level, 3) == 1
+end
+
+function HeroProgression:ProcessAbilityPointGain(hero, level)
+    DebugPrint('Processing the ability point for ' .. hero:GetName() .. ' @ level ' .. level)
+    if not self:ShouldGetAnAbilityPoint(level) then
+        DebugPrint('...taken it away!')
+        hero:SetAbilityPoints(hero:GetAbilityPoints() - 1)
```
chrisinajar: What happens once the hero reaches the point where dota thinks you have enough ability points? Or does it just always try to add them no matter what?

mokonaDesu: It doesn't check for that case - just keeps adding those every three levels. we might tweak the self:ShouldGetAnAbilityPoint so that it stops giving them out completely after certain level if that's needed. Or do you mean the case where all of the abilities are already maxed? not sure... hope settings ability points to -1 doesn't crash the game ^_^

PR #303, opened 2017-03-24
Functions for getting sets of player IDs (using Lua Fun)

Trildar:
I also modified the gold module to use the new functions, as an example.


Chronophylos:
I resolved your conflict. Git was beeing silly

chrisinajar:
FINE 👍

PR #312, opened 2017-03-24
removed combiner from recipes

warpdragon:
welp. here you go. looks like shit. not much i can do that i haevn't already done. i need to figure out why my line endings are constantly fucking me.

I tried so many things. i did this same goddamn commit 3 times. i'm not redoing it.


chrisinajar:
> blindly merges PR from warpdragon

PR #315, opened 2017-03-24
Stoneskin Delay

spar36:
added in delay for Stoneskin toggle and made toggle respect cooldown


chrisinajar:
looks simple enough!

PR #316, opened 2017-03-24
Add Nimbus KV Override

spar36:
Added the zuus nimbus KV for editing later for nerfing.
*Stolen from Redux*
Thnx Darklord!


chrisinajar:
which value controls the speed?

PR #319, opened 2017-03-25
26 second timer fix attempt 2

RamonNZ:
Hi Chronos,

I knew there would be a cleaner way to do it.

This fixes it with one line of code, and all the checks are already done within events.lua

(P.S. I'm not sure if there's a way to attach another file to my PR using github.com's gui. I need to learn commandline I know)

### DIFF
--- a/game/scripts/vscripts/internal/events.lua
+++ b/game/scripts/vscripts/internal/events.lua
@@ -84,6 +84,8 @@ function GameMode:_OnEntityKilled( keys )
 
     if not killerPlayerEntity:IsRealHero() then
       killerPlayerEntity = killerEntity:GetOwnerEntity() or killerEntity
+      --Change respawn timer for hero killed by creep.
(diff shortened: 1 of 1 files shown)


game/scripts/vscripts/internal/events.lua
```diff
@@ -84,6 +84,8 @@ function GameMode:_OnEntityKilled( keys )
 
     if not killerPlayerEntity:IsRealHero() then
       killerPlayerEntity = killerEntity:GetOwnerEntity() or killerEntity
+      --Change respawn timer for hero killed by creep.
+      killedUnit:SetTimeUntilRespawn(5)
```
chrisinajar: i think you should create a new conditional and check the team of the killer. this would also resolve the last of the non-hero-entities problems if in the else block you put the hero killed event.

chrisinajar:
we can fix that later

PR #327, opened 2017-03-25
Fix hero creep respawn

Chronophylos:


chrisinajar:
fixed it while i was working on something else, thanks though!

PR #332, opened 2017-03-25
added some new icons for stoneskin, dashstaff, demon blood, and renam…

warpdragon:
added some new icons for stoneskin, dashstaff, demon blood, and renamed old stoneskin


darklordabc:
Look pretty good imo

PR #347, opened 2017-03-25
Get rid of script error.

Aesylwinn:
The dropped item despawner does not check if the item exists before removal which causes errors similar to the following:
> scripts/vscripts/components\creeps\item_drop.lua:62: This object has been deleted from C++ and no longer exists. (Use :IsNull() to detect this.)

I'm not sure if this is the correct way to fix it, but it seems to work. On a side note, a potentially unintended consequence of the current implementation is that loot items which are dropped within the despawn time will still be deleted.


Aesylwinn:
Looks like this is partially implemented in the Bottles PR. I'll leave this up for now, but it probably shouldn't be merged.

Chronophylos:
I like your solution. Let's use this as soon as Bottles gets merged.

chrisinajar:
> Detect entity has been deleted from C++
https://developer.valvesoftware.com/wiki/Dota_2_Workshop_Tools/Scripting/API/CBaseEntity.IsNull

DoctorGester:
IsNull won't work. This warning is wrong. IsValidEntity is a correct solution.

Chronophylos:
Comparing with `nil` also works.
I'm implementing this in Bottles 2.0.
Thanks

chrisinajar:
The value is not `nil` when it has been deleted from C++ but still exists in hscript. Instead you just get that error. `IsValidEntity` seems like the most correct, I assume it also checks for `isNull`.

Aesylwinn:
Just to note, when I wrote:
> loot items which are dropped within the despawn time will still be deleted.

I meant to write:
> loot items which the player chooses to drop within the despawn time will still be deleted.

It is an edge case that may or may not be important.

Aesylwinn:
So... after looking at the main repo and the opent PR's, I don't see this fix being implemented anywhere. Was it fixed in another way? It wasn't specific to the bottles PR when I tested it. Anyway, let me know in the next couple of days if this fix is wanted since I'll be deleting it otherwise.

Chronophylos:
Oh yeah, I forgot this.
This should get merged.

PR #349, opened 2017-03-25
moved icons to custom subfolder, also duplicated them into panorama c…

warpdragon:
…ontent folder as instructed by moddota


warpdragon:
This should fix some of the bugs people have been seeing with missing icons.

PR #351, opened 2017-03-26
Balance 26 03

l34Um1:


KarmaticNeutral:
Bristle backs quillspray has 7 levels in the d9a2a59 balance item

Trildar:
Technically the ability still has a max level of 6, but yes, there are 7 values for `MaxDamage` there.

Issue #210, opened 2017-03-20
Frozen Sigil LUA help


warpdragon:
I created the NPC units for Frozen Sigil and added them to npc_units_custom.txt and they still don't spawn. Same thing happened with beastmaster. Anyone know the solution?

Chronophylos:
Does Dota know what to do at level 5 & 6?

ProstoSanja:
I looked through files, and I couldn't find any Lua associated with tusk or sigil...
Pretty strange

KarmaticNeutral:
Demonic Conversion levels 5&6 work. Maybe compare to that?

warpdragon:
maybe i made a mistake. there's a first for everything right?

Trildar:
@ProstoSanja There is no Lua currently written for Tusk sigil cause it's currently using the built-in ability. A Lua rewrite of the ability will probably need to be done though.

@KarmaticNeutral Demonic Conversion only works cause it was rewritten in Lua to spawn the level 5 and 6 stuff. The Eidolons aren't even named numerically internally, so it would never work with the built-in ability. Beastmaster Call of the Wild needed to be rewritten for some reason even though the units *are* named numerically (probably Valve being silly about not looking for units dynamically like a number of the other summon abilities seem to do). Possibly the same thing is happening here with Tusk sigil.

PR #367, opened 2017-03-27
Add level 5 and 6 Tusk Frozen Sigil unit

### LINKED ISSUE #190: Tusk sigil levels 5 & 6
Doesn't spawn

Trildar:
Closes #190 and #210.

Unit names for level 5 and 6 Treants are also in here because I'm not sure about how well conflicts on `addon_english.txt` and other localisation files is handled. If it would be better in a separate PR just leave a comment.


chrisinajar:
Github wont automatically merge in changes to addon_english, so any time it's changes it'll instantly conflict with any other pending change, and will require the author of the other PR's (in this case also you) to manually merge the changes before they can go in.

short answer: For `addon_english` and `addon_english` only, fewer PR's is better.

PR #376, opened 2017-03-27
Implement charger boss

chrisinajar:
This is a work in progress.


chrisinajar:
This is probably as ready as it will get...

PR #379, opened 2017-03-28
Anti-Spawn Camp: Immune to Damage while near Foutain

darklordabc:
There is dozens of way to stop fountain camping this is just one approach. This declares a basic damage filter, and when any hero receives damage it checks if they have the fountain rejuvenation buff and if they do it sets the damage to zero. This also changes the default modifier to give player some information. 
image

The buff can linger for a few seconds after leaving fountain, which means that players can fight outside their area with an advantage, and if they teleport to another player, they will be invulnerable for a couple of seconds. Also, we may want to add a filter on the damage to allow some types of damage being applied while heros are in fountain area, like Zeus's ult prehaps.


Haganeko:
One question: what about hook?

DoctorGester:
It's just a damage filter. What about hook?

darklordabc:
Hook would still work, but from my knowledge Baumi wanted hooks to still be effective. All spell disables are still enabled, just nothing will cause damage. Also, when players get hooked out, they will still be invulnerable for a few seconds, so they might be able to escape back to fountain.

Haganeko:
I don't know... IMO Hook is the worst part of fountain camping. Especially since no one can go into the fountain to kill people - fountain is too strong for that

darklordabc:
Hooking will be drastically reduced in its effectiveness for a couple of reasons. 
* Players will be invulnerable for about 3 or 4 seconds after getting hooked, so after a successful hook and the enemy team unloading damage, all of it will be ignored, and the hooked player will have 4 seconds to get back to the aura. 
* Camping outside the foutain will become a easier battle to win for the defending team, because of previously mentioned linger of god mode, and the fact they simply have to walk back a bit to get invulnerability. This means defenders can more easily attack heroes who are trying to siege them.

darklordabc:
It would be trivial to prevent hook strategy, like making any players that damage a hero with fountain rejuvenation get teleported back to their base but I think that the current change in the PR will be more enough.

chrisinajar:
addon_english conflicts D:

darklordabc:
Conflicts resolved

Chronophylos:
Nope 
> Conflicting files
game/resource/addon_english.txt

darklordabc:
Should hopefully be fixed

chrisinajar:
don't you love how we get to merge in 1 addon_english change at a time? D: D: D:

chrisinajar:
lololololololololololol

game/scripts/vscripts/gamemode.lua
```diff
@@ -76,7 +76,30 @@ require('components/index')
...
+function GameMode:FilterDamage(filterTable)
+  local victim_index = filterTable["entindex_victim_const"]
+  local attacker_index = filterTable["entindex_attacker_const"]
+  local ability_index = filterTable["entindex_inflictor_const"]
+  if not victim_index or not attacker_index then
+      return true
+  end
+  local victim = EntIndexToHScript(victim_index)
+  local attacker = EntIndexToHScript(attacker_index)
+  print("damaged")
```
chrisinajar: maybe don't print on damage?

darklordabc: removed

Chronophylos:
Don't worry about the changelog.
I'm doing it myself. In fact the version on github is not up-to-date.

chrisinajar:
I don't think the solution to spawn camping is to make the camp stronger anyway, it's better to not have massive power spikes that cause heroes to become so strong they can take on your whole team while fountain healing.

Perhaps we can add this in later, but closing for now to clean things up. It should also probably have been a component.

PR #380, opened 2017-03-28
chargeBKB Fix

spar36:


chrisinajar:
I assume you made sure none of the other ones had this problem? Seems like a find -> replace problem when warp did the icon fix

spar36:
This was just the one item that Karmatic told me about. none of the other items have had trouble so far being made.

KarmaticNeutral:
I am 92% confident that this does not effect any items that are currently implemented.

PR #386, opened 2017-03-28
Infinite Bottle Charges

darklordabc:
This PR allows players to pick up endless amount of bottles and have infinite charges. 
image
**Details:**
* When a hero uses a bottle, they receive a modifier that shows how many extra charges of bottle they have left.

* When hero picks up a 2nd bottle, that bottle is removed and its 3 charges are added to the stack.
* When a bottle gets down to 0 charges, it checks the stackcount, if there is more than 0 stacks, it fills the bottle back up to 1 charge, allowing players to use the bottle for as long as they have charges. If no stacks left, the bottle and the modifier is removed.
* Bottles are no longer purchasable from shop, and picked up bottles cannot be sold to the shop. 

They may be a few corner cases that slipped through the net, so there may be bugs in rare cirumstances, if so I'll patch up.

P:S: Ehh I screwed up a little on the PR, and for some reason this also has unrelated improvements to bots, it makes them level up abilities when they have free points, before they would get stuck at their default dota 2 levels and not upgrade to max levels.


game/scripts/npc/items/item_bottle.txt
```diff
@@ -0,0 +1,67 @@
...
+            }           
+            "03"
+            {
+                "var_type"              "FIELD_FLOAT"
+                "restore_time"          "2.5"
+            }
+            "04"
+            {
+                "var_type"              "FIELD_INTEGER"
+                "movement_speed_percent_bonus"          "-30"
```
zelding: Why is this here? Does the bottle really reduce ms, or is it just an orphan stat?

game/scripts/npc/npc_items_custom.txt
```diff
@@ -46,6 +46,7 @@
 #base "items/item_bloodthorn_3.txt"
 #base "items/item_bloodthorn_4.txt"
 #base "items/item_bloodthorn_5.txt"
+#base "items/item_bottle.txt"
```
zelding: Shouldn't this item be in the `npc_item_override.txt` ?

game/scripts/npc/items/item_bottle.txt
darklordabc: Good find, but it appears to be just an orphan stat, I copied and pasted the bottle definition from the official source : https://github.com/SteamDatabase/GameTracking-Dota2/blob/master/game/dota/pak01_dir/scripts/npc/items.txt

game/scripts/npc/npc_items_custom.txt
darklordabc: I don't think we have that file, we only have npc_abilities_override

game/scripts/npc/items/item_bottle.txt
Chronophylos: >  If Bottle does not have full charges and is placed on a courier, the courier will move 30% slower.

http://www.dota2.com/items/

Chronophylos:
This is way better than my solution.

game/scripts/npc/items/item_bottle.txt
darklordabc: ahhh

zelding: We should put that on the tooltip ;) Maybe just as a NoteX

Chronophylos:
But does this work if your inventory is full?

darklordabc:
I dont think so.

Declow0:
If bottle has infinte charges and it is not purchasable and sellable may be easy way write simple implemenation without "modifier_bottle_charges". I tryed to do it and I had only 1 problem: bottle can't be filled in the fountain.

game/scripts/vscripts/libraries/chatcommand.lua
```diff
@@ -120,11 +120,20 @@ function ChatCommand:OnPlayerChat(keys)
...
-	    mode:SetFogOfWarDisabled(true)
+	    GameRules:GetGameModeEntity():SetFogOfWarDisabled(true)
 
 	  -- Bring back the fog of war
 	  elseif string.find(text, "-fog") then
-	    mode:SetFogOfWarDisabled(false)
+	    GameRules:GetGameModeEntity():SetFogOfWarDisabled(false)
+
+	  -- Bring back the fog of war
+	  elseif string.find(text, "-fixspawn") then
```
darklordabc: It was added by accident like the bot stuff, I think those changes are already merged previously.

Haganeko:
I don't know if our bottle needs to be fillable in the fountain

darklordabc:
@Chronophylos I've updated it so that it works even if you have full inventory.

Chronophylos:
And what about the backpack?

darklordabc:
It wont count the backpack, but if the player ever has more than 1 bottle in his actual inventory, it will remove the extra.s

Declow0:
I try to run this branch and bottles don't stack...

darklordabc:
Can you screen shot it in shop when I checked I couldn't see it.

Declow0:
Shop

Trildar:
@Declow0 Just to clarify in case you misunderstood. Infinite doesn't mean one Bottle lasts forever. It just means you can acquire an unlimited amount of charges by picking up more Bottles (one charge per bottle I think), which are supposed to drop from creeps (and probably at a significantly lower rate than they currently do).

Declow0:
@darklordabc Why did bottle disappear with no charges?

darklordabc:
That was the design brief I got from Chris they are meant to dissappear when depleted.

Haganeko:
Because you can't sell them, so there's no use for an empty bottle.

game/scripts/vscripts/events.lua
```diff
@@ -470,4 +470,4 @@ function GameMode:OnPlayerChat(keys)
     }
     CustomGameEventManager:Send_ServerToPlayer(PlayerResource:GetPlayer(playerID), "ngp_add_item", item )
   end
-end
+end
```
chrisinajar: whitespace errors can be avoided by using an editorconfig plugin

game/scripts/vscripts/modifiers/modifier_bottle_charges.lua
```diff
@@ -0,0 +1,78 @@
...
+            if self:GetStackCount() >= 1 then
+              item:SetCurrentCharges(1)
+              self:SetStackCount(self:GetStackCount() - 1)
+            else
+              item:RemoveSelf()
+              haveBottle = false
+            end
+          -- If somehow the player has done some tricky to get another bottle, remove it
+          elseif haveBottle == true then
+            item:RemoveSelf()
```
chrisinajar: absorb the charges here, also should really be checked first in the `elseif` chain.

Get a bottle with 3 charges, move it to your backpack
Get another bottle with 3 charges
Move backpack bottle into inventory, one will disappear and you lose all 3 charges.

chrisinajar:
It's appearing in the shop because we have the base shop doing normal + secret because we never changed all the secret shop items to be available at the normal shop.

darklordabc:
I thought you are going with the other bottle fix?

darklordabc:
Conflict resolved

darklordabc:
This seems redundant now

PR #388, opened 2017-03-28
Bottle and drops

Declow0:
Simple stackable bottle (fountain doesn't fill it). Stackable clarity, mango, shared tango.

Bottle applies only on caster.


darklordabc:
I've tested this and it is better than mine.
Only 3 ways that it may be inferior:
* This doesn't do the Ctrl + clicking to heal allies
* It wont capture runes if they are ever added
* And fountain refilling wont work (as you said)
But besides those 3 things, this is more visually appealing as it has the stack count on the item itself, and its technically less convoluted.

Declow0:
@darklordabc Yes, this solution disables some mechanics of bottle.

game/scripts/vscripts/items/bottle.lua
```diff
@@ -0,0 +1,51 @@
...
+function modifier_bottle_regeneration:OnCreated()
+	self.health_restore = self:GetAbility():GetSpecialValueFor( "health_restore" )
+	self.mana_restore = self:GetAbility():GetSpecialValueFor( "mana_restore" )
+	self.restore_time = self:GetAbility():GetSpecialValueFor( "restore_time" )
+end
+
+function modifier_bottle_regeneration:DeclareFunctions()
+	local funcs = {
+		MODIFIER_PROPERTY_HEALTH_REGEN_CONSTANT,
+		MODIFIER_PROPERTY_MANA_REGEN_CONSTANT,
```
chrisinajar: Needs to listen on taking damage so that you can cancel bottle regen when you take hero damage. Bottles aren't intended to be good against heroes mid-fight.

Trildar: Why even rewrite `modifier_bottle_regeneration`? Or did you already try to use the existing one and it didn't work?

Declow0: @chrisinajar I think I can delete all part about bottles now because #386 has better mechanic.

Haganeko:
#386 has the problem of not remaking the item, however, which makes this the better version, as was discussed at #coding-discussion.

PR #390, opened 2017-03-29
Item Icons

darklordabc:
I generally think that no ability or item in a custom game should ever use the same icon, even if its only slightly varied. 
If this is approved by Baumi and Chris, I'll keep adding to this PR with new icons for some of the abilities/items that are currently using duplicate graphics. If anybody wants to post an icon on this thread I'll add it to the PR, e.g. preemptive cores all have the same Icons, it'd be great if somebody could come up with 4 different icons for the upgraded versions. 
image
The two 3c icons are not currently assigned to the item as the code for the item does not yet exist. 
image
This icons have been added to the repo but not attached to any items yet as they dont exist. 
image
This icons have also been added to the repo but not attached to any items yet as they dont exist. 
image


game/scripts/npc/items/item_radiance_3.txt
```diff
@@ -1,5 +1,5 @@
-"DOTAItems"
-{
+"DOTAItems"
+{
```
warpdragon: Since they're in red, I believe that means they've been removed.

game/scripts/npc/items/item_radiance_2.txt
```diff
@@ -1,84 +1,84 @@
...
+				"blind_pct"				"17"
+			}
+			"04"
+			{
+				"var_type"				"FIELD_INTEGER"
+				"aura_radius"			"700"
+			}
+		}
 	}
 }
```
warpdragon: None of these files appear to have blank line endings.

game/scripts/npc/items/item_radiance_2.txt
```diff
@@ -1,84 +1,84 @@
 "DOTAItems"
 {
-	//=================================================================================================================
```
warpdragon: If people aren't careful with their line endings this shit can happen. In SublimeText3 you go to View > Line Endings > and make sure it says Unix, not Windows. Also you can enable EditorConfig in your SublimeText3. Furthermore there's also a way to make Git convert them automatically. I have no fucking idea. The best bet is just to check everytime you edit one. It's a bitch aint it?

game/scripts/npc/items/item_radiance_3.txt
Chronophylos: No. They need to be removed.
`^M` means CR from CRLF line endings
but we want LF line endings

warpdragon: yeah fuck those line endings. just read what i wrote below in order to remove them using Sublime Text 3.

warpdragon:
they were never cores to begin with they're reflex items. People just started calling them cores for some reason and now everyone does it.

Trildar:
Reflex Core, at least right now, technically refers to the boss item required to build a tier 1 Reflex Item. They kind of got mixed up. Probably because they're both kind of generic and I think people tend to think "Reflex Item" sounds too generic to be what they're called.

darklordabc:
I've uploaded all the reflex icons Baumi provided with the correct names and folder locations, I've only attached the ones that have item data to attach too. As we finish creating the reflex items we can hook up their equivalent icons. I think there is some icon duplication with stone skin, so we may have to replace postactive2 and postactive2b with something else.

warpdragon:
there's no duplication. stone skin has the stone arm icon.

chrisinajar:
This is looking awesome. Once you're happy with it remove WIP from the title and i'll merge em all in.

darklordabc:
Feel free to merge this, I'll make a new PR for new icons later

PR #392, opened 2017-03-29
Added -getpos command to print vector of current position

darklordabc:
Simple and straightforward tool to find what position the hero is in, useful if you want to know vector coordinates of a location.


chrisinajar:
Can you make it print to chat as well for multiplayer testing?

darklordabc:
Yeah I can just add sendcustomgamemessage

darklordabc:
Although it won't work on multiplayer yet for the v reasons we talked about.

darklordabc:
Updated

Issue #358, opened 2017-03-27
Duel location if dead when duel begins can put you back into a team fight.


KarmaticNeutral:
I think the best option is when saving a players location at the start of a duel just check if they are dead and if so set their location to be their fountain instead.

warpdragon:
>>be me. get killed in gangland fashion from a gank that will be heralded through the ages
>>go to duel, lose.
>>respawn where i died originally, proceed to relive the terror of my original death.
FeelsBadMan

PR #393, opened 2017-03-29
If Hero is dead at duel start return him to foutain

darklordabc:
This addresses this issue https://github.com/OpenAngelArena/oaa/issues/358

### DIFF
--- a/game/scripts/vscripts/components/duels/duels.lua
+++ b/game/scripts/vscripts/components/duels/duels.lua
@@ -331,6 +331,15 @@ function Duels:SavePlayerState (hero)
     mana = hero:GetMana()
   }
 
+  -- If hero is dead during start of the duel, make his saved location his foutain area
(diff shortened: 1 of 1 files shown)


game/scripts/vscripts/components/duels/duels.lua
```diff
@@ -331,6 +331,15 @@ function Duels:SavePlayerState (hero)
     mana = hero:GetMana()
   }
 
+  -- If hero is dead during start of the duel, make his saved location his foutain area
+  if hero:IsAlive() == false then
+    if hero:GetTeam() == DOTA_TEAM_GOODGUYS then
+        state.location = Vector(-5221.958496, -139.014923, 387.999023)
```
chrisinajar: this should use named zones on the map instead, however we'll do that later as a separate unit of work because merging the map is a dick in the dick.

PR #395, opened 2017-03-29
Mostly Stun duration fixes

l34Um1:


Chronophylos:
Resolved the confict

KarmaticNeutral:
Juggernaut already falls off in the late game, rather than nerfing his number of jumps to prevent him from being in the air so long would it be possible to change the bounce time trait? It's current;y 0.4 seconds per bounce which is why 40 bounces was so insane. What about something like 0.1 seconds and leave bounces with scepter at 40 so he's in the air for 4 seconds and still get the damage? Either that or buff the damage because past t2 bosses the jug ult is no longer a threat.

Haganeko:
Most of the damage from jugg ult isn't from the slashes, it's from attacks while he's slashing. Reducing the time would need us to buff his BAT while ulting, to preserve his total damage.

KarmaticNeutral:
True, what if we met in the middle? Maintain the new time (18 jumps at 0.4 sec a jump = 7.2s) and raised that to 36 jumps at 0.2 sec. This nerf cost Jug 4.6k dmg during the duration of his ult as is. ( 40 jumps at avg 212 dmg =~ 8400 || 18 jumps at the same 212 =~ 3800) So even if a larger portion of the damage is from attacking during ult, this is still too significant.

l34Um1:
I would be very surprised if Juggernaut ends up too weak because of this change. The hero is inherently very powerful. I'll consider a buff if it's needed but I don't think he needs it atm

game/scripts/npc/abilities/juggernaut_omni_slash.txt
```diff
@@ -48,7 +48,7 @@
 			"02"
 			{
 				"var_type"					"FIELD_INTEGER"
-				"omni_slash_jumps"			"3 6 9 14 32"
+				"omni_slash_jumps"			"3 6 9 12 18"
```
warpdragon: Honestly I don't think its' gonna hurt him too much. Omnislash is more about the initiation, the attack damage, and the invulnerability window. Having it last longer than it needs to is just overkill.

game/scripts/npc/abilities/magnataur_reverse_polarity.txt
```diff
@@ -38,7 +38,7 @@
 			"02"
 			{
 				"var_type"				"FIELD_INTEGER"
-				"polarity_damage"		"50 125 200 350 1000"
+				"polarity_damage"		"50 125 200 300 400"
```
warpdragon: Harsh.

game/scripts/npc/abilities/abyssal_underlord_atrophy_aura.txt
```diff
@@ -31,7 +31,7 @@
 			"02"
 			{
 				"var_type"						"FIELD_INTEGER"
-				"damage_reduction_pct"			"4 16 28 40 52 70"
+				"damage_reduction_pct"			"4 16 28 40 40 40"
```
warpdragon: Absolutely.

game/scripts/npc/abilities/magnataur_reverse_polarity.txt
l34Um1: RP would be insane while dealing 0 damage, it doesn't need 1000.

Haganeko:
About Underlord - the aura needed a nerf,  but remember it does fall off pretty hard when people start to get more and more bonus damage.
On reverse polarity - could it go 275/425? At least level 5 is not that underwhelming this way.

PR #397, opened 2017-03-29
Made Upgrade Cores Unlockable again

### LINKED ISSUE #350: Locking Boss Cores
If for any reason a boss core becomes locked from combining, you can't unlock it. 
Happened mostly when dissembling another item that included an upgrade core.

@KarmaticNeutral If you want a video, just ping me, but seemed pretty self explanatory.

Chronophylos:
This should fix #350.
#### Changes
* Added Item Selling Blacklist
* Added upgrade cores to that list
* Made upgrade cores sellable to allow unlocking


chrisinajar:
`components/shop/sell_filter` might be a better name? then we can do buy and other fancy logic in other files, scoped to shop.

PR #398, opened 2017-03-30
KV fix of damage reduction for tooltip

spar36:
On request of Warp


spar36:
Changes the Value in the Preemptive KV files to a positive value to allow the Tooltips to correctly represent damage reduction. Multiplied by -1 in the lua files to allow the lua to still function.

warpdragon:
Sweet!

PR #411, opened 2017-04-01
Add KV entries for heroes with Gold or XP talents

Trildar:
Initial entries so that these talents can be changed later. Based on Haganeko's list in #design-discussion on Discord.


Haganeko:
Still gotta add the heroes with respawn talents, I assume?

Haganeko:
I have redone those talents as well

Trildar:
Yeah. I haven't done those since I didn't have a list. There is some overlap though at least.

Haganeko:
I updated the doc... should still be somewhere around the discord, but I can reupload it if necessary.

Issue #325, opened 2017-03-25
Automatic Courier


spar36:
Have a Courier automatically spawn at the beginning of the match. 
Make it flying and boosted all the time. Basically super fast, but not to fast.

Haganeko:
I'd say don't permaboost it. We want sniping to be a thing.

PR #412, opened 2017-04-01
Add courier

Chronophylos:
Added a courier for each team on game start as requested in #325


chrisinajar:
It'd be better if there was an event added for game loaded that you could observe in init, but other than that this is great :+1:

game/scripts/vscripts/gamemode.lua
```diff
@@ -153,6 +155,7 @@ function GameMode:OnGameInProgress()
   InitModule(NGP)
   InitModule(HeroProgression)
   InitModule(SellBlackList)
+  InitModule(Courier)
```
Trildar: I think the `Courier` module is getting initialised too late for couriers to actually spawn since `OnGameInProgress` doesn't fire until minute 0 (a few seconds after heroes spawn).

Chronophylos: Oh yeah, thanks.
Looks like I forgot to commit that.

PR #413, opened 2017-04-01
Balance

l34Um1:


Chronophylos:
Before I resolve any conflict and to make sure everything is right:
What CD should Charge BKB have `20` or `10`?

PR #417, opened 2017-04-02
Add Lua Fun library

Trildar:
Separate PR for the Lua Fun stuff from #303 in case there's some issue with merging that.


chrisinajar:
redundant now, i merged in the original one

PR #418, opened 2017-04-02
Add FilterManager for compositing Filters (ExecuteOrder, Damage, etc.)

Trildar:
Requires Lua Fun from #303 or #417.

### DIFF
--- a/game/scripts/vscripts/components/filters/filtermanager.lua
+++ b/game/scripts/vscripts/components/filters/filtermanager.lua
@@ -19,7 +19,7 @@ if not FilterManager then
   FilterManager.TrackingProjectile = 10
   FilterManager.NumFilterTypes = 10
 
-  Debug.EnabledModules["filters:*"] = true
+  Debug.EnabledModules["filters:*"] = false
 end
 
 function FilterManager:Init()
--- a/game/scripts/vscripts/components/filters/filtermanager.lua
+++ b/game/scripts/vscripts/components/filters/filtermanager.lua
@@ -23,23 +23,32 @@ if not FilterManager then
 end
 
 function FilterManager:Init()
-  -- Filter function table
+  -- Registered filter function table
   self.Filters = {}
+  -- Master filter function table
+  self.MasterFilters = {}
   for i=1, self.NumFilterTypes do
+    -- Populate self.Filters with empty tables so that AddFilter can table.insert Filter entries
     self.Filters[i] = {}
+
+    -- Generate master filter functions
+    function MasterFilter(self, keys)
+      return self:RunFilterForType(keys, i)
+    end
+    self.MasterFilters[i] = MasterFilter
   end
 
   local gameMode = GameRules:GetGameModeEntity()
-  gameMode:SetAbilityTuningValueFilter(Dynamic_Wrap(self, "AbilityTuningValueFilter"), self)
-  gameMode:SetBountyRunePickupFilter(Dynamic_Wrap(self, "BountyRunePickupFilter"), self)
-  gameMode:SetDamageFilter(Dynamic_Wrap(self, "DamageFilter"), self)
-  gameMode:SetExecuteOrderFilter(Dynamic_Wrap(self, "ExecuteOrderFilter"), self)
-  gameMode:SetItemAddedToInventoryFilter(Dynamic_Wrap(self, "ItemAddedtoInventoryFilter"), self)
-  gameMode:SetModifierGainedFilter(Dynamic_Wrap(self, "ModifierGainedFilter"), self)
-  gameMode:SetModifyExperienceFilter(Dynamic_Wrap(self, "ModifyExperienceFilter"), self)
-  gameMode:SetModifyGoldFilter(Dynamic_Wrap(self, "ModifyGoldFilter"), self)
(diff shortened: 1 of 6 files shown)


Chronophylos:
Looks cool

game/scripts/vscripts/gamemode.lua
```diff
@@ -140,6 +140,7 @@ function GameMode:OnGameInProgress()
   DebugPrint("[BAREBONES] The game has officially begun")
 
   -- initialize modules
+  InitModule(FilterManager)
```
Chronophylos: This should get initialized earlier, like in [`InitGameMode()`](https://github.com/Trildar/oaa/blob/b005493ff18f02490e4cc6702018ef41a3b7f3da/game/scripts/vscripts/gamemode.lua#L168).

game/scripts/vscripts/components/filters/filtermanager.lua
```diff
@@ -0,0 +1,97 @@
...
+
+function FilterManager.ApplyFilter(keys, context, filterFunc)
+  return filterFunc(context, keys)
+end
+
+---------------------------------
+-- Master filter functions which call filter functions registered through FilterManager:AddFilter
+
+function FilterManager:AbilityTuningValueFilter(keys)
+  return reduce(operator.land, true, map(partial(self.ApplyFilter, keys), map(unpack, iter(self.Filters[self.AbilityTuningValue]))))
```
Chronophylos: Can you explain what this is doing?

game/scripts/vscripts/gamemode.lua
Trildar: Should I also move the initialisation for the other modules that add Filters? `InProgress` only fires at minute 0 which leaves a few seconds after hero spawn where the Filters aren't actually in place.

game/scripts/vscripts/components/filters/filtermanager.lua
Trildar: You mean the giant line of function calls? Going from the right side, `iter` creates an iterator for the given table; it's not too important as it's just implementation detail, just think of the other functions as looping through a table passed to it. `map` calls a function for each member in a table and returns the results (still in a table). `unpack` is to return the context object and callback function as separate values because I paired them in tables (`self.Filters[self.*]` is a table of tables). The next `map` then passes each of those context, function pairs to `self.ApplyFilter` (with the `keys` argument fixed using `partial`). That results in a table of booleans which are the results of applying all registered Filters. Finally, `reduce` effectively `and`s together all the booleans. I think `land` is "left and" or something.

I'm not that great at explaining things like this, and I typed this on a phone, so sorry if it's still confusing.

chrisinajar: If a single line needs this much description then it needs to be rewritten.

either use whitespace, comments, temporary variables, or some mixture to make this code maintainable.

game/scripts/vscripts/components/filters/filtermanager.lua
```diff
@@ -0,0 +1,97 @@
...
+  return reduce(operator.land, true, map(partial(self.ApplyFilter, keys), map(unpack, iter(self.Filters[self.ModifyGold]))))
+end
+
+function FilterManager:RuneSpawnFilter(keys)
+  return reduce(operator.land, true, map(partial(self.ApplyFilter, keys), map(unpack, iter(self.Filters[self.RuneSpawn]))))
+end
+
+function FilterManager:TrackingProjectileFilter(keys)
+  return reduce(operator.land, true, map(partial(self.ApplyFilter, keys), map(unpack, iter(self.Filters[self.TrackingProjectile]))))
+end
```
chrisinajar: with all the wizardry here i'm surprised you didn't generate these methods :)

game/scripts/vscripts/components/filters/filtermanager.lua
chrisinajar:
```lua
function FilterManager.CreateFilterIterator (filterType)
  -- create iterator on table and then unpack each entry
  return map(unpack, iter(self.Filters[filterType]))
end

function FilterManager.ApplyFilterToIterator(keys, filterIter)
  -- applyfilter(keys, entry) for each entry and return the results
  return map(partial(self.ApplyFilter, keys), filterIter)
end

function FilterManager.RunFilterForType (keys, type)
  local filterIter = CreateFilterIterator(type)
  local filterResults = ApplyFilterToIterator(keys, filterIter)

  return reduce(operator.land, true, filterResults)
end

function FilterManager:AbilityTuningValueFilter(keys)
  -- so simple, so smooth
  return self.RunFilterForType(keys, self.AbilityTuningValue)
end
```

Trildar:
Yeah, I was kind of lazy to think of a proper way to not copypasta, so I kind of copypasta'd :P
I'll fix it when I can.

game/scripts/vscripts/components/filters/filtermanager.lua
Chronophylos: `land` is logical and

Trildar:
Hopefully much neater now.

P.S. Force pushing a rebased branch does some weird things (go figure). Unsure if I should really do that in future when the branch is already in a PR.

chrisinajar:
Nope, no harm at all. I do it with my PR's constantly.

PR #419, opened 2017-04-02
Add tier 1 Postactive Reflex Item

Trildar:
Requires Lua Fun from #303 or #417.


game/scripts/vscripts/items/reflex/postactive.lua
```diff
@@ -0,0 +1,41 @@
...
+  --Applies the modifier to a test unit, purges the unit with a basic dispel affecting debuffs only,
+  --then checks if the modifier was purged (All because IsDebuff and IsPurgable don't exist in the Lua API
+  --for built-in modifiers)
+  function IsPurgableDebuff(modifier)
+    local testUnit = CreateUnitByName("npc_dota_lone_druid_bear1", Vector(0, 0, 0), false, caster, caster:GetOwner(), caster:GetTeamNumber())
+    testUnit:AddNewModifier(modifier:GetCaster(), modifier:GetAbility(), modifier:GetName(), nil)
+    testUnit:Purge(false, true, true, false, false)
+    local modifierIsPurgableDebuff = not testUnit:HasModifier(modifier:GetName())
+    testUnit:ForceKill(false)
+    return modifierIsPurgableDebuff
```
chrisinajar: 😂

chrisinajar:
10/10 is purgable hack

Haganeko:
If I cast, say, Reverse Polarity Level 5 on a hero. Does this cast it on the bear? That spell deals 400 damage IIRC. Is there a risk of this killing the bear and causing an error?
Also, will this unit spawning and killing work with duels?

Trildar:
It applies the modifier, so it doesn't do spell damage, only damage from the modifier would matter, and it should run within like 1 frame or something, so unless a modifier does a large amount of damage when applied it shouldn't be a problem. Could make some modifier for the bear to give it 100% damage resist to be safe I guess.
I wouldn't know about Duel interaction.

Haganeko:
OK, so what you're saying is we need to design some modifier which causes over 300 damage on its first frame. How much does level 5 doom do?

Trildar:
200, but you can't use items while Doomed so that's irrelevant for this case. Also, there's a typo in the Doom KV, the level 4 sceptre damage is 800...

PR #440, opened 2017-04-05
Add semistandard format to all js files, add basic test case scripts

chrisinajar:
sloppy sloppy sloppy....


chrisinajar:
The javascript and KV formats are now very strict and checked by travis automatically

Chronophylos:
Semistandard, I like it!

PR #460, opened 2017-04-06
Fixed all items with the heart update I did

KarmaticNeutral:
Fixed all items to use the correct line endings and be able to be purchased multiple times without manual building.


chrisinajar:
wow...

chrisinajar:
Well, it's not like there's a way to review this...

PR #461, opened 2017-04-07
Changed Power Treads scatter shot to splash

SquawkyArctangent:
This means that crit now applies to enemies in the radius. Includes a
quick particle.


chrisinajar:
lolomg what is this particle effect format 😂

PR #462, opened 2017-04-07
removed SLOW from tooltip on bottle.

warpdragon:
yeah bottle doesn't slow. it slows the courier but that doesn't need to be in the tooltip.


Chronophylos:
And our bottle doesn't do that anyway.

warpdragon:
NICE

PR #466, opened 2017-04-07
Compressed Load Screen Images

darklordabc:
Load screen images are way too big (like 30mg), this is mainly because the main background is png and takes up 27mg. In this PR, its converted to JPG and compressed, making it like 500kb, the other images could not be converted to JPG because they have transparent parts and JPG does not support transparency They have just been compressed. 

In all, the 30mbs have been reduced to 1mb.


Chronophylos:
Nice, this should increase loading times!

warpdragon:
>> Increase loading times
turn around 360° and walk away

Haganeko:
Chronophylos lives up to his name. Increasing loading times! What a great idea!

Chronophylos:
I don't know what you're talking about :smile:

PR #469, opened 2017-04-07
Added Greater Phase Boots with Phase Dive

SquawkyArctangent:
Pretty experiment; no farming passive, but active attacks neutral creeps that the user collides with. Feels fun, but definitely needs a look-over.


Chronophylos:
Can you make a video with the particle?

SquawkyArctangent:
https://gfycat.com/EnviousSoulfulGuillemot

Keep in mind, this is a pretty dev art particle, even moreso than the Greater Power Treads once. It could probably do with some touching up once we get around to the polish stage.

PR #478, opened 2017-04-08
New addon english update-1

warpdragon:
updated the names of reflex core items. also fixed a filename extension.


warpdragon:
Done for now!

PR #483, opened 2017-04-08
Added Items to shop

Chronophylos:
I created a file where dota reads which items to display in which category and shop. I some of our items. But because space is spare and not enough not all items a displayed correctly.
Image


Haganeko:
Remove small items from the shop, for now? Not like anyone us going to be buying bracers.

Chronophylos:
Yeah, but how do I sort it? I could put the cores to the left most column..

Haganeko:
Can you display the basic reflex core? Then clicking it would display all its upgrades...

Chronophylos:
Current Layout 
Basic Upgrades

Haganeko:
Looks good to me.

PR #500, opened 2017-04-09
Changelog

Chronophylos:
I'M #500


Trildar:
The way these version numbers are progressing, 1.0 seems pretty close. Are we just not gonna really care, or is there maybe some list of stuff we want to at least have before 1.0? Even though that's probably still not gonna be near full release ready.

Chronophylos:
`0.9.0` -> `0.10.0` 
http://semver.org/

PR #503, opened 2017-04-09
Cosmetic Improvements to Shroud of Shadows

darklordabc:
*Added two color varied icons
*Rearranged special values so that in the text, the small text comes all before the large texts
*Added "Upgradeable" to have consistency with dagon and necromincon. 
*Moved some extra info into a NOTE  because the tool-tip was too sprawling. For some reason also removed mentioning magic resistance percentages its in description, its details are shown below, no need for it to also be in top part.
*Commented out the "SHADOW ADDED MAGIC DAMAGE" line until its something other than 0% (this change not shown in image below)

_Before and after_
image


Haganeko:
I believe it is 0 by design, and actually shouldn't be a thing.

Issue #121, opened 2017-03-10
Implement OAA unique items


warpdragon:
Here's a quick rundown of the items
- [x] Refresher Core - Octarine with Refresher
- [x] Satanic Core - Octarine with Satanic
- [x] Shivas Cuirass - Shivas plus AC
- [x] Shivas Rush - Shivas plus Drums (both actives combined)
- [x] Shroud of Shadows - Make self/ally invisible + Not attackable
- [ ] Urn of Sorcery - AOE Urn that gains charges on enemy spellcast
- [ ] Vlads Helm - Vlads plus Dominator
- [x] Stoneskin Armor - Toggle to gain armor
- [x] Trumps Fists - Deso + Skadi
- [ ] Aether Glasses - Increases AOE of spells, not range
- [ ] Azazel's Crossbow - Double Calldown
- [x] Charge BKB - Magic Immunity Active charges on enemy spellcast
- [ ] Dash Staff - Push self forward, 0.1s invinc. frame, charge based CD
- [ ] Allied Euls - Cyclone self, allies, or enemies
- [ ] Jump Dagger - Jump to unit
- [x] Lucience - Switch between heal and movespeed auras
- [x] Martyr's Mail - Heal allies for % damage taken
- [ ] Rainbow Radiance - Toggle between MS/AS/dmg/Heal/Armor/Magic auras
- [ ] Pull Staff - Move self or other unit backward
- [x] Proximity Radiance - Aura that does high damage at a set distance from caster
- [x] Siege Mode - Toggle to gain splash attack (+range?), lose ability to move
- [ ] Staff of Confusion - Target enemy's spells target themselves
- [ ] Staff of Confusion - Target ally's or enemy's abilities' heal/dmg are reversed

Edit by @Chronophylos: Added checkboxes

spar36:
I was in the call today (3/11) but had to leave before the end. I know you guys were talking about a central place to store info. What on this list has been done so far?

chrisinajar:
@warpdragon can you make these checkboxes with ` * [ ] ` and remove the completed ones? That or we can just close the ticket and open granular ones later, your call

Chronophylos:
Checkboxes look cool. I added them.

Haganeko:
Refresher Core, Stoneskin Armor, Lucience, Charge BKB should be checked.

warpdragon:
Nearly half-way there! Still plenty of cool items that need creating! See above.

Rasor1911:
[...]

PR #514, opened 2017-04-09
Satanic core

Declow0:
Add implementation of Satanic Core. Issue #121.

Delete visual buff from octarine core passive.

Only 1 problem has this item: it's not unique attack modifier!


Declow0:
@chrisinajar I move information about item to separate file.

PR #517, opened 2017-04-10
Illusions having too many stats FIX 

darklordabc:
Basic logic:
Listener for unit spawns, if the spawned unit is an hero illusion, search within 100 radius for all nearby heros, and any hero that matches the illusion hero name, it copies its stats. 

Still Todo: This illusion fix only applies to illusions of yourself. It doesnt yet apply to illusions of allied or enemy illusions, like with wall of replica and disruption.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 [![All Contributors](https://img.shields.io/badge/all_contributors-29-orange.svg?style=flat-square)](#contributors)
 A really great Dota 2 game mode built by [really great people](#contributors).
 
-# Installation
+# Installation1
 This addon uses the [Dota 2 Addon Manager](https://github.com/chrisinajar/dota2-addon-manager).
 
 For detailed instructions, check [here](docs/install.md)
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 [![All Contributors](https://img.shields.io/badge/all_contributors-29-orange.svg?style=flat-square)](#contributors)
 A really great Dota 2 game mode built by [really great people](#contributors).
 
-# Installation1
+# Installation
 This addon uses the [Dota 2 Addon Manager](https://github.com/chrisinajar/dota2-addon-manager).
 
 For detailed instructions, check [here](docs/install.md)
--- a/game/scripts/vscripts/events.lua
+++ b/game/scripts/vscripts/events.lua
@@ -47,6 +47,7 @@ function GameMode:OnNPCSpawned(keys)
 
   local npc = EntIndexToHScript(keys.entindex)
 
+  -- ILLUSION HAVING WRONG STATS FIX START --
   local realHero = nil
   if npc.IsIllusion and npc:IsIllusion() and npc:IsHero() then
     -- Search nearby radius to find the real hero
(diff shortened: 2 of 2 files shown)


Trildar:
There's an `OnIllusionsCreated` event in `events.lua` that has the original entity info. But... it doesn't have references to the created illusions, because that apparently makes sense.

Trildar:
Approximation formula for stats after 25: G * 12 * ln((2*(n-13)+1)/(2*13-1)) + B + G*25, where G is the stat growth, B is the base value of the stat, and n is the level.

Haganeko:
We can calculate the stats at level n>=25 as the equivalent to level

L = 25 + 12 * (H_(n-13) - H_12)

Where H_n is the nth harmonic number, which can be approximated by

H_n = ln(n) + 1/2n - 1/12n^2 + 1/120n^4 + O(n^-6)

We can probably get away with only using up to n^2

chrisinajar:
Or you just read the hero's current stats :P

Haganeko:
You can, but you'll need to find the stats of the hero used for the illusion.

darklordabc:
My initial solution was just copying the hero stats, but that's complicated by illusions that are not illusions of the owner, I'm not sure how to find the hero that the illusion is based on, I'm sure its possible, but its probably simpler to adjust the heros stats by the initial formula we use to reduce hero stats by. Honestly both methods are probably as difficult, just depending on what knowledge we have.

darklordabc:
Okay this has been updated, it now searches within a radius for a hero which has the same hero name as the illusion and them assumes that is the real hero, works in most circumstances, it may pick the incorrect hero in very rare circumstances, e.g. if there is multiple players playing the same hero within the same 1000 radius.

Haganeko:
I don't think we have same heroes allowed, so that should be OK.

chrisinajar:
Should be able to check levels to differentiate 2 people playing the same hero. If they're the same level it doesn't really matter anyway. hp might jump around strangely, but that could probably be fixed by reading in the value before changing the stats.

Haganeko:
Does this ignore bonus stats? I assume so, but just to check.

Trildar:
`GetLevel` and the get stat functions work on illusions. Assuming `GetBaseStat` is the level 1 stat and not the stat without item/ability bonuses, I think it would be better to do use the formula method. The illusions spawn with the correct current health and mana values, so you can save those values and reapply them after adjusting stats.

darklordabc:
GetLevel didnt work when I tried it, it returned zero for the illusions. GetBaseStat might work but its the same as finding by unit name, also illusions dont spawn with correct mana and health. 
GetGold, GetCurrentXP dont work either.

Haganeko:
Would GetLevel work on the base hero?

Trildar:
I was testing it just yesterday... Make sure you have a delay between spawn event and checking the level. They don't immediately have the correct level. And illusions spawn with the same *current* health and mana as their original hero. It's only their max that's wrong.

Haganeko:
When you fix their max they'll probably lose comensurate health... that's how most attribute/health/mana changes work in DotA (armlet being the notable exception)

darklordabc:
@Trildar not having a time delay was probably the problem I was having, do you want to make a pull request improving it, otherwise I can.

Trildar:
You can go ahead. I won't have much time for another day or two.

PR #520, opened 2017-04-10
Satanic Core Icon

warpdragon:
added custom/satanic_core to npc files. added satanic core to panorama and flash3 in both items for item slot and spellicons for debuff


darklordabc:
Fyi I don't think ite necessary to add to panorama folder, afail adding to game resource folder is enough.

warpdragon:
Yeah i cant figure out why either. Theres lots of posts saying to do it tho

chrisinajar:
It seems like using the format that darklord uses worked perfectly during the test stream. Having lots of copies of the same image can be harder to maintain and result in a larger filesize for the gamemode.

works for now, perhaps we should throw the bug testers at it to confirm or something.

warpdragon:
I'm fine for deleting the ones in panorama as i agree theyre probably not needed. However, any icon used for items which have associated Buffs/debuffs need to be uploaded to items/custom as well as spellicons/custom, right?

PR #522, opened 2017-04-10
Patched a permadeath case where if people died then reconnect after duel

KarmaticNeutral:
There is a standing bug that caused permadeath in the case that a player is dead during a duel and disconnects, then the duel ends and the player returns. I figured this could be fixed by respawning that hero when they rejoin the game. If we fix the underlying cause (Duels not respawning DCed heroes) we can always remove this later. It's just 10 lines of code.

Temporary fix where if a player reconnects to the game, it checks to see if they are dead and respawns them. If we had longer respawn times I would think this could be a problem as it could be abused, but as it I think it's fine.

### DIFF
--- a/game/scripts/vscripts/events.lua
+++ b/game/scripts/vscripts/events.lua
@@ -113,6 +113,16 @@ function GameMode:OnPlayerReconnect(keys)
   OnPlayerReconnectEvent(keys)
   DebugPrint( '[BAREBONES] OnPlayerReconnect' )
   DebugPrintTable(keys)
+
+  local playID = keys.PlayerID
+  if not playID then return end
+
+  local hero = PlayerResource:GetSelectedHeroEntity(playID)
+
+  hero:SetRespawnsDisabled(false)
+  if not hero:IsAlive() then
+    hero:RespawnHero(false,false,false)
+  end
 end
 
 -- An item was purchased by a player


Haganeko:
Check how long their respawn time is before respawning them, maybe?

game/scripts/vscripts/events.lua
```diff
@@ -113,6 +113,16 @@ function GameMode:OnPlayerReconnect(keys)
   OnPlayerReconnectEvent(keys)
   DebugPrint( '[BAREBONES] OnPlayerReconnect' )
   DebugPrintTable(keys)
+
+  local playID = keys.PlayerID
+  if not playID then return end
```
chrisinajar: make this 3 lines
```lua
if not playID then
  return
end
```

KarmaticNeutral: I will make these changes when I get home from work.

KarmaticNeutral:
Currently, they would not be able to continue fighting in the duel. They would respawn in base so you could do that in order to continue farming.

Haganeko:
Check if there is a duel currently ongoing, and check their respawn timer - I think it behaves differently depending on whether one died in a duel or outside of it.

KarmaticNeutral:
checks to see if there is currently a duel when respawning on reconnect. Tested and working.

PR #527, opened 2017-04-11
Cleaning duels

Chronophylos:
I don't know how to make it accept classes and not base classes.
All blacklisted entities get deleted in an radius around the duel arena. Places that don't overlap get omitted and entities get deleted.
And I disabled Debugging of some Modules

### DIFF
--- a/game/scripts/vscripts/components/boss/ai.lua
+++ b/game/scripts/vscripts/components/boss/ai.lua
@@ -4,7 +4,7 @@ if BossAI == nil then
   DebugPrint ( 'creating new BossAI object' )
   BossAI = class({})
 
-  Debug.EnabledModules['boss:*'] = true
+  Debug.EnabledModules['boss:*'] = false
 end
 
 BossAI.IDLE = 1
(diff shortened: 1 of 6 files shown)


KarmaticNeutral:
Techies remote mines were the original problem that led us to this solution. Those at least need to be uncommented. I also believe that we wanted to remove wards and enchanted treant protectors trees.

Looks great though overall

spar36:
I would think that we remove everything from the duel arenas. They should all be a completely new fight.

warpdragon:
I'm afraid that any solution to this problem will prove to be arbitrary and inelegant. Cleaning the arena is an arbitrary nerf to trap-based heroes like Aghs-Treant, Techies, etc. It's a similar situation with stack-reliant spells like Atrophy Aura and Essence Shift. Keeping stacks at the beginning of a duel is a buff, while removing stacks would constitute a nerf. But no one solution is "correct". It's all just an attempt to make duels "work".

Haganeko:
There is no possible response to an arena spawn point left full of mines by a techies. Treant aghs is less clear cut, but again there isn't much possibility people can go and clean up the arena, since it's locked off

chrisinajar:
If the tree eyes can't just be removed then you could use the respawn tree method, for each tree you could call that to destroy the eyes. I haven't tested it at all.

Chronophylos:
@KarmaticNeutral These are base classes - `"npc_dota_techies_mines"` contains `"npc_dota_techies_land_mine"` and `"npc_dota_techies_remote_mine"`.

PR #537, opened 2017-04-12
Map Update

warpdragon:
put boss arena entrance closer to team spawn. push central camps further away. fixed camp neutral ground overlays. illustrated the paths between camps better. switched mid lvl 3 camps and first lvl 2 camps on each side.


warpdragon:
Video here:  https://www.youtube.com/watch?v=qldGdCyMD8s

Issue #538, opened 2017-04-13
Implement Farming Cave


Haganeko:
The farming cave is a sequence of four rooms located past a team's fountain. The creeps there don't respawn automatically - they do so only after the whole cave is cleared. Their power, and the amount of gold and experience they give, depend on how many times the cave has been cleared. Note that the cave has only four rooms - the first space is only an entrance and has no creeps.

Tasks:
- [x] Implement barriers so each room can only be accessed after the previous has been cleared.
- [x] Implement a teleport back to fountain for once the cave has been cleared.
- [x] Implement a creep respawn handler which can respawn the creeps once all the creeps are killed and all heroes have gone out.
- [x] Implement gold, exp and creep power progression as outlined below.
- [x] Implement additional 1.5^(t-60) XP multiplying factor for creeps spawned after minute 60.

Sometimes the values are not integers. The decimal parts are relevant for the multipliers.
Notice the NOTE under room four.

### Room One: Howl's it Going?
Ten Alpha Wolves. Should be killable at level 1+8k, after the cave has been cleared k times.
Base worth: 38 gold, 120 XP
Multiplier: (16k+5)/5 for gold, (168k^2 + 2k + 15)/15 for XP, after the cave has been cleared k times.

### Room Two: Horse Tomatina
Five Hellbear Smashers, Five Centaur Conquerors. Should be killable at level 3+8k, after the cave has been cleared k times.
Base worth: 68.4 gold, 123.5 XP
Multiplier: (16k+9)/9 for gold, (84k^2 + 43k + 13)/13 for XP, after the cave has been cleared k times.

### Room Three: Draggin' it Around
Eight Black Dragons. Should be killable at level 5+8k, after the cave has been cleared k times.
Base worth: 136.5 gold, 167.25 XP
Multiplier: (16k+13)/13 for gold, (84k^2 + 85k + 29)/29 for XP, after the cave has been cleared k times.

### Room Four: Roashes Everywhere
Two Mini Roshans. Should be killable at level 7+8k, after the cave has been cleared k times.
Base worth: 646 gold, 753.5 XP
Multiplier: (16k+17)/17 for gold, (56k^2 + 85k + 37)/37 for XP, after the cave has been cleared k times.
NOTE: Gold from these should be split across the team.

chrisinajar:
ZoneControl API can be used for locking players in/out of the farming zone areas.

Haganeko:
[...]

PR #547, opened 2017-04-15
Added Farmingcave

Chronophylos:
As described in #538
Additional Changes:
* Added Spawns and Zones to the map *while hoping no one changes the map as well*
* Added Baby Rosh Mini Boss
* Each time the Cave has been cleared Baby Rosh gets smaller
* Added Teleport Animations for all Units inside the cave

Creeps are **not** balanced and have **no** progression!
I don't know what stats to give them.
*(I think room 3 is currently weaker than room 2)*

Known (not tested) potential Bugs
* You *can* get into the cave after it has been cleared and before it got reset and get stuck between two rooms
* You *can* get stuck between two rooms (zones are overlapping but still)


warpdragon:
"Each time the Cave has been cleared Baby Rosh gets smaller" I support this.

Chronophylos:
`newRoshSize = (0.01 * cave.timescleared)`

warpdragon:
once this is committed i can apply baumi's map changes from this morning.

PR #549, opened 2017-04-15
Add audiovisual effects for Greater Boots of Travel teleport

Trildar:
The effects are still a little bit off. Particularly one of the circles for the destination effect moves around weirdly when the target moves.


Trildar:
Put in the minimap circle thingy as well. Sadly doesn't move with moving targets. Not sure if that's even possible without maybe some annoying client side code.

P.S. `return false` in `OnSpellStart` doesn't seem to actually stop channeling spells from starting, which results in self-casts simply teleporting to the last target.

PR #553, opened 2017-04-15
Trumps fists update

lindleya1:
Tried to fix trumps fists. Removed the skadi and desolator effects from
the item, and edited the tooltip. For some reason the tooltip still says
that it reduces armor and slows movement, and I can't seem to change
that


lindleya1:
hey, can you just merge this so i can fix my thing?

Honeth:
He accidentally pushed a change to master so if this pr gets merged he can start using branches properly.

Trildar:
@lindleya1 It's probably still showing because you didn't edit the main addon_english.txt. Editing the individual files doesn't actually change the tooltips in-game. Chris compiles from the individual files whenever he uploads Workshop versions, but I think it's been a while since he's pushed an updated addon_english.txt into the git repo. Otherwise, the armor reduction and slow already wouldn't be showing because they were commented out in an earlier PR. (The double slash `//` in KV means comment)

PR #558, opened 2017-04-15
Fix Baby Rosh in Farming Cave beeing too smol

Chronophylos:
I can do fancy plots as well https://www.wolframalpha.com/input/?i=0.25+%2F+(0.5+*+(x+%2B+1))


Haganeko:
So... 1/(2x+2) ?

Chronophylos:
Yeah but I set Baby Roshs Size in KV to 0.25.
How do you do your plots?
I want to play around with these numbers

Haganeko:
I mostly use wolfram alpha or Google spreadsheet. If it's really complicated I'll cracked open a Mathematica notebook.

Issue #519, opened 2017-04-10
Update abilities for parity with 7.05


Trildar:
Ability KVs were written based on 7.03, so changes from patches 7.04 and 7.05 need to be reviewed and merged into OAA. If anyone is able and willing to make a list of the necessary changes, that could make things easier. Only values for ability levels existing in normal Dota should be changed (correct me if I'm wrong).

Chronophylos:
We need a script to do that in the future as well

onc:
maybe i could help on this one... do you prefer any programming language for the script? do you need a graphical user interface for merging the changes or are you ok with having a command-line tool?

Chronophylos:
I don't think so Chris has been using JS and I like Python a lot.
We already have some JS scripts, try to do that.
no GUI.

Haganeko:
Gotta make sure it doesn't touch custom stuff, though.

spar36:
I'll be doing work on the talents this wknd, so I will do all the parity with talents then.

Trildar:
@onc Chris also made a simple JavaScript KV parser called parse-kv. I'm not sure it's good enough for this purpose, but do take a look at it. Chris said he wanted to improve it at some point, but I don't think he's found the time yet. Perhaps you could even contribute to that.

onc:
thank you for the hint to the parse. i also found your guide on how to handle kv files. 
so just to make sure we are talking about the same goal right here. 

you need a script to update ability stats on new dota releases. 
if i understand your file structure these files are included in `oaa/game/scripts/npc/npc_abilities_override.txt`.

i found [this](https://github.com/dotabuff/d2vpkr) repository which (i think) contains up-to-date data from original dota, including KV's.

so should we try to write a script, which is able to parse this file
https://raw.githubusercontent.com/dotabuff/d2vpkr/master/dota/scripts/npc/npc_abilities.txt
and brings updates into your KV's?

Chronophylos:
Yes that is what is should do. Dotabuffs Repo  is always up to date.

Trildar:
Done long ago. Already have parity with 7.06c I believe. Would still be nice to have that script at some point though.

Issue #475, opened 2017-04-08
Update Talents


Haganeko:
We have a bunch of talents which need to be changed. List of what needs changing attached.
[Talents.pdf](https://github.com/OpenAngelArena/oaa/files/907144/Talents.pdf)

spar36:
@Trildar I was looking at the talents to see how easy it would be for me to put them in this weekend.

Did you hand make them, or find them somewhere? We will also need all the heroes that have a respawn time talent.

*Plus if it is easy, having all the heroes would be nice 1) to see what all the talents look like in KV, and 2) if we want to change any others later, the files would be there.

Haganeko:
If we do the talents as they are here, some will take serious work. Do the easy ones for now - balance and design shifts may change these.

Trildar:
@spar36 I've not actually changed anything myself yet. I just put in the hero KVs with their talents so that they can be changed later. They all currently point to their default talents. Vanilla Talent KVs can be found [here](https://raw.githubusercontent.com/dotabuff/d2vpkr/master/dota/scripts/npc/npc_abilities.txt).

As for adding all heroes. That's probably scriptable, but I'm not too sure how to do it. Plus, I don't personally see much advantage in having all hero KVs in there just for the sake of having them available, mainly because I think it makes things more annoying when vanilla Dota changes hero's Talents. The KVs themselves aren't particularly interesting, just ability lists. The Talent KVs usually just have a single value in their AbilitySpecial.

Most if not all the Talents that change specific ability values will probably require rewriting the abilities. The more generic stuff like +stats and +spell amp can reuse the existing vanilla Talents, and possibly use `BaseClass` for creating ones with values that don't exist in normal Dota.

spar36:
Here is the link to a spreadsheet that says what is done.
All of the "add" heroes can be done, just will take time to put them in by hand.
All "special" heroes will require an ability to be created.

https://docs.google.com/spreadsheets/d/1l7yKkpZlyLGWE-gyBBwEjGVK8InvrezDPFy8UPXfisk/edit?usp=sharing

Haganeko:
[Talents706.pdf](https://github.com/OpenAngelArena/oaa/files/1006113/Talents706.pdf)
Updated talents table, with all the RT talents removed, and the techies problem fixed.  Also moved one or two of the RT talents into Gold or XP talents when they were of the same level, because the RT substitute was cooler.

Haganeko:
[...]

PR #562, opened 2017-04-16
Talents Update

spar36:
#475 
These are all the talents that we changed to built in dota talents. Will need to make any talents that are not built in. Denoted by special in the below spreadsheet.

https://docs.google.com/spreadsheets/d/1l7yKkpZlyLGWE-gyBBwEjGVK8InvrezDPFy8UPXfisk/edit?usp=sharing

*Added //OAA comment to the end of any line that is a talent we changed.

#519 
Updated all talents to parity with 7.05


Haganeko:
You finished all the non special ones? Spar, you're a beast.

PR #564, opened 2017-04-16
Add proper ability level dot styling for heroes with 5 or 6 abilities

Trildar:
Also restructured ability level dot styler to handle different styles based on number of abilities.

P.S. Will I ever be done with these dumb dots? XD
...still need to figure out how to handle Invoker's 100 billion ability levels, even if that's only done when we maybe have custom UI.

### DIFF
--- a/content/panorama/scripts/custom_game/leveldots.js
+++ b/content/panorama/scripts/custom_game/leveldots.js
@@ -10,33 +10,39 @@
   });
 }());
 
-function InjectAbilityDotsStyle (abilitiesPanel, dotStyles, containerStyles) {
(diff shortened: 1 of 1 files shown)


chrisinajar:
What's wrong with it now?

Trildar:
The HUD apparently doesn't keep the same horizontal space per ability icon when a hero has more than 4 abilities, so the dots need to be even more squished together to show them all properly. Invoker just has too many ability levels to ever show them all. Interestingly, Dota doesn't even assign a proper CSS class to indicate how many ability levels there are.

PR #575, opened 2017-04-16
…Awesome map changes

warpdragon:
switched 2 camps. changed the boss arena layout to go counter clockwise not clockwise. updated texturing for sharpness. enhanced final boss room entrance in farming cave


warpdragon:
Video: https://youtu.be/127ejnO1iIM

PR #580, opened 2017-04-17
Tooltip Refresher Core, Power Treads, Phase Boots

imaGecko:
Added missing information to items: Refresher Core Active, Greater Phase Boots Passive, Greater Power Treads Passive


warpdragon:
Nice! Keep it up!

warpdragon:
This is what they're supposed to do. Can merge this once the ability ahs been fixed.

Haganeko:
The redesign didn't go through, then?

imaGecko:
It was a small communication error. It's already updated to their current function.

Trildar:
From what I know, the split shot is outdated in any case. The current plan is either flat bonus damage against creeps, or what it is now, but possibly without on-hit effects or something so that it's not so ridiculous.

game/scripts/npc/items/farming/item_greater_phase_boots.txt
```diff
@@ -109,21 +109,21 @@
...
+			//{
+			//	"var_type"				"FIELD_INTEGER"
+			//	"projectile_total"		"4 4 4 4 4"
+			//}
+			//"13"
+			//{
+			//	"var_type"				"FIELD_INTEGER"
+			//	"projectile_range"		"700"
+			//}
 			"14"
```
chrisinajar: 8, 9, 10, 14....

imaGecko:
Fixed. Also added the missing tooltip for the tiny ultimate. Was as far as I'm aware the only hero ability that was missing one.

warpdragon:
Ok so from the looks of it, the requested changes have been made, and the tooltips for phase boots are now correct (greater phase instead of split shot).

PR #583, opened 2017-04-17
Various cleanup of (Greater) Boots of Travel KVs

Trildar:
Fun fact: The Boots of Travel 2 recipe was configured such that Boots of Travel would basically just spontaneously upgrade themselves without any additional items.


chrisinajar:
I exploited that so much, RIP awesome bug.

We should 1/2 the cost because even with instant upgrade they were almost too expensive to be worth it.

Haganeko:
We had each recipe at 1000 gold. That was too much? Hopefully removing exponential gold will help make them worth it

chrisinajar:
I exploited that bug a lot.

BOTs are already probably the weakest boots, we should consider halving the recipe cost so that this change isn't a nerf. It would break our rule of not changing base items, but you need to upgrade them before you can get the farming boots, and they're designed for supports, so it just makes sense.

Haganeko:
What. We'd already broken that rule when we designed our BoT's. Normally they cost 2000 gold per recipe, and we changed that to 1000...or did someone "fix" deliberate design decisions again?

warpdragon:
we need to put control towers in the game for this item's default teleport ability to be remotely viable. That being said: go go control towers!

warpdragon:
OR. taking from LoL some more, make it where you can teleport to wards!

PR #590, opened 2017-04-18
Courier Fix

darklordabc:
This adds a 1 second timer to courier spawning. It checks if there is an item blocking first slot in inventory, if so, sells the item (players get full refund), adds the courier, casts it immediately, adds the flying courier and casts that immediately.

### DIFF
--- a/game/scripts/vscripts/components/courier/courier.lua
+++ b/game/scripts/vscripts/components/courier/courier.lua
@@ -15,8 +15,25 @@ function Courier:Init ()
 end
 
 function Courier:SpawnCourier (hero)
-  DebugPrint("Creating Courier for Team " .. hero:GetTeamNumber())
-  courier = hero:AddItemByName('item_courier')
-  flying = hero:AddItemByName('item_flying_courier')
-  Courier.hasCourier[hero:GetTeamNumber()] = true
+  Timers:CreateTimer(1, function ()
+    DebugPrint("Creating Courier for Team " .. hero:GetTeamNumber())
+
+    -- Check if there is an item blocking slot 1, if so sell it
+    slot1Item = hero:GetItemInSlot(0)
+    if slot1Item then
+      hero:SellItem(slot1Item)
+    end
+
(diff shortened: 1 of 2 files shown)


game/scripts/vscripts/components/courier/courier.lua
```diff
@@ -15,8 +15,25 @@ function Courier:Init ()
...
+    -- Create couriers and then cast them straight away
+    local courier = hero:AddItemByName('item_courier')
+    if courier then
+        hero:CastAbilityImmediately(courier, hero:GetPlayerID())
+    end
+    local flying = hero:AddItemByName('item_flying_courier')
+    if flying then
+        hero:CastAbilityImmediately(flying, hero:GetPlayerID())
+    end
+
```
spar36: Could we then do `hero:BuyItem(slot1Item)` ? I get that this works, but kinda bad user design to just have items disappear from a players inventory.

darklordabc: You also have to account for charges on an item.

Haganeko: Could also have a dummy hero spawn the courier instead of the first player...

chrisinajar:
Is it possible to do with a dummy unit instead of a real hero? That'd be much simpler on the items...

Honeth:
should be possible i believe it might even be possible if we give the fountain an inventory to use the fountain instead of anything else

warpdragon:
Hell yeah!

warpdragon:
Ok I'm confused. Wouldn't ```function GameMode:InitGameMode()
GameMode = self
InitModule(Courier)```
put the Courier in their inventory before they even pick their teams because of InitGameMode()?

chrisinajar:
It'd probably throw errors since the heroes don't exist yet, could give it a try though.

Chronophylos:
No `SpawnCourier()` gets called when a Hero spawn in-game.

PR #594, opened 2017-04-18
Implement Preemptive 3c Reflex Item (Bubble Orb)

Trildar:
Could possibly randomly let non-targeted spell effects through. I wasn't really able to determine the cause. Sometimes Lina Light Strike Array would somehow stun even though the Lina was outside the bubble.

Also some notes about the current implementation:
- Does not block the Meat Hook dragging (damage is blocked)
- Seems to sometimes still display visual effects for a split second (such as screenshake from stun)
- Does not prevent effects from non-targeted spells from being refreshed (such as getting hit by Light Strike Array while still stunned from LSA)


Trildar:
I'm somehow constantly getting "postactive" and "preemptive" mixed up it seems.

PR #598, opened 2017-04-18
lmao

l34Um1:


l34Um1:
yeah did a test game and its super broken lol. but I think the in between is exactly what we want

chrisinajar:
Whole team standing in macropyre, nothing wrong here.

PR #600, opened 2017-04-19
Added "Stop fighting yourself" Boss

Chronophylos:
I just want to get that #600.


Haganeko:
Please alter the internal name; we already have a "Mirror Boss".

Chronophylos:
Sure l, although "stop fighting yourself" is pretty long.

Haganeko:
Does "Karma" work for you? How about "Reaper" (as in "you reap what you sow")?

Chronophylos:
Eh, I went with the long name.
I removed some repetition on file names.
Maybe I'll rename him again.

warpdragon:
Mirror is just the name of the debuff from what i can tell

warpdragon:
Also 600 hype!

Chronophylos:
I renamed it.

Chronophylos:
so many commits..

KarmaticNeutral:
I thought my boss PR had a lot of commits lol

Chronophylos:
I still have some Problems with my AI...
I don't really know what to do :D

Haganeko:
"Neutrals don't need a courier" 
What.

Chronophylos:
If a hero spawns a courier gets spawned. Even if the hero is of team `DOTA_TEAM_NEUTRALS`.

game/scripts/vscripts/abilities/stopfightingyourself/dupe_heroes.lua
```diff
@@ -0,0 +1,145 @@
...
+      if caster.illusions == nil then
+        caster.illusions = {}
+      end
+      local index = #caster.illusions + 1
+      caster.illusions[index] = illusion
+
+      illusion:OnDeath(function()
+        -- create particle
+        -- NOTE I have yet to see those particles
+        Timer:CreateTimer(0.8, function()
```
Trildar: Don't know if you've fixed this already, but should be `Timers`.

game/scripts/vscripts/abilities/stopfightingyourself/dupe_heroes.lua
```diff
@@ -0,0 +1,145 @@
...
+      -- NOTE this doesn't seem to work
+      if math.random(100) <= 10 then
+        EmitAnnouncerSoundForPlayer('sounds/vo/announcer_dlc_rick_and_morty/generic_illusion_based_hero_02.vsnd', unit:GetPlayerID())
+      end
+
+      if caster.illusions == nil then
+        caster.illusions = {}
+      end
+      local index = #caster.illusions + 1
+      caster.illusions[index] = illusion
```
Trildar: This can be done simply with `table.insert(caster.illusions, illusion)`, but since you need the index for later anyway, using `table.insert` may not be that much better.

game/scripts/vscripts/abilities/stopfightingyourself/dupe_heroes.lua
```diff
@@ -0,0 +1,145 @@
...
+
+      illusion:OnDeath(function()
+        -- create particle
+        -- NOTE I have yet to see those particles
+        Timer:CreateTimer(0.8, function()
+          ParticleManager:CreateParticle('particles/generic_gameplay/illusion_killed.vpcf', PATTACH_ABSORIGIN, illusion)
+        end)
+        illusion:RemoveSelf()
+        caster.illusions[index] = nil
```
Trildar: This has potential to create holes in the `illusions` table. i.e. you can end up with indices 1, 2, 4, and 5 being defined with 3 missing. This doesn't play well with `ipairs` which will only iterate through 1 and 2 in that case. You should use `table.remove(caster.illusions, index)` instead.

game/scripts/vscripts/abilities/stopfightingyourself/dupe_heroes.lua
Chronophylos: I couldn't think of an easy and fast way to delete them afterwards.

Chronophylos:
I rebased it and fixed some things

game/scripts/vscripts/abilities/stopfightingyourself/dupe_heroes.lua
```diff
@@ -0,0 +1,145 @@
...
+      caster.illusions[index] = illusion
+
+      illusion:OnDeath(function()
+        -- create particle
+        -- NOTE I have yet to see those particles
+        Timers:CreateTimer(0.8, function()
+          ParticleManager:CreateParticle('particles/generic_gameplay/illusion_killed.vpcf', PATTACH_ABSORIGIN, illusion)
+        end)
+        illusion:RemoveSelf()
+        table.remove(caster.illusions, index)
```
Trildar: Just realised that this would actually break pretty easily as well, since the indices won't be fixed. Possible solutions include:
- Searching for the correct illusion instance to delete using `index` (this could possibly result in rare race conditions if illusions die at the same time)
- Doing garbage collection in the boss thinker instead by deleting illusions that are no longer alive
- Going back to what you had before and using `pairs` to loop through the illusions instead. Not sure if this has any other potential issues

game/scripts/vscripts/abilities/stopfightingyourself/debuff_mirror.lua
```diff
@@ -0,0 +1,106 @@
...
+
+    -- Tests if given modifier is a debuff and purgable with a basic dispel
+    --  Applies the modifier to a test unit, purges the unit with a basic dispel affecting debuffs only,
+    --  then checks if the modifier was purged (All because IsDebuff and IsPurgable don't exist in the Lua API
+    --  for built-in modifiers)
+    function IsPurgableDebuff(modifier)
+      local testUnit = CreateUnitByName("npc_dota_lone_druid_bear1", Vector(0, 0, 0), false, caster, caster:GetOwner(), caster:GetTeamNumber())
+      testUnit:AddNewModifier(testUnit, nil, "modifier_purgetester", nil)
+      testUnit:AddNewModifier(modifier:GetCaster(), modifier:GetAbility(), modifier:GetName(), nil)
+      testUnit:Purge(false, true, true, false, false)
```
Trildar: Should probably set the last two flags to true as well (`bRemoveStuns` and `bRemoveExceptions`) so that it tests like a strong dispel.

game/scripts/vscripts/abilities/stopfightingyourself/debuff_mirror.lua
```diff
@@ -0,0 +1,106 @@
...
+    --print(modifier:GetCaster():GetName())
+    --print(modifier:GetAbility():GetAbilityName())
+    --print(modifier:GetName())
+    --print('--')
+    local duration = modifier:GetDuration() * self:GetAbility():GetSpecialValueFor('duration_multiplier')
+    local stacks = modifier:GetStackCount()
+    local newModifier = target:AddNewModifier(attacker, modifier:GetAbility(), modifier:GetName(), {
+      duration = duration,
+      duration_ranged = duration,
+      duration_melee = duration,
```
Trildar: I would have assumed the `duration` key always applied to everything, but if you've already tested this to be required then I suppose I can't argue.

game/scripts/vscripts/abilities/stopfightingyourself/debuff_mirror.lua
```diff
@@ -0,0 +1,106 @@
...
+      return modifierIsPurgableDebuff
+    end
+
+    return IsPurgableDebuff(modifier)
+  end
+
+  --print('-------')
+  local modifiers = caster:FindAllModifiers()
+  --each(function(x) print(x:GetName()) end, iter(modifiers))
+  local purgableDebuffs = filter(IsDebuff, iter(modifiers))
```
Trildar: Not technically correctly named since this isn't exclusively purgable stuff, but not that important I suppose.

game/scripts/vscripts/abilities/stopfightingyourself/dupe_heroes.lua
```diff
@@ -0,0 +1,145 @@
...
+  for _,unit in ipairs(units) do
+    if unit:IsRealHero() and unit ~= caster then
+      self:StartCooldown(self:GetSpecialValueFor('cooldown'))
+
+      local illusion = CreateUnitByName(
+        unit:GetName(),
+        unit:GetAbsOrigin(),
+        true,
+        caster,
+        caster,
```
Trildar: [This](http://moddota.com/forums/discussion/62/illusion-ability-example) says that the unit owner should be `nil`, but since this doesn't seem to crash the game, it could be outdated information.

game/scripts/vscripts/abilities/stopfightingyourself/dupe_heroes.lua
```diff
@@ -0,0 +1,145 @@
...
+    self:GetAbilityTargetType(),
+    self:GetAbilityTargetFlags(),
+    FIND_ANY_ORDER,
+    false
+  )
+
+
+  for _,unit in ipairs(units) do
+    if unit:IsRealHero() and unit ~= caster then
+      self:StartCooldown(self:GetSpecialValueFor('cooldown'))
```
Chronophylos: I'll do that. Although it doesn't matter as this is not player controlled and the AI checks for heroes in the area

game/scripts/vscripts/abilities/stopfightingyourself/dupe_heroes.lua
Chronophylos: I could create a time to do garbage collection

game/scripts/vscripts/abilities/stopfightingyourself/debuff_mirror.lua
Chronophylos: I just copied code

game/scripts/vscripts/abilities/stopfightingyourself/debuff_mirror.lua
Chronophylos: It didn't work for skadi. Can't remember if this fixed it.

game/scripts/vscripts/abilities/stopfightingyourself/dupe_heroes.lua
Chronophylos: I'm not sure what the effect of npc_owner and unit_owner.

Chronophylos:
@chrisinajar can you review this?

game/scripts/vscripts/abilities/stopfightingyourself/dupe_heroes.lua
```diff
@@ -0,0 +1,148 @@
...
+      illusion:SetRespawnsDisabled(true)
+
+      -- Set Stats to original hero
+      illusion:SetHealth(unit:GetHealth())
+      illusion:SetMana(unit:GetMana())
+
+      -- Randomly play sound to player (10% chance)
+      -- NOTE this doesn't seem to work
+      if math.random(100) <= 10 then
+        EmitAnnouncerSoundForPlayer('sounds/vo/announcer_dlc_rick_and_morty/generic_illusion_based_hero_02.vsnd', unit:GetPlayerID())
```
chrisinajar: You probably need to preload it in the KV

game/scripts/vscripts/libraries/garbagecollection.lua
```diff
@@ -0,0 +1,23 @@
+
```
chrisinajar: Remove this, it's not needed and enforces bad habits.

game/scripts/vscripts/abilities/stopfightingyourself/dupe_heroes.lua
```diff
@@ -0,0 +1,148 @@
...
+      end
+
+      if caster.illusions == nil then
+        caster.illusions = {}
+      end
+      table.insert(caster.illusions, illusion)
+
+      illusion:OnDeath(function()
```
chrisinajar: You can remove the illusion entry from within this handler so that you don't need that race-condition-powered-garbage-collector.

game/scripts/vscripts/units/stopfightingyourself.lua
```diff
@@ -0,0 +1,194 @@
...
+
+  ABILITY_dupe_heroes = thisEntity:FindAbilityByName('boss_stopfightingyourself_dupe_heroes')
+
+  thisEntity.GC = GarbageCollection:Add(thisEntity.illusions, 10)
+end
+
+function Think(state, target)
+  -- NOTE: I'm thinking too long
+  -- BUG: Not attacking
+  -- BUG: illusions do not get deleted correctly
```
chrisinajar: Illusions like to stay around in memory and then get reused when another illusion is created. It's a dota thing and it can cause lag in overly illusion heavy games. I think we're probably fine, and we can always tweak the max illusions if it becomes an issue.

game/scripts/vscripts/units/stopfightingyourself.lua
```diff
@@ -0,0 +1,194 @@
...
+        OrderType = DOTA_UNIT_ORDER_MOVE_TO_POSITION,
+        Position = GLOBAL_origin, --Optional.  Only used when targeting the ground
+        Queue = 0 --Optional.  Used for queueing up abilities
+      })
+      return 5
+    end
+  end
+
+  if thisEntity:IsIdle() and IsHeroInRange(thisEntity:GetAbsOrigin(), 900) then
+    UseRandomItem()
```
chrisinajar: I've only ever seen people assign queue to a number, if you're having problems with only the last thing you run working or only the one you put a 0 in working, i would change this to instead either queue up sequentially or fall through conditional if it worked so that they don't stomp over each other.

game/scripts/vscripts/abilities/stopfightingyourself/dupe_heroes.lua
Chronophylos: I thought can only precache "soundevent" not "sound".

game/scripts/vscripts/libraries/garbagecollection.lua
Chronophylos: There are a lot of files that do this though...
I'll add some comments about usage.

game/scripts/vscripts/units/stopfightingyourself.lua
Chronophylos: I already set a cap at 32. And technically I'm not using illusions but heroes of team `DOTA_TEAM_NEUTRALS`.

game/scripts/vscripts/units/stopfightingyourself.lua
Chronophylos: I'll change this so one of the three will be done and not all of them sequentially.

game/scripts/vscripts/abilities/stopfightingyourself/dupe_heroes.lua
Trildar: He was doing that before, but the issue was that doing that would either cause `caster.illusions` to not be array-like, i.e. it would sometimes have in-between indices undefined like only keys 1, 2, 4, and 5 defined, with 3 being `nil` (making it unusable with `ipairs`). Or, you would need to search for the index of an illusion in the table every time you wanted to remove it from `caster.illusions`, which I was concerned could result in race conditions because removing entries would shift the indices.

game/scripts/vscripts/libraries/garbagecollection.lua
chrisinajar: No, I mean the whole file and the functionality that it does.

game/scripts/vscripts/abilities/stopfightingyourself/dupe_heroes.lua
chrisinajar: You can either index by entid or use the built in lua functions for making a table act like an array. Both options would work.

Trildar: If you mean `table.insert` and `table.remove`, the issue is that `table.remove` shifts indices, so illusions won't have stable indices in `caster.illusions`. Using entindex would make the table non-array-like still. An array in Lua needs to have consecutive numeric keys, starting from 1. But using entindex and just abandoning having `caster.illusions` be an array might just be better. Then `pairs` has to be used instead of `ipairs`

chrisinajar: Yup, both would work and would be much better than the current solution in both performance and reliability (current version is just a giant race condition). I vote entindex.

game/scripts/vscripts/abilities/stopfightingyourself/dupe_heroes.lua
```diff
@@ -0,0 +1,147 @@
...
+        illusion:EmitSound('sounds/vo/announcer_dlc_rick_and_morty/generic_illusion_based_hero_02.vsnd')
+      end
+
+      if caster.illusions == nil then
+        caster.illusions = {}
+      end
+      caster.illusions[illusion:entindex()] = illusion
+
+      illusion:OnDeath(function()
+        -- create particle
```
chrisinajar: pop a `caster.illusions[illusion:entindex()] = nil` in here.
I think we should also do `cast.illusions.count = cast.illusions.count + 1` when adding in the illusion entry and then `- 1` here... That would allow us to use the count variable in the above conditional to enforce max size.

game/scripts/vscripts/abilities/stopfightingyourself/dupe_heroes.lua
Chronophylos: entindex and `pair`?

Chronophylos:
rebased again

Chronophylos:
Ready to merge

PR #603, opened 2017-04-19
balance

l34Um1:


Haganeko:
I also feel that giving it a coat in cores would be truer to the concept of the rapier. I mean, that's a lot of money, but long games will get flooded with the things as is.

l34Um1:
Rapier has a recipe from what I can tell. We may need to increase the cost but I like it being this high power level item that you can just buy but its super expensive. If 100k is too cheap we can always put it at 500k or something

warpdragon:
Very helpful housekeeping changes here. Thank you for taking the time. Also rapier lul

warpdragon:
Also i think the way it works is that everything has a recipe when its cost is higher than the combined cost of its components.

spar36:
Yeah, the recipe is how we will do balance concerning the cost of the item. it is a single value that is easy to change to change the overall cost of the resulting item.

Haganeko:
I goofed out. He's changing the recipe cost, not the item cost. It should work.

Rachol:
You are not able to buy rapier as it is....

Single shop item currently cannot cost more than 50 000 gold... make 2 recipes? :)

game/scripts/npc/items/item_rapier.txt
```diff
@@ -13,7 +13,7 @@
 
 		// Item Info
 		//-------------------------------------------------------------------------------------------------------------
-		"ItemCost"						"0"
+		"ItemCost"						"100000"
```
chrisinajar: Our shop doesn't support this yet, it needs to be 80k or less.

Trildar: 80k? Isn't the internal normal Dota gold currently capped to 50k?

Chronophylos: It's capped to 50k

PR #617, opened 2017-04-21
Make all currently implemented Reflex Items use modifier_generic_bonus

### LINKED ISSUE #569: Add stats to postactive and reactive reflex items
Currently only preemptive give stats, making them wildly more powerful than the others

Trildar:
This should be the Lua side fix so that everything actually gives stats. Closes #569.


Haganeko:
Thank you for fixing that.

PR #626, opened 2017-04-23
Custom MKB and Heavens Halberd icons

RionTwist:
Added custom icons that evolve for every level of MKB and Heavens
Halberd


addons/definition.json
```diff
@@ -0,0 +1,4 @@
+{
+  "game": "game/dota_addons/{ADDON_NAME}/",
+  "content": "content/dota_addons/{ADDON_NAME}/"
+}
```
Trildar: Not sure where this came from, but it shouldn't be in the folder that Git scans.

RionTwist: Whoops, that one was caused by a mistake I made when setting up node.js. I should be good to just delete it, right?

PR #627, opened 2017-04-23
Rewrite Lucience as Lua item

Trildar:
Notable changes include:
- Icon changing on toggle
- Cooldown when toggling both ways
- Various tooltip changes

Another not-so-obvious change is the interaction when a hero has multiple Lucience's of different levels. Higher levels effectively mute the aura effect of lower levels when in the same inventory. You'll still get the bonus damage though, and can also toggle the lower level Lucience, it just won't do anything.

The interaction when different heroes on the same team have different levels of Lucience is not really defined, so that will have all the usual Dota weirdness of auras of different levels.

P.S. I'm slowly growing more annoyed at how Dota auras work internally. It never seems conducive to whatever I'm trying to do.


game/scripts/vscripts/items/lucience.lua
```diff
@@ -0,0 +1,262 @@
...
+function modifier_item_lucience_regen_effect:GetTexture()
+  return regenIconName
+end
+
+------------------------------------------------------------------------
+
+modifier_item_lucience_movespeed_effect = class({})
+
+function modifier_item_lucience_movespeed_effect:OnCreated()
+  self.movespeedBonus = self:GetAbility():GetSpecialValueFor("speed_bonus")
```
chrisinajar: you need an OnRefresh also, if you're going to store values on self.

Trildar: I'm not sure it even calls `OnRefresh`, but I can put it in anyway. Shiva's Cuirass doesn't seem to have an `OnRefresh`, yet somehow its aura effect updates, which confuses me.

chrisinajar: It's possible auras get around it, it's just bitten me enough times that i think we should always do both if we choose to do one.

Trildar: Well, the weirder part is that Lucience effects don't seem to update properly. Even now that I've put in `OnRefresh`. Could be related to the item not directly giving the aura modifier, but I don't know.

Trildar:
Should be okay to merge now. Still has issues with multiple heroes carrying different levels not updating properly.

PR #633, opened 2017-04-23
Change shielder think method since it only runs once

chrisinajar:


Honeth:
i like the title as your changing shielder

PR #642, opened 2017-04-23
Twin Boss

KarmaticNeutral:
The basics of the twin boss.

Phases:
100 - 75% health
Set both twins to the higher HP every 2 seconds
75 - 50% health
Set both twins to the higher HP every 1.5 seconds
50 - 0% health
Set both twins to the higher HP every second


game/scripts/npc/units/twin/npc_dota_boss_twin_dumb.txt
```diff
@@ -0,0 +1,95 @@
+DOTAUnits"
+{
+//=================================================================================
+// Boss tier 2
+//=================================================================================
+	"npc_dota_boss_twin"
```
chrisinajar: should be `npc_dota_boss_twin_dumb`

game/scripts/npc/abilities/twin/boss_twin_spawn_twin.txt
```diff
@@ -0,0 +1,28 @@
+"DOTAAbilities"
+ {
+   //=================================================================================================================
+   // Generic boss resistance for % based damage
+   //=================================================================================================================
+   "boss_twin_spawn_twin"
```
chrisinajar: This ability can be removed.

game/scripts/vscripts/abilities/twin/boss_twin_twin_empathy.lua
```diff
@@ -0,0 +1,11 @@
+LinkLuaModifier("modifier_boss_twin_twin_empathy", "abilities/twin/modifier_boss_twin_twin_empathy.lua", LUA_MODIFIER_MOTION_NONE)
+
+boss_twin_twin_empathy = class({})
+
+function boss_twin_twin_empathy:GetIntrinsicModifierName()
+  return "modifier_boss_twin_twin_empathy"
```
chrisinajar: This should be removed since we want just 1 instance of the modifier, and for it to have proper Parent/Caster setup.

game/scripts/vscripts/units/twin.lua
```diff
@@ -0,0 +1,29 @@
+LinkLuaModifier( "modifier_boss_phase_controller", "modifiers/modifier_boss_phase_controller", LUA_MODIFIER_MOTION_NONE )
+
+function Spawn (entityKeyValues)
+  thisEntity:FindAbilityByName("boss_twin_twin_empathy")
```
chrisinajar: These two lines can be removed, they're repeated below properly.

game/scripts/vscripts/units/twin.lua
```diff
@@ -0,0 +1,29 @@
...
+  thisEntity:FindAbilityByName("boss_twin_spawn_twin")
+
+  thisEntity:SetContextThink( "TwinThink", partial(TwinThink, thisEntity) , 1)
+  print("Starting AI for " .. thisEntity:GetUnitName() .. " " .. thisEntity:GetEntityIndex())
+
+  ABILITY_spawn_twin = thisEntity:FindAbilityByName("boss_twin_spawn_twin")
```
chrisinajar: Since we don't need to use an ability to spawn the twin, we can just remove this.

game/scripts/vscripts/units/twin.lua
```diff
@@ -0,0 +1,29 @@
...
+  ABILITY_empathy = thisEntity:FindAbilityByName("boss_twin_twin_empathy")
+
+  local phaseController = thisEntity:AddNewModifier(thisEntity, ABILITY_empathy, "modifier_boss_phase_controller", {})
+  phaseController:SetPhases({ 75, 50 })
+  phaseController:SetAbilities({
+    "boss_twin_twin_empathy"
+  })
+end
+
+function TwinThink()
```
chrisinajar: Rename this function to something indicative of it's purpose, such as `SpawnDumbTwin`. Don't forget to change it above, as well.

game/scripts/vscripts/units/twin.lua
```diff
@@ -0,0 +1,29 @@
...
+  ExecuteOrderFromTable({
```
chrisinajar: Instead of using an ability to do this, since we only do it once and kind of want strong control over it, lets do something like...
```lua
local twin = CreateUnitByName("npc_dota_boss_twin_dumb", self:GetAbsOrigin(), true, self, self:GetOwner(), self:GetTeam())
twin:AddNewModifier(self, ABILITY_empathy, "modifier_boss_twin_twin_empathy", {})
```

This makes it so there's only 1 instance of the modifier, and it'll have the empathy ability's KV and levels at run time.

KarmaticNeutral:
And here folks we see the difference between a student and a master lol

game/scripts/npc/units/twin/npc_dota_boss_twin_dumb.txt
KarmaticNeutral: check

game/scripts/npc/abilities/twin/boss_twin_spawn_twin.txt
KarmaticNeutral: check

game/scripts/vscripts/abilities/twin/boss_twin_twin_empathy.lua
KarmaticNeutral: check

game/scripts/vscripts/units/twin.lua
KarmaticNeutral: check

game/scripts/vscripts/units/twin.lua
KarmaticNeutral: check

game/scripts/vscripts/units/twin.lua
KarmaticNeutral: check

game/scripts/vscripts/units/twin.lua
KarmaticNeutral: check

KarmaticNeutral:
[ W Entity System        ]: Cannot create an entity because entity class is NULL -1

game/scripts/npc/units/twin/npc_dota_boss_twin.txt
```diff
@@ -0,0 +1,95 @@
+DOTAUnits"
```
Honeth: this 1 missing "

game/scripts/npc/units/twin/npc_dota_boss_twin_dumb.txt
```diff
@@ -0,0 +1,95 @@
+DOTAUnits"
```
Honeth: missing a " aswell

game/scripts/vscripts/abilities/twin/modifier_boss_twin_twin_empathy.lua
```diff
@@ -0,0 +1,36 @@
+--LinkLuaModifier("modifier_boss_twin_twin_empathy", "abilities/twin/boss_twin_twin_empathy.lua", LUA_MODIFIER_MOTION_NONE)
+
+modifier_boss_twin_twin_empathy_buff = class({})
+
+--This may need to be in the abil not the mod
+function modifier_boss_twin_twin_empathy_buff:OnCreate()
+  local interval = self:GetSpecialValueFor( "heal_timer" )
+  mod:StartIntervalThink(interval)
```
chrisinajar: change mod to `self`

game/scripts/vscripts/abilities/twin/boss_twin_twin_empathy.lua
```diff
@@ -0,0 +1,7 @@
+LinkLuaModifier("modifier_boss_twin_twin_empathy_buff", "abilities/twin/modifier_boss_twin_twin_empathy.lua", LUA_MODIFIER_MOTION_NONE)
```
chrisinajar: You can remove this file and just make the ability data driven, it doesn't do anything and is only there for KV data and tooltips.

KarmaticNeutral: Isn't it called in twin.lua?

ABILITY_empathy = thisEntity:FindAbilityByName("boss_twin_twin_empathy")
twin:AddNewModifier(thisEntity, ABILITY_empathy, "modifier_boss_twin_twin_empathy_buff", {})

Trildar: The same could be done if it was DataDriven, but I think it would make more sense to make the ability give the modifier rather than adding the modifier in `twin.lua`. Then make the ability passive if it isn't actually cast for anything.

game/scripts/vscripts/abilities/twin/modifier_boss_twin_twin_empathy.lua
```diff
@@ -0,0 +1,36 @@
+--LinkLuaModifier("modifier_boss_twin_twin_empathy", "abilities/twin/boss_twin_twin_empathy.lua", LUA_MODIFIER_MOTION_NONE)
+
+modifier_boss_twin_twin_empathy_buff = class({})
+
+--This may need to be in the abil not the mod
+function modifier_boss_twin_twin_empathy_buff:OnCreate()
```
Trildar: Shouldn't it be `OnCreated`?

game/scripts/vscripts/abilities/twin/boss_twin_twin_empathy.lua
KarmaticNeutral: It is only ever applied to another target. It is never used as a passive.

game/scripts/vscripts/abilities/twin/modifier_boss_twin_twin_empathy.lua
```diff
@@ -0,0 +1,36 @@
...
+  return false
+end
+
+function modifier_boss_twin_twin_empathy_buff:IsPurgable()
+  return false
+end
+
+function modifier_boss_twin_twin_empathy_buff:OnIntervalThink()
+  if IsServer() then	
+  end
```
Trildar: Left this by accident? Or is this suppose to enclose the whole function body?

game/scripts/vscripts/abilities/twin/modifier_boss_twin_twin_empathy.lua
Trildar: `StartIntervalThink` is only called once here, which means the interval won't update when the phase changes. Not sure if calling `StartIntervalThink` at the end of `OnIntervalThink` would work.

game/scripts/vscripts/abilities/twin/boss_twin_twin_empathy.lua
Trildar: The ability itself doesn't technically do anything though, it's not cast to apply the modifier, that's just done directly when spawning. Since the caster and parent need to refer to each of the two twins though I guess making it the intrinsic modifier of the ability wouldn't work. You might want to consider rewriting it to use casts of this ability to actually do the healing. Mainly because the modifier has an issue with updating the interval.

game/scripts/vscripts/abilities/twin/modifier_boss_twin_twin_empathy.lua
chrisinajar: Modifier intervals keep running until you run StartIntervalThnk(-1)... Either way `mod` isn't defined anywhere so you're not going to get any intervals until you change that to `self`.

PR #645, opened 2017-04-24
Mostly tooltip changes

imaGecko:
Cleaning up after myself:
Fixed tooltip oversights in preemptive 2b and 3a

Bosses:
Added shielder to npc_bosses
added tooltip for shielder shield ability
Moved and changed the phase controller modifier from charger to boss

Reflex Cores:
Changed the order of AbilitySpecial to the have passive stats first

Other:
Bottle counter now counts bottles instead of charges and has a tooltip

### DIFF
--- a/game/scripts/vscripts/components/filters/bottlecounter.lua
+++ b/game/scripts/vscripts/components/filters/bottlecounter.lua
@@ -31,7 +31,7 @@ function BottleCounter:Filter(filterTable)
 
     if item:GetName() == "item_infinite_bottle" and not item.firstPickedUp then
       item.firstPickedUp = true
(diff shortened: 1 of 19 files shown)


game/scripts/vscripts/components/filters/bottlecounter.lua
```diff
@@ -31,7 +31,7 @@ function BottleCounter:Filter(filterTable)
 
     if item:GetName() == "item_infinite_bottle" and not item.firstPickedUp then
       item.firstPickedUp = true
-      player.bottleCount = player.bottleCount + 3
+      player.bottleCount = player.bottleCount + 1
```
chrisinajar: I guess there's no reason to count charges...

PR #654, opened 2017-04-25
Now purging modifiers when resetting players

Chronophylos:


chrisinajar:
The only problem is you don't get your modifiers back..

Chronophylos:
I'll add that later.

PR #655, opened 2017-04-25
Made Units not walk away in duels

Chronophylos:
Just calling Stop() when a duel starts. This prevents players to just walk away or cast spells ordered before the duel.


chrisinajar:
That chunk of code is duplicated, you need to put the stop commands in each.

PR #675, opened 2017-04-29
Added Greater Tranquil Boots

SquawkyArctangent:
Complete with not breaking from neutral attacks, regenerating creep
damage over time.

Also adds a "life sap" passive that slowly kills neutrals that attack
you. Mostly just for feel - can easily be completely disabled without
diving into the lua by just setting either of it AbilitySpecials (
"creep_sap_damage" and "creep_sap_duration" ) to 0.


Haganeko:
Apparently the sap itself is missing a tooltip. Also, it was not in the original design, and was added at the implementer's sole discretion.

chrisinajar:
I like the idea, I'll review the code later. Fix the tooltip or make it hidden, you can use `node scripts\find-tooltips.js` to scan for more missing tooltips before submitting

game/scripts/npc/items/farming/item_greater_tranquil_boots_2.txt
```diff
@@ -116,6 +117,18 @@
...
+			// of creeps "bleeding" from attacking the wearer
+			"12"
+			{
+				"var_type"				"FIELD_FLOAT"
+				"creep_sap_damage"		"1.6 1.9 2.1 2.5 5"
+			}
+			"13"
+			{
+				"var_type"				"FIELD_FLOAT"
+				"creep_sap_duration"	"2.0"
```
chrisinajar: This is a float, but since the timer is on 1 second, any decimal has no effect. Should be integer for clarity, or could should respect decimal values.

game/scripts/vscripts/items/item_greater_tranquil_boots.lua
```diff
@@ -0,0 +1,261 @@
...
+if IsServer() then
+	function modifier_item_greater_tranquil_boots:OnIntervalThink()
+		if self.storedDamage and self.storedDamage > 0 then
+			local parent = self:GetParent()
+			local spell = self:GetAbility()
+			local maxHeal = math.min( spell:GetSpecialValueFor( "regen_from_creeps" ) * 0.1, self.storedDamage )
+			
+			parent:Heal( maxHeal, parent )
+			
+			self.storedDamage = self.storedDamage - maxHeal
```
chrisinajar: So you heal all creep damage you take every 0.1 seconds? Seems legit.

Haganeko: I believe it heals you for a fixed amount every .1s, up to the damage you've taken from creeps. As in, taking damage from creeps will never result in net healing.

SquawkyArctangent: You missed the line right above it - it heals for either 10% of regen_from_creeps or the remainder of stored damage, whichever is lower.

chrisinajar: Oh i see, so you heal for the amount of creep damage you've taken at a rate of `regen_from_creeps`. That makes sense.

PR #676, opened 2017-04-29
Particle effects for rune hill

### LINKED ISSUE #646: Add particle effects to the rune hill
Currently being in the rune hill (high ground area in duels which gives you buffs if you stay in there) applies a modifier but no visual effects to the hero. Lets get some sort of "powering up" effects to appear around heroes actively holding the high ground area.

Bonus points for making the effects scale based on current charges on the modifier.

kallleeeeh:
Added particles to rune hill. One effect triggers on entry and the others whenever a rune is given. I think we might need custom effects to indicate what effect is given; the existing ones are not that great.
Also made the counter stop at 130 so it's clear that there are no more rewards.
I did not add a particles indicating the area.
Partially closes #646

### DIFF
--- a/game/scripts/vscripts/modifiers/modifier_duel_rune_hill.lua
+++ b/game/scripts/vscripts/modifiers/modifier_duel_rune_hill.lua
@@ -59,6 +59,10 @@ function modifier_duel_rune_hill:OnIntervalThink()
     return
   end
 
+  if self:GetStackCount() == 130 then
+    return
+  end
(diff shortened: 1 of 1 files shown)


Haganeko:
The counter refreshes the buffs though...

kallleeeeh:
I could trigger the first effect at 4 stacks. For some reason the stacks sometimes flutter between 0 and 2.

Issue #677, opened 2017-04-29
Create A Literal Roshan for T1


KarmaticNeutral:
Make Roshan but with boss resistance.

chrisinajar:
Does roshan really need boss resistance? What about literally roshan? Before you kill T1 your heroes are just literally dota.

KarmaticNeutral:
Literally Fixed

PR #680, opened 2017-04-30
Make T1 be literally rosh with boss resistance

KarmaticNeutral:
As discussed during the 4/29 TH meeting


KarmaticNeutral:
closes #677

Trildar:
Just a note that this would cause the boss to scale in health and damage over time. Not at a very fast rate, and not really that relevant since the boss should be taken early, but we should think about whether we really want the scaling or not.

Trildar:
Isn't this Rosh still vulnerable to %damage though? Also, you didn't change the health, so without boss resist he has 10% the ehp he should.

PR #688, opened 2017-05-01
Added Schema

Chronophylos:
Added new getter functions
Changed Bottle Counter to use a table
Reworked Points Manager


chrisinajar:
I assume I should run write version each time I tag a new version?

Chronophylos:
Yes

PR #689, opened 2017-05-01
Moving doors

Chronophylos:
"Animations" are a bit laggy.


warpdragon:
Hey! That's pretty good!

chrisinajar:
We can remove the animations if they suck. Nice work!

PR #691, opened 2017-05-01
Icon update 2017-05-01

warpdragon:
Adding the new artwork from Ben.


warpdragon:
Tested and working.
image

PR #695, opened 2017-05-02
luna

l34Um1:


game/scripts/npc/abilities/luna_eclipse.txt
```diff
@@ -25,22 +25,22 @@
...
+
 		// Special
 		//-------------------------------------------------------------------------------------------------------------
 		"AbilitySpecial"
 		{
 			"01"
 			{
 				"var_type"					"FIELD_INTEGER"
-				"beams"						"5 8 11 17 40"
+				"beams"						"5 8 11 13 15"
```
Haganeko: Shouldn't these be 14 and 17? To keep with the progression.

l34Um1: no, since lucent beam also scales, scaling both leads to silly things. this progression is mildly above the average progression of most spells

Minnakht: I'm also in favour of 14 and 17 so that it keeps being +3 per rank. I don't think the 2 extra beams will break everything again.

chrisinajar:
She's also a carry, and CM isn't. CM's ult *should* scale harder than Luna's because luna also has other shit going for her. If she's too weak, we'll buff.

TheObbyDragon:
But luna's ultimate requires the enemies to be near it so keeping it with high bursts of damage is better but you could just buff her lucent beam stun

PR #699, opened 2017-05-03
Variable Exponential XP onset 

Haganeko:
Changed onset of exponential experience based on chosen game length.

### DIFF
--- a/game/scripts/vscripts/components/creeps/creep_power.lua
+++ b/game/scripts/vscripts/components/creeps/creep_power.lua
@@ -18,7 +18,7 @@ local ExponentialGrowthOnset
   end
 
   if minute > ExponentialGrowthOnset[PointsManager:GetLimit()] then
-    multFactor = 1.5 ^ (minute - 60)
+    multFactor = 1.5 ^ (minute - ExponentialGrowthOnset[PointsManager:GetLimit()])
   end
 
   return {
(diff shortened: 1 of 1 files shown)


game/scripts/vscripts/components/creeps/creep_power.lua
```diff
@@ -10,7 +10,7 @@ function CreepPower:GetPowerForMinute (minute)
     return {   0,        1.0,      1.0,      1.0,      1.0,      1.0,      1.0 * self.numPlayersXPFactor}
   end
 
-  if minute > 60 then
+  if (minute > 40 and PointsManage:GetLimit() = 50) or (minute > 50 and PointsManage:GetLimit() = 100) or (minute > 120 and PointsManage:GetLimit() = 200) then
```
spar36: Looks to be right. I believe it should be double equals though `==`, not `=`.

chrisinajar: `=` assigns, `==` compares.

Haganeko: Oh yeah. I had forgotten. Will fix.

Trildar: It also should be `PointsManager` I believe. It might also be better to use a lookup table that maps kill limit to time of onset.
i.e. something like
```local exponentialGrowthTime = {
  50 = 40,
  100 = 50,
  200 = 60
}
if minute > exponentialGrowthTime(PointsManager:GetLimit()) then
...```

Trildar:
Also, you didn't change the formula. `1.5^-20` multiplication factor at minute 40.
You can use the table again to make the formula dynamic.

PR #701, opened 2017-05-04
Fix more illusion stats

### LINKED ISSUE #276: Fix illusions to use diminishing stat return
Currently any time you summon an illusion passed level 25, the illusion is stronger than you. This is because dota calculates the stats for that level when they spawn them, so our level up code doesn't edit the stats. We need to hook on them being summoned and adjust their stats to match the hero they were created from.

### LINKED ISSUE #663: Morphling's Hybrid Illusion stats
Hybrid uses Morphling's stats not the targets

Trildar:
Fixes stats for Hybrid and Tempest Double. Closes #663. Hopefully closes #276.

### DIFF
--- a/game/scripts/vscripts/components/progression/hero_progression.lua
+++ b/game/scripts/vscripts/components/progression/hero_progression.lua
@@ -80,9 +80,8 @@ function HeroProgression:ReduceIllusionStats(illusionEnt)
   local SetBaseStat = partial(self.SetBaseStat, illusionEnt)
   local GetStatGain = partial(self.GetStatGain, illusionEnt)
 
-  -- Note: Will not run on very first spawn of Tempest Double in a game due to the flag only being set on the next frame
-  if illusionEnt.IsIllusion and (illusionEnt:IsIllusion() or illusionEnt:IsTempestDouble()) and illusionEnt:IsHero() then
-    -- Set one frame delay because illusions won't immediately have the correct level
+  -- Set one frame delay because illusions won't immediately have the correct level
+  function ReduceStats()
     Timers:CreateTimer(function()
       local currentHealth = illusionEnt:GetHealth()
       local currentMana = illusionEnt:GetMana()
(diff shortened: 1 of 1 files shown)


game/scripts/vscripts/components/progression/hero_progression.lua
```diff
@@ -5,96 +5,114 @@ if HeroProgression == nil then
     ChatCommand:LinkCommand("-levelup", "OnLevelUpChatCmd", HeroProgression)
 end
 
+GameEvents:OnPlayerLevelUp(function(keys)
```
chrisinajar: Why remove this from init?

Trildar: For some reason the event listeners seem to reset when `script_reload` is called. So having these in init made this component pretty much stop working whenever `script_reload` was done.

chrisinajar: What's `script_reload`?

Trildar: A console command. Reloads Lua scripts so that at least some things can be changed and tested without having to restart the whole map.

Trildar:
Probably hold on this anyway. I just realised Rubick could potentially derp the system. Will work on that tomorrow.

Haganeko:
Couldn't a morphling with manta also derp the system? As in, the illusions would have pre morph stats

Trildar:
Nah, I covered that case already. Morphling illusions copy strength and agility from the original by just vanilla code, so I just made it only adjust intelligence for any Morphling illusion. Rubick is a potential issue due to incomplete coverage for Tempest Double. I thought it would be fine because there's only an issue if Tempest Double is first cast when over level 25.

Haganeko:
What about a morphed Rubick?

Trildar:
Morphed Rubick doesn't copy stats for illusions in vanilla. The current system should work fine and apply unmorphed stats.

Trildar:
Should be fine to merge now. Assuming the event registry thing is fine. I can move it back if you really want since it should be fine in actual games.

PR #713, opened 2017-05-05
Hopefully fix travis

Chronophylos:


Chronophylos:
This should fix the errors during deploy

Haganeko:
It's this also obsolete?

chrisinajar:
This edits a broken script to still be broken.

Chrono, when you're trying to find stuff relative to yourself use
```js
var path = require('path');

// join takes any number of paths and puts the correct slashes in and normalizes the path
// __dirname is the full path to wherever *this* script is
var filePath = path.join(__dirname, '../../some/path/to/file');
```

Chronophylos:
Ohh.
The `(os.)path.join` I know just puts `/` in between

PR #716, opened 2017-05-06
Fixed cave teleporting players outside the cave and other things

Chronophylos:
GDS Fixes
Increased Door speed


chrisinajar:
`tooltip_stopfightingyourself.txt` is somehow broken

PR #727, opened 2017-05-06
Fix roshlings (I hope)

Haganeko:
Got rid of inherent roshan buffs from cave roshlings. Hope it'll fix them.

### DIFF
--- a/game/scripts/npc/units/npc_dota_mini_roshan.txt
+++ b/game/scripts/npc/units/npc_dota_mini_roshan.txt
@@ -16,7 +16,7 @@
 		"Ability1"					"roshan_spell_block"	// Ability 1.
 		"Ability2"					"roshan_bash"		// Ability 2.
 		"Ability3"					"roshan_slam"		// Ability 3.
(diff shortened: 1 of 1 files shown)


Trildar:
You'll also want to remove Devotion if you don't want them to have vanilla Roshan scaling armour bonus. Possibly some other scaling tied to that as well. Not too sure.
Also, you probably want to change the AttackRate, unless you actually want them to have 1 attack every 2 seconds, because the inherent buffs gives them attack speed such that they will have 1 attack per second.

Haganeko:
Oh. Can we give them a passive which grants attack speed, so they won't be super vulnerable to slows?

chrisinajar:
Giving them a passive to increase attack speed seems silly compared to just making them attack faster. We could also do some responsive modifier that gives attack speed based on minus attack speed (like a % resistance to attack speed slow) but that's another-PR-work for future chris

Haganeko:
Increasing base attack speed will make them vulnerable to slows. Creating a whole attack slow resistance is way more complicated than just giving them AS

PR #728, opened 2017-05-07
Necromastery

Haganeko:
Someone raised the concern that this ability doesn't progress past level 4, which is weird and bad design all around.
EIther way, even with the talent and scepter he doesn't really get to as high a damage bonus as Underlord consistently keeps late.


chrisinajar:
more than 2x buff to damage at max level, I would love for Baumi's feedback on the balance here. Personally I think it's fine, and will merge for now, just make sure he's aware.

PR #738, opened 2017-05-07
Fix Cave TP

Chronophylos:
IsTouching doesn't work. Reverting and fixing initial Bug.


chrisinajar:
You could manually check bounds, also you only need to use room 1 zone since they all span to the very back (to prevent getting stuck between rooms)

Chronophylos:
I figured that.

PR #744, opened 2017-05-07
The Great Creep Update

Haganeko:
Or something. Reduced easy creeps from 4 to 3, mediums from 3 to 2.5, and hard creeps from 2.5 to 2. Implemented black dragon. Waiting for a commit from Chrono with black dragon KVs.


Haganeko:
Will work more on this

PR #747, opened 2017-05-07
The Great Creep Update

Haganeko:
Added three new kinds of creeps. Should make aoe farming harder.


Haganeko:
Seems to work, though Travis did some weird things at the beginning of the test.

PR #761, opened 2017-05-08
Doom Doom typo fix

Minnakht:
Rescaled Scepter upgrade damage from 20/35/50/800/200 to 20/35/50/80/200, a very necessary nerf


Trildar:
Not that necessary. Last I checked the values aren't actually used since Doom Agh's doesn't upgrade the damage anymore. Can't hurt to fix it though.

PR #788, opened 2017-05-13
Add illusion, reincarnating, and distant kill checking for Bloodstone charge handling

### LINKED ISSUE #777: PL illusions and upper level bloodstone charges
everytime you kill an ilusion of pl u get a bloodstone charge

Trildar:
Closes #777. There is, however, still the issue of the current implementation not granting charges for distant kills by Bloostone owners. I was in the middle of rewriting it to do the charge gain in `modifier_item_bloodstone_oaa:OnDeath` when I realised that has the issue of incrementing charges on all Bloodstones in the inventory rather than just the most top-left one.


Trildar:
Managed to figure out a way to put in the distant kill charge credit as well.

Haganeko:
Need to make the charge collector hidden, and make IsBloodstone local.

Trildar:
The charge collector isn't even used now. Forgot to take out the LinkLuaModifier line for it. Not really sure what's going on with the IsBloodstone thing. It's a function declaration, which should be locally scoped I think. And the following error about accessing an undefined variable makes even less sense.

Haganeko:
The first one is clearly about global scope of the function it's calling a variable. The second error is a mystery and I was hoping fixing the first would magically fix it.

Trildar:
Well, TIL that we were leaking global functions everywhere until that Luacheck PR a while back. And that Lua really likes to make everything global by default.

Haganeko:
It did!

PR #790, opened 2017-05-14
Added Siege Mode item

SquawkyArctangent:
This is going to need a pass-over by the balance team, as I kept the
values from the keyvalues that were already there - which, due to the
fact that there was no number for splash damage, means that it deals
100% splash damage. Have fun exploiting this before they get to it!

Will also need another pass-over code-wise later on, both to implement
active icons ( trying to do so ahead of time would just result in game
crashes ) and to fix the fact that you can shove it into your backpack
and keep it active, like with Stoneskin Armor. The latter of which I
struggled to fix, but OnInventoryContentsChanged just seems to be dead
with this repo.


Haganeko:
Travis disapproves your use of whitespace.

SquawkyArctangent:
I don't actually have any idea what that means.

KarmaticNeutral:
It disapproves of your style. All the code is okay as far as it can tell, but it doesn't like the way it looks. I'm assuming it's your tabbing. We are supposed to use 2 spaces as tabs not 6.

Haganeko:
I think it's about lines with only white space and no other content, in this case.

Haganeko:
It works! \o/

warpdragon:
Wow siege mode in the house!

chrisinajar:
I think the best way to catch items going away from inventory are destroy events of intrinsic modifiers

SquawkyArctangent:
Belatedly, that unfortunately doesn't work if you're reusing a vanilla item's modifier.

Haganeko:
So this is bugged?

SquawkyArctangent:
That was literally established in the commit's comment.

PR #795, opened 2017-05-15
Patch 7.06 Talents and Abilities

imaGecko:
Adds 7.06 changes to Hero abilities and talents
Deleted custom talent files for heroes that no longer have talents to be replaced

PR #796, opened 2017-05-16
Update abilities for new Talents

Trildar:
Mass Serpent Wards and Call of the Wild needed to be updated for 7.06 Talents since they're rewritten in Lua. Mass Serpent Wards was also probably not applying the old +1 Ward Health Talent. That was also fixed.


Trildar:
Should mention that this PR doesn't include giving the heroes the Talents themselves since that's already in #795.

PR #799, opened 2017-05-16
crimson and silver edge

hallelsegel:
I think it is good now (is it?)


warpdragon:
Looks good. Also, are u sure the lvl 1 icons have the correct pips since u are not reuploading them? Just be double sure.

hallelsegel:
Indeed they do

PR #812, opened 2017-05-20
Fixed Siege Mode becoming unusuable if you die while its active

SquawkyArctangent:
very complicated stuff here


chrisinajar:
Should you set self.mod to nil if it's isNull?

SquawkyArctangent:
Yes, because IsNull doesn't check if it's nil, but rather if the object doesn't exist any more.

Basically, it's a check to see if a pointer points to a dead object, rather a check to see if ... well, the variable isn't anything.

chrisinajar:
Ah, i didn't expand the code to see the `else` block, i was saying it should be nil'd out so that the other code would recreate it. Looks good :+1:

Haganeko:
Conflicts D=

Chronophylos:
I resolved your nonexisting conflicts

PR #817, opened 2017-05-21
Make Bottle drops non-random and rework creep spawner internals

Trildar:
Added a system for per creep spawn drops and made Bottles drop 1 for every 4 creeps spawned in a camp. 

Also changed the creep spawner to track creeps in a camp rather than doing a unit search in the camp. This could potentially break under some circumstances. I've done some testing, but I don't think it was particularly comprehensive. I also don't know if this is really desirable in the first place. The motivation for this was that, with the new map, a few of the camps are so close to each other that doing a unit search could potentially make the spawner think that creeps in another camp belong to a different camp (this messes with the max creep spawn and power/bounty distribution logic, though total bounty is unaffected). With the current range of 1000, it's actually doing that with the creeps just in their idle position. Simply reducing the range would work, but I think we would need to either accept that camps can still overlap sometimes if they're aggro'd into the correct spots, or that camps can be "stacked", though the only advantage of that would be causing more creeps rather than stronger ones.

### DIFF
--- a/game/scripts/vscripts/components/creeps/item_drop.lua
+++ b/game/scripts/vscripts/components/creeps/item_drop.lua
@@ -82,6 +82,12 @@ function CreepItemDrop:CreateDrop (itemName, pos)
 end
 
 function CreepItemDrop:AddFixedDropsToCamp(creeps, numberOfCreepsSpawned)
+  local function IsNotNumber(value)
+    return not (type(value) == "number")
+  end
+
+  local sequencedCreeps = totable(filter(IsNotNumber, pairs(creeps)))
+
   local function AddItemToCamp(item)
     local from = item[FROM_ENUM]
     local to = item[TO_ENUM]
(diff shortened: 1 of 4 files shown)


Trildar:
Is this awaiting code review, design consensus, or was the wall of text a bit much? :P

Trildar:
Ah right, forgot about domination. Not sure how to handle that for the item drops. Could redistribute them I guess. The alive creep table should be workable by listening for OnDominated.

As for the fake stacking. The old range of 1000 made it near impossible, if not actually impossible. And the new range of 800 makes it only marginally viable still. Plus we'll be telling people that there isn't really any point to traditional Dota stacking. Trying to explain fake stacking might get a bit confusing. Of course, not like Dota is the easiest to understand either sometimes.

Trildar:
Anybody else have opinions​ on this? (Assume domination abilities aren't an issue).

Maybe I should have brought this up in Townhall. :P

Trildar:
Rebased and added handling for creeps getting dominated. A bit difficult to test the domination thing thoroughly, but it seems to be working as intended.

Trildar:
Planned pseudo-random Bottle drops instead.

PR #818, opened 2017-05-21
Add insta-kill dev Dagon

Trildar:
Should target just about everything, including allies, magic immune, and invulnerable units. Though I didn't really test for all of those.


Chronophylos:
A Linux user!

Trildar:
Eh, not really. I've done some stuff in Linux, but I've never had a Linux PC as my main environment.

Chronophylos:
Me neither, I have Arch Linux and Windows in Dual Boot but playing games sadly requires Windows.
I saw `kill -9` and thought of a Unix user.

PR #832, opened 2017-05-22
Temporary Core redistribution fix

KarmaticNeutral:
This will use an awkward distribution of bosses. It will do to improve gameplay until we completely revamp boss spawning. (We were planning to do that anyway.)


KarmaticNeutral:
i just changed the starting tier on boss pits so that there is a T3 farthest from base, T2 just closer, and the rest are T1. This drastically reduces our control of which bosses spawn in each pits but seeing as we don't currently have many bosses and we were planning a complete remodel of boss pits anyway to improve control, this will work until that happens.

PR #835, opened 2017-05-23
Potential up preduel time w/o upping farm time

KarmaticNeutral:
Attempting to cause the first creep spawn to occur at 30 seconds
Keeping the second at 60
Then continuing at every minute from then on.

Also, move the starting time 5 seconds sooner so you have more time before duel to buy items


Haganeko:
It's the farming cave initial spawn also delayed?

KarmaticNeutral:
I haven't done that yet. I probably should. I'll worry about that once I get the rest of the camps working.

KarmaticNeutral:
Ready for Merging. Tested and functioning.

PR #837, opened 2017-05-23
Rewrite Rearm in Lua and make it not refresh Reflex Items

### LINKED ISSUE #606: Disable Rearm on Reflex Items
Currently Tinker can rearm Reflex Items. He shouldn't be able to.

Trildar:
Closes #606


Haganeko:
Ooh, it's rewritten? That means we get to play with it!

chrisinajar:
Change the name of the ability. Overwriting built in abilities with the same name can cause random crashes (seen with both bottle and tiny grow)

Trildar:
Mass Serpent Wards, Demonic Conversion, Call of the Wild, and Nature's Call all overwrite though. Guess those should be changed too? Should the IDs also be changed?

Chronophylos:
I resolved the conflicts created by the now clean kv files

chrisinajar:
I think it only causes random crashes when dota has special code that interacts with that item or ability, such as bottle's special sharing properties or *everything about tiny grow*. Those other abilities are probably fine, but it's a trivial change to make sure it doesn't happen with rearm, which i wouldn't be surprised to see special interactions with.

Trildar:
Should be good to go.

chrisinajar:
Should you also respect that IsRefreshable method or whatever it's called, or is that not desired?

Trildar:
I'm not sure. Vanilla Rearm didn't seem to when I tested it before. Refresher Core probably should though.

PR #840, opened 2017-05-23
Added Version Display in upper right corner

Chronophylos:
image


chrisinajar:
*love* it, fix conflicts!

PR #842, opened 2017-05-23
Revert "Armor Progression"

chrisinajar:
Reverts OpenAngelArena/oaa#822
Turns out this was rather contentious, but none of the discord conversation made it into github.


chrisinajar:
I do, however, really agree with the stoneskin buildup change. #makestoneskinstrongagain

Haganeko:
The problem was mostly about -armor items other than Crest using the same progression as Crest, I believe, and whether Shiva's Cuirass, which this doesn't touch due to conflicts with Warp's PR, should get a small shift of some of its armor to the aura.

Issue #452, opened 2017-04-06
Tier 1 bosses can be stolen


Haganeko:
The opposing team can walk into one's tier 1 bosses and kill them, preventing the team from obtaining farming and reflex cores.

Haganeko:
Just to be clear, this is happening and it is not supposed to.

chrisinajar:
It was decided long ago that each team is guarenteed both their tier 1 bosses, and that the enemy team can't even get up there to try and interrupt them. This is a good candidate for zone control, and a basic damage filter custom ability.

 * Create custom abilities for tier 1 bosses that makes them not take damage from one team or the other.
   * Make 2 abilities so that the tooltips / icons can be super pleasant.
 * Give the tier 1 bosses the correct passives inside of `boss/spawn.lua`
 * Create zones around the tier 1 boss areas and lock the enemy teams out until the bosses have been killed.

darklordabc:
Do you mean do one or the other? You wouldn't need the damage filter AND the zone blocking ya?

Rachol:
What about pre-asigning dropped items to the right team? If other team wants to help killing the boss, then why not? :)

Haganeko:
The zone blocking is mostly so people can't come in to steal the boss/kill your team while you're trying to do the boss.
The damage block is mostly so people can't use long range abilities (or attacks) to steal the boss/kill your team while you're trying to do the boss.

RamonNZ:
I agree with Rachol here. Simplest solution.

Haganeko:
This is not a design discussion. This is an incomplete implementation. But yes, items should be preassigned.

warpdragon:
Yeah but remember Baumis point about making sure players understand why theyre not getting the items when they kill the boss. And remember the design rules: the players should understand this based on design, without being told. That being said i recommend, at minimum, OBVIOUS terrain decorations, such as making the cliff itself look very similar to an actual dire or radiant structure. This can be accomplished with simple terrain modifications which, incidentally, I'm *totally* down to do. It will also give the same non-verbal message "this area is only for team x".

Haganeko:
Can we please not relitigate this? This is the current plan:
 - the areas where the tier 1 bosses are will be made inaccessible to the enemy till they are dead;
 - upgrade and farming cores will be preassigned;
 - attacking the enemy's tier 1s should be ineffective. 
This has not been implemented.

warpdragon:
[...]

PR #849, opened 2017-05-24
Lock out enemy heroes from tier 1 bosses

chrisinajar:


chrisinajar:
Partially closes #452, should really give rewards / unlock regardless of who kills

Minnakht:
Rewards regardless of killer is a solution I'd be more in favor of as well if it was up to discussion at this point

chrisinajar:
We need both.

chrisinajar:
We have both!

PR #855, opened 2017-05-24
Handle leveling up talents properly in the UI

chrisinajar:


game/scripts/vscripts/components/abilities/level.lua
```diff
@@ -7,20 +7,29 @@ end
...
   local canLevelUp = {}
+  local hasNoPoints = hero:GetAbilityPoints() == 0
 
   for index = 0, hero:GetAbilityCount() - 1 do
     local ability = hero:GetAbilityByIndex(index)
     if ability then
       local abilityName = ability:GetAbilityName()
-      table.insert(canLevelUp, AbilityLevels:GetRequiredLevel(hero, abilityName))
+      if hasNoPoints then
+        table.insert(canLevelUp, 9999)
```
Haganeko: This is just a dumb question, but... Why 9999?

chrisinajar: because you can't possibly get to that level, just in case you wait until the last possible level to skill your lvl 25 talent. on the ui it compares you level to the value there as a "required level"

chrisinajar: that said, i don't think it works, but it visually works, so i think the ui code is just handling the bad value

chrisinajar: it does work, but -1 is better

Haganeko: Wouldn't any level return true for level > - 1 though? Eh. I guess I don't get Lua.

chrisinajar: The UI consuming code looks like this:
`      if (!abilityPanel.BHasClass('could_level_up') || requiredLevel === -1 || data.level < requiredLevel) {`

chrisinajar: Technically the first part of the or block is getting caught, so we don't even need to send -1

Haganeko: I don't know the triple equal sign or the double pipe. I assume they are neq and or, respectively?

PR #863, opened 2017-05-26
Minimap icon png assets

warpdragon:


warpdragon:
ok so is there some kind of git ignore on /game/materials? i keep resetting to upstream/master and my /game/materials folder still has weird shit in it.

warpdragon:
anyway it shoudl be fine. the vmat_c stuff should compile at build time i think.

Trildar:
All compiled files are ignored. So yeah, probably all that stuff is being ignored by git.

Chronophylos:
what will have happened?

PR #876, opened 2017-05-26
Fix modifier and channel icons for Greater Travel Boots

Trildar:
"item_" prefix is apparently necessary for the channel icon to know to search in the items folder. Valve pls, can you ever make sense.


chrisinajar:
I can't believe this works.

PR #891, opened 2017-05-27
Various item fixes and updates

SquawkyArctangent:
- Fixed Greater Power Treads, Greater Tranquil Boots, and active Greater
Phase Boots and Siege Mode spitting out errors when upgrading them.
- Fixed Greater Power Treads splash damage not splashing off of or
affecting spell immune creeps.
- Fixed Greater Phase Boots phase attack not proccing on spell immune
creeps.
- Greater Phase Boots Phase attack now deals reduced damage, using the
"phase_attack_outgoing" AbilitySpecial.
- Siege Mode now uses its alternative icon while active. Additionally, a
"siege_cast_range" AbilitySpecial was hooked up, but is currently set to
0 - basically, it's for experimenting when there's time to experiment.
- Updated the tooltips of Greater Power Treads, Greater Phase Boots, and
Siege Mode.


game/scripts/npc/items/farming/item_greater_phase_boots.txt
```diff
@@ -114,6 +114,11 @@
         "var_type"                                        "FIELD_INTEGER"
         "phase_radius"                                    "200"
       }
+      "12"
+      {
+        "var_type"                                        "FIELD_INTEGER"
+        "phase_attack_outgoing"                           "30 45 60 75 90"
```
chrisinajar: This seems really weak. perhaps scale from 50 up to 100?

game/scripts/vscripts/items/farming/greater_tranquil_boots.lua
```diff
@@ -61,6 +61,11 @@ function modifier_item_greater_tranquil_boots:OnCreated( event )
 
 		self:StartIntervalThink( 0.1 )
 	end
+
+	self.moveSpd = spell:GetSpecialValueFor( "bonus_movement_speed" )
```
chrisinajar: If you're going to change this to use self.attributes then you need to set them up inside of OnRefresh too.

game/scripts/vscripts/items/siege_mode.lua
```diff
@@ -58,6 +73,31 @@ end
 
 --------------------------------------------------------------------------------
 
+function modifier_item_siege_mode_siege:OnCreated( event )
+	local spell = self:GetAbility()
+
+	spell.mod = self
+
+	self.atkRange = spell:GetSpecialValueFor( "siege_attack_range" )
```
chrisinajar: Same deal, needs refresh handling. you can also remove the or's if you make this change

SquawkyArctangent: Actually, in this case, there is no reason to do this, as the only way to apply the modifier ( aka activating Siege Mode ) _will_ remove the modifier should it already exist. So it can't actually be refreshed.

chrisinajar: If that's the case then there's no need to have all those `or`s... 

My vote is to add the refresh methods even if they're never called, it allows the modifier to be used in different ways if we rework the item without needing to be aware of that particular behavior.

explicit > implicit

PR #893, opened 2017-05-27
Make Lucience a fake toggle so that it doesn't reset to regen mode on death

Trildar:
Also tweak muting logic for multiple Luciences in inventory.


chrisinajar:
conflicts

Trildar:
Should be fixed

PR #911, opened 2017-05-27
Nerf creeps for release

Haganeko:
Also undid front loaded XP and removed the player number XP factor.

### DIFF
--- a/game/scripts/vscripts/components/cave/cave_types.lua
+++ b/game/scripts/vscripts/components/cave/cave_types.lua
@@ -32,12 +32,12 @@ CaveTypes = {
   [1] = { -- 1 "Howl's it Going?"
     {                                            --HP  MANA  DMG   ARM   GOLD  EXP RESIST
       units = {
-        {"npc_dota_neutral_alpha_wolf",           400,  0,    45,   1,    32,  100, 25},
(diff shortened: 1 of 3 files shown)


chrisinajar:
I assume that the new equations make up for the hugely nerfed exp?

chrisinajar:
Also why undo frontloaded xp? isn't that a good thing? we just want the creeps to be easier to kill but otherwise unmodified. I could easily just add a..
```
if minute < 10 then
  armor = armor * (20 - minute) * 0.05
  damage = damage * (20 - minute) * 0.05
end
```

Haganeko:
After minute six or so. The old version gave you a boatload of XP at minute 0 and kept it there for some time. This progresses earlier. But please do plot the expected values of 98, 123, 98 vs the old ones of 240, 360, 240.

Haganeko:
Well, people are going to kill creeps way more easily. The old version lets you reach level 2 from clearing a single easy camp, once! This one will mean that for everyone to get their levels, they'll need to use everything they can, including the farming cave and the occasional kill.

chrisinajar:
Seems legit, lets give it a shot! Test game anyone @here? :D

PR #916, opened 2017-05-28
Remove Shielder AI

KarmaticNeutral:
Baumi expressed desire for an AI free shielder and none of this worked anyway


Trildar:
I think this would cause the Boss to have literally no AI. As in no leash either. Not too familiar with the Boss system, but I think invoking the basic Boss AI from shielder.lua might be the best way if it's possible.

chrisinajar:
nah, when your base class in neutral you get neutral creep AI, it's base creature that does nothing

Trildar:
Huh. I assumed all the bosses used creature as base class. Could that have anything to do with why the custom AI wasn't working?

Haganeko:
What was wrong with the AS?

PR #922, opened 2017-05-28
Satanic Core tooltip cleanup

Trildar:
Remove repeated CDR values in Satanic Core KV so that it doesn't say "25/25/25" in tooltip.

Make Satanic Core use its own modifier for active effect rather than the built-in modifier_item_satanic_unholy. Mainly so that the modifier tooltip can be changed.


game/scripts/vscripts/items/satanic_core.lua
```diff
@@ -124,3 +125,23 @@ function modifier_item_satanic_core:OnTakeDamage( kv )
 end
 
 --------------------------------------------------------------------------------
+
+modifier_satanic_core_unholy = class({})
+
+function modifier_satanic_core_unholy:DeclareFunctions()
+  return {
+    MODIFIER_PROPERTY_TOOLTIP
```
chrisinajar: Did the old modifier really do nothing? I assumed it added lifesteal..

Trildar: `modifier_item_satanic_core` increased lifesteal based on the presence of the active modifier. So if it did add lifesteal it was causing double lifesteal anyway.

PR #925, opened 2017-05-28
KotL Illuminate value fix

Minnakht:
Wasn't actually scaling damage up at higher ranks. This also very
slightly buffs rank 6 illuminate, kotl op now


Minnakht:
Actually, it's the amount of damage spell amp from 2 intelligence would provide, not 5. Haga pls I thought you'd have done the math

Haganeko:
5/1500 is 0.3%, no? Much akin to 5/14 * 1%

PR #929, opened 2017-05-28
Be quiet

Chronophylos:


Minnakht:
chrono I don't think that's how you comment things out in lua

Chronophylos:
Thanks

Minnakht:
it's apparently `--` to comment something out because lua's gotta be different, but, well, this works too

Chronophylos:
eh

PR #932, opened 2017-05-28
bloodthorne in shop

warpdragon:
bloodthorne in shop


imaGecko:
bloodsthorne

warpdragon:
don't worry i spelled it right in the oaa_shop.txt

PR #937, opened 2017-05-28
Reactive Item fixes

### LINKED ISSUE #619: Make Invulnerability and Reflection Shards block everything (or as much as possible)
The Reactive Items mentioned in the title currently only block damage rather than blocking all effects. Using actual Dota invulnerability is likely undesirable as it would prevent most targeting during the buff period. Unless something better can be thought of, the Bubble Orb method of using a `ModifierGained` filter and `AbsorbSpell` is probably a good place to start.

Magic Immunity could also be used, possibly instead of the `ModifierGained` filter. This would require figuring out a way to dodge attacks fully instead of just blocking the damage though, in order to prevent attack effects.

Trildar:
Closes #619.
All Reactive Items should now block all enemy spell effects and all damage for their duration. Reflection Shards should actually reflect things now (targeted spells only).

Also added effects for the Teleportation Shard blink.

Don't know if this actually makes the Reactive Items any good now. I'll be a little sad now if the whole class of Reflex Item ends up getting redesigned still. :p


chrisinajar:
This looks awesome

PR #942, opened 2017-05-29
Simple Bosses R 2 Weak

warpdragon:
Nuff Said


Minnakht:
also r 2 smol

they are not even pupper

warpdragon:
I also increased the model size 2x

warpdragon:
Google Doc in Balance Private if anyone wants to make any further proposed adjustments.

PR #967, opened 2017-05-30
Added real Cliff Walk to all Bosses

Chronophylos:


Trildar:
Yay for no more floaty bosses.

PR #980, opened 2017-05-31
Prevent Disconnected Heroes from gaining XP

Chronophylos:
I haven't tested this, but  this should prevent dc'ed heroes from gaining xp

### DIFF
--- a/game/scripts/vscripts/components/progression/hero_progression.lua
+++ b/game/scripts/vscripts/components/progression/hero_progression.lua
@@ -34,6 +34,7 @@ function HeroProgression:Init()
     "Intellect"
   }
 
+  FilterManager:AddFilter(FilterManager.ModifyExperience, self, Dynamic_Wrap(HeroProgression, "ExperienceFilter"))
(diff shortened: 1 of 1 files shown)


game/scripts/vscripts/components/progression/hero_progression.lua
```diff
@@ -155,3 +156,9 @@ function HeroProgression:OnLevelUpChatCmd(keys)
   DebugPrint('Levelling up ' .. hero:GetName() .. ' now at level ' .. hero:GetLevel())
   hero:HeroLevelUp(true)
 end
+
+function HeroProgression:ExperienceFilter(keys)
+  local playerID = keys.player_id_const
+
+  return PlayerResource:GetConnectionState(playerID) ~= DOTA_CONNECTION_STATE_DISCONNECTED
```
Trildar: Should have looked at this sooner, but anyway, I think this should be `== DOTA_CONNECTION_STATE_CONNECTED` because players could also be abandoned or some other state.

PR #983, opened 2017-05-31
Improve test cases to catch common KV problems

chrisinajar:


Chronophylos:
Can you also display which IDs are free to use?

warpdragon:
^ OH YES plese that would be amazing

chrisinajar:
Done, also checks for duplicate item definitions (there was one, spectacula was defined twice for no reason)

PR #1000, opened 2017-06-03
Fix Fountain auto-doom

Trildar:
Should now perform its intended purpose of disabling passives on heroes hit by Fountain. Also gave Fountain True Strike.

P.S. We should discuss whether we want the Fountain to outright kill everything. Including through things like Borrowed Time and Ghost Scepter.

P.P.S. Woo! #1000!


Chronophylos:
Why does the damn modifier work for you? It doesn't for me and I have roughly the same file.
Valve pls

imaGecko:
nobody even commented that this was PR #1000

Chronophylos:
> P.P.S. Woo! #1000!

Haganeko:
> nobody even commented that this was PR #1000
Shame, Gecko.

PR #1002, opened 2017-06-03
Farming cave progression

Haganeko:
Changed for 8 minutes per clear to 10.


Haganeko:
I'm putting this on hold because there was a bug.

PR #1003, opened 2017-06-03
Fixed the farming cave

Haganeko:
There was a weird bug. Thanks to Squawky for reporting, Karmatic for testing and Trildar for helping me find it.


Haganeko:
PS Thanks Melon for exploiting bugs.

PR #1004, opened 2017-06-03
Fix higher level Midas not giving bonus gold and XP

Trildar:
Also made levels of Midas share cooldown and make Midas not refreshable by Refresher Orb.


Trildar:
Doesn't give Reliable Gold. But not sure whether we really want to care anyway. Infinite Gold system doesn't really differentiate the two once Gold passes in and out of it.

Trildar:
More importantly, can target Necronomicon units, which vanilla Hand of Midas doesn't. Will fix later.

PR #1016, opened 2017-06-04
Better fountain

Chronophylos:
Made Fountain attack only enemies inside the base
Kills them after 2 seconds by removing Health and Mana
Purges all Buffs
Plays Visual Effects (no sound)
Attacks only enemy Heroes or summoned Units, but no Creeps or Bosses


Chronophylos:
I'll redo this a ability

PR #1023, opened 2017-06-05
Removed Dragon magic immunity

Haganeko:
I don't expect this to be a good change, but the majority has spoken.


Haganeko:
Marked as WIP so Baumi can actually say "yes" or "no" to it before it gets merged.

Haganeko:
I believe this one got the go ahead. Waiting on the other one.

PR #1025, opened 2017-06-05
Reimplement Expanding Shield and Particles

KarmaticNeutral:


warpdragon:
Yooo

warpdragon:
What work does this branch need to work?

KarmaticNeutral:
Removed the particle bits from here because I did it better in the spell rather than do it on the entity itself which makes way more sense

warpdragon:
This is gonna conflict

warpdragon:
How should we handle this? My latest pr adds code to modifier_shielder_shieldex.lua

chrisinajar:
Is this still needed? It doesn't appear to conflict, I just don't know what it does...

KarmaticNeutral:
We very much need this. Currently the shielded doesn't have phase control cause I'm an idiot. This fixes that.

PR #1030, opened 2017-06-05
Fix Martyr's Mail Infinite Echo Damage

### LINKED ISSUE #1029: Martyr's Mail can cause infinite damage echo on self
Martyr's Mail doesn't ignore damage dealt to yourself, which can cause infinite damage instances which will either kill you or, depending on the situation, crash the game. 

How to recreate - 

* Pick a hero who can do self-damage - centaur for now.
* Max out Double Edge, get Martyr's Mail. DO NOT SKILL ANY TALENTS FOR NOW.
* Use MM followed by Double Edge. You'll be denied. Check the log, you'll find this - 

* Next, skill talents and get the +10%magic resist talent.
* MM and Double Edge again, game crashes. 

All _echoed_ damage instances of MM happen on the same server tick. If the echoed double edge kills you within a few bounces (like my screenshot), no issues, you die
If you have sufficient reduction to reduce the double edge echoed damage to less than 1 eventually, you create infinite echoes that can't kill you, in a single tick

First instance reported by _Юра Лох_ on Discord, I dug deeper.

Chronophylos:
I think that should fix #1029

### DIFF
--- a/game/scripts/vscripts/items/martyrs_mail.lua
+++ b/game/scripts/vscripts/items/martyrs_mail.lua
@@ -120,6 +120,11 @@ end
 function modifier_item_martyrs_mail_martyr_active:OnTakeDamage( kv )
 	if IsServer() then
 		local hCaster = self:GetParent()
+    
+    -- Prevent Infinite Loop
+    if kv.attacker == hCaster then
+      return
+    end
 
 		if kv.unit == hCaster then
 			local damageTable = {
--- a/game/scripts/vscripts/items/martyrs_mail.lua
+++ b/game/scripts/vscripts/items/martyrs_mail.lua
@@ -120,7 +120,7 @@ end
 function modifier_item_martyrs_mail_martyr_active:OnTakeDamage( kv )
 	if IsServer() then
 		local hCaster = self:GetParent()
-    
+
     -- Prevent Infinite Loop
     if kv.attacker == hCaster then
       return


imaGecko:
does this also fix martyrs mail reflecting reflected damage?
That also causes infinite damage

CrazyByDefault:
@imaGecko Guess not. That damage can be prevented by adding a no-reflect flag to the outgoing damage on Martyr's Mail

Chronophylos:
I can close this PR if you want to do this (adding a no-reflect flag)

CrazyByDefault:
Nah, I think you should do it. I'd love to contribute and I definitely plan on, but I'm still new to Lua and scripting for Dota. 

You should see more of me around, however 😄

imaGecko:
Always in need for more lua people

CrazyByDefault:
And the world is always in need for better custom games :D

CrazyByDefault:
@Chronophylos Also, checked for the no-reflect flag in the docs. Seems to be missing.
image


Btw is there a separate discord/slack/gitter or something for devs? Or is it just the Minimis one?

Trildar:
@CrazyByDefault I believe `DOTA_DAMAGE_FLAG_REFLECTION` is what we want actually. Though I'm not sure if you'd set it to true or false.
As for the Discord, there's a link somewhere in the docs for the OAA Discord.

warpdragon:
He found it!

PR #1031, opened 2017-06-05
New TAP formatter

Chronophylos:
we still need to reduce the actual amount of tests

example of formatter
example of formatter


chrisinajar:
Yeah, it creates a test for each item, I guess it really doesn't need to do that.

PR #1035, opened 2017-06-06
Fancy Lazor

Chronophylos:


chrisinajar:
rebase

chrisinajar:
There's concern this will be laggy, on discord we discussed having a particle that emits this instead of producing a new particle every 0.1 seconds

allanvbhg:
We should test both imo. See what looks cooler and what lags more.

Chronophylos:
I will change this and the fountain attack to auras

PR #1040, opened 2017-06-06
Give modifier_intrinsic_multiplexer MODIFIER_ATTRIBUTE_MULTIPLE

Trildar:


chrisinajar:
Why? It doesn't give attributes...

Trildar:
MODIFIER_ATTRIBUTE_MULTIPLE means you can have multiple of the same modifier on a unit. It defaults to only allowing one instance otherwise.

PR #1042, opened 2017-06-06
Blink Fix

imaGecko:
Buying Blink should no longer crash the game


Haganeko:
What was the problem?

imaGecko:
blink was missing a baseclass and the recipes were blinks

PR #1056, opened 2017-06-07
Various Greater Guardian Greaves fixes and tweaks

Trildar:
Most notably, the heal/mana numbers that appear over units affected by the active should more closely match vanilla behaviour now. And Greater GG actually describes its farming effect in tooltip now. Sadly, the aura attributes had to be cut from the bottom section of the tooltip due to an apparent character limit (still listed in description).


Chronophylos:
You need to fix these https://travis-ci.org/OpenAngelArena/oaa/builds/240317316#L1548-L1588

PR #1058, opened 2017-06-07
Make (hopefully) all icon filenames lowercase

### LINKED ISSUE #1055: Missing icons in new items
mask of madness 2-5 and aether lens 3 are all missing.

Trildar:
Should fix #1055. Turns out the icons were there. The names just didn't have the same case.


Chronophylos:
`fail: 0` seems you got them all

chrisinajar:
fwiw all that matters is that the case matches, not that they're lower case.

Trildar:
Yeah, but it was bugging me that there were random capitalised names.

PR #1065, opened 2017-06-07
Create Fountain Laser particle

Trildar:
Essentially a continuous version of the TI6 Laguna Blade particle. Implementation of particle to come. Probably by Chrono since he's reworking the Fountain Laser ability.


KarmaticNeutral:
is there something wrong with what we are currently using? I really like it

Haganeko:
What we are currently using is a way more resource intensive version of what Trildar has made - essentially, the current version creates a particle every .1s, which is a bit taxing on the engine.

Trildar:
It should look essentially the same. It's just that we are currently spawning the Laguna Blade particle every 0.1 seconds, which is unnecessary particle spam.

KarmaticNeutral:
ah okay so an efficiency issue. Sounds good.

Chronophylos:
Yaay Thank you!

Chronophylos:
Probably should merge this as I merged it in my `better-better-foutain` branch which I'll create a PR from

PR #1068, opened 2017-06-07
Pips

hallelsegel:
just a quick pips pr


imaGecko:
the pips are wrong. highest vanilla item should get 1/6 then 2/6 to 6/6 for the 5 farming upgrades

KarmaticNeutral:
highest vanilla = 1/6
add Fcore = 2/6
add UC1 = 3/6
etc. Gecko is right

Trildar:
You need to change the KVs in the files in `game/scripts/npc/items/farming`. They all currently point to the `custom/greater_arcanes`. Just change the value path after `AbilityTextureName`.

hallelsegel:
change the first one to "_1" or leave it as it is?

hallelsegel:
and where is regular arcanes?

imaGecko:
https://github.com/dotabuff/d2vpkr/blob/master/dota/resource/flash3/images/items/arcane_boots.png

Trildar:
Match the file name of the icon file. So if you put _1 for that, put it in the KV as well. We don't have a KV file for vanilla Arcane Boots, so just put the icon for it in `items` instead of `items/custom`. Should work.

hallelsegel:
and the txt? all the other boots have files and arcanes don't as far as i can see

hallelsegel:
ok i see thanks

chrisinajar:
i'm already fixing the KV in my test case branch, just do the base arcanes

PR #1074, opened 2017-06-08
Changed Arcane Progression to be in line with other boots

Haganeko:
ImaGecko pointed out Treads on intelligence give more mana than Arcanes; this makes no sense.
Waiting for balance review.


Trildar:
You need to change Greater Guardian Greaves as well, or they'll give less mana than Arcane Boots. :P

Haganeko:
I am honestly OK with that, but I guess it makes sense.

l34Um1:
This all seems fine, it's not very significant.

PR #1075, opened 2017-06-08
Fix Greaves Sidegrade

Haganeko:
I was told it was decided on a Town Hall that all side grade costs should be the same as vanilla; this should fix that.


warpdragon:
According to Travis:
✖ there are no missing tooltips
shouldn't that say "there are missing tooltips", Chris?

Haganeko:
It's failing the "don't have missing tooltips" check

warpdragon:
Basically, because the name "item_greater_guardian_greaves_dummy.txt" doesn't follow the <item_name_level> pattern as regular GGG, it needs its own tooltip file - I think.

Haganeko:
Thank you for the assistance... way ahead of you, though.

warpdragon:
✖ can run parse script [pass: 0, fail: 1, duration: 89ms]

no idea about that one, friend.

Haganeko:
Found it missing "

warpdragon:
oh i see. parse means "syntax bonkered" got it

warpdragon:
All checks have passed
1 successful check
HELL YES

PR #1076, opened 2017-06-08
Test item upgrade paths

chrisinajar:


BotDifficulty:
I assume "aura_negative_armor" is in Dota, not as a custom whateveritscalled. Just browsing around in steam. There's complaints that some tooltips don't show up/arent updated, but maybe they're just crazy or maybe not. Trying to help, going around the code, maybe debug items (because that seems ez). It's probably lowest priority, but I'm still snooping around steam. Thanks.

could be negative armour doesn't have that % thing innately. might need a custom script. cant even find the code for aura_negative_armor so feck me. i imagine its just an attach aura, and im not familiar enough with dota to see if its the same with sf .A. i dont even know where item abilities are implemented and from where...
ill show myself out

Haganeko:
Thanks, but this may not be the place for that post.

BotDifficulty:
No, I mean I could do it myself. Not immediately though. Yeah I figured out where everything was. Uh, bye.

PR #1077, opened 2017-06-08
Better better fountain

Chronophylos:
Reworked Fountain Attack as a Aura
Added Developer Version of Fountain Attack the Dev Lazer available with -lazer

Video


chrisinajar:
There's a dota method to tell if an entity is touching a zone, pretty sure zone control exposes it

Chronophylos:
I know but I couldn't get it to work. It always returned false

PR #1084, opened 2017-06-08
spellicons and a tooltip

imaGecko:
Fixed a bunch of spellicons


Trildar:
The info about the sidegrade is already in the description text.

imaGecko:
I completely missed that somehow

PR #1088, opened 2017-06-08
Bristleback rewrite or how ever you want to call it

Chronophylos:
I copied the Bristleback from [Pizzalol's SpellLibrary](https://github.com/Pizzalol/SpellLibrary)
[KV](https://github.com/Pizzalol/SpellLibrary/blob/master/game/scripts/npc/abilities/bristleback/bristleback_datadriven.txt) and [Lua file](https://github.com/Pizzalol/SpellLibrary/blob/master/game/scripts/vscripts/heroes/hero_bristleback/bristleback.lua)
I modified it a bit.
The result is a Bristleback that respects the Cooldown of Quill Spray


game/scripts/vscripts/abilities/oaa_bristleback.lua
```diff
@@ -0,0 +1,103 @@
...
+
+  local quill_spray = params.unit:FindAbilityByName("bristleback_quill_spray")
+
+  -- If the amount of damage taken since the last Quill Spray proc is equal to or exceeds what's defined as the threshold, release a Quill Spray.
+  if params.unit.quill_threshold_counter >= ability:GetSpecialValueFor("quill_release_threshold") and quill_spray:IsCooldownReady() then
+    -- This should be Quill Spray, but in case something weird like AD is going on, we'll check anyway.
+    quill_spray:CastAbility()
+
+    -- I'm not entirely sure if this is how Bristleback actually works, but this seems like a safe bet.
+    params.unit.quill_threshold_counter = params.unit.quill_threshold_counter - 250.0
```
chrisinajar: We should set it to zero since we're cooling down the quills. This will cause bristleback to store all of the damage he takes and slowly emit all of the quills that he would have emitted.

game/scripts/vscripts/abilities/oaa_bristleback.lua
```diff
@@ -0,0 +1,102 @@
...
+        -- Increase the Quill Spray damage counter based on how much damage was done *post-Bristleback mitigation*.
+        params.unit.quill_threshold_counter = params.unit.quill_threshold_counter + (params.Damage - (params.Damage * side_reduction_percentage))
+      end
+    end
+  end
+
+  local quill_spray = params.unit:FindAbilityByName("bristleback_quill_spray")
+
+  -- If the amount of damage taken since the last Quill Spray proc is equal to or exceeds what's defined as the threshold, release a Quill Spray.
+  if params.unit.quill_threshold_counter >= ability:GetSpecialValueFor("quill_release_threshold") and quill_spray:IsCooldownReady() then
```
chrisinajar: This is checking on quill_spray's cooldown... it should check it's own cooldown, and also go onto cooldown when it happens. That way the passive and the active have their own cooldowns. I believe the old KV has a cooldown value in it for this purpose.

Chronophylos:
This should do it

game/scripts/vscripts/abilities/oaa_bristleback.lua
```diff
@@ -0,0 +1,108 @@
...
+      end
+    end
+  end
+
+  local quill_spray = params.unit:FindAbilityByName("bristleback_quill_spray")
+
+  -- If the amount of damage taken since the last Quill Spray proc is equal to or exceeds what's defined as the threshold, release a Quill Spray.
+  if params.unit.quill_threshold_counter >= ability:GetSpecialValueFor("quill_release_threshold") and ability:IsCooldownReady() then
+    if quill_spray:IsCooldownReady() then
+      quill_spray:EndCooldown()
```
chrisinajar: I think you're missing a CastAbility line before this EndCooldown

game/scripts/vscripts/abilities/oaa_bristleback.lua
```diff
@@ -0,0 +1,108 @@
...
+    else
+      local cooldown = quill_spray:GetCooldownTimeRemaining()
+      quill_spray:CastAbility()
```
Trildar: Try calling `quill_spray:OnSpellStart` instead. I think it possibly doesn't trigger the cooldown on the spell? Not too sure.

game/scripts/vscripts/abilities/oaa_bristleback.lua
```diff
@@ -0,0 +1,108 @@
...
+        -- Set Control Point 1 for the back damage particle; this controls where it's positioned in the world. In this case, it should be positioned on Bristleback.
+        ParticleManager:SetParticleControlEnt(back_damage_particle, 1, params.unit, PATTACH_POINT_FOLLOW, "attach_hitloc", params.unit:GetAbsOrigin(), true)
+        -- Increase the Quill Spray damage counter based on how much damage was done *post-Bristleback mitigation*.
+        params.unit.quill_threshold_counter = params.unit.quill_threshold_counter + (params.Damage - (params.Damage * back_reduction_percentage))
+      end
+    else
+      -- Check if the reduced damage is lethal
+      if (params.unit:GetHealth() - (params.Damage * (1 - side_reduction_percentage))) >= 1 then
+        -- This is the actual "damage reduction".
+        params.unit:SetHealth((params.Damage * side_reduction_percentage) + params.unit:GetHealth())
```
Trildar: This is kind of a hack. Should probably rewrite the ability to use `MODIFIER_PROPERTY_INCOMING_DAMAGE_PERCENTAGE`

game/scripts/vscripts/abilities/oaa_bristleback.lua
```diff
@@ -0,0 +1,108 @@
...
+        -- Increase the Quill Spray damage counter based on how much damage was done *post-Bristleback mitigation*.
+        params.unit.quill_threshold_counter = params.unit.quill_threshold_counter + (params.Damage - (params.Damage * side_reduction_percentage))
+      end
+    end
+  end
+
+  local quill_spray = params.unit:FindAbilityByName("bristleback_quill_spray")
+
+  -- If the amount of damage taken since the last Quill Spray proc is equal to or exceeds what's defined as the threshold, release a Quill Spray.
+  if params.unit.quill_threshold_counter >= ability:GetSpecialValueFor("quill_release_threshold") and ability:IsCooldownReady() then
```
Trildar: This results in all damage during the cooldown being stored, such that if the threshold is reached during the cooldown, any amount of damage once it's off cooldown will trigger it. Not sure that's really what we want.

Haganeko: Is that the same as what Chris pointed out above? Before the "set to zero" commit.

Trildar: Not really. That one was an issue of how the stored damage reset. Even vanilla behaves like set to zero I'm pretty sure, so I'm not sure why it was like that in the first place.
What happens with that code would be that if Bristle took 5 times the threshold damage in the back, he would trigger quills for any damage the next 4 times it was off cooldown, in addition to the initial proc.

Haganeko: If vanilla worked like that, I think he wouldn't proc over 10 times from a single attack in OAA like he does. But I catch your point.

Trildar: Nah, it's just too many high damage hits. Just tested it. If you Dagon 9 Bristle and then hit him afterwards, you just get a single proc from the Dagon 9 and nothing afterwards.

game/scripts/vscripts/abilities/oaa_bristleback.lua
chrisinajar: MODIFIER_PROPERTY_INCOMING_DAMAGE_PERCENTAGE sucks

Haganeko: It is very akin to what Bristleback uses, though - it should stack additively with damage manipulators.

game/scripts/vscripts/abilities/oaa_bristleback.lua
Chronophylos: I just found this http://moddota.com/forums/discussion/60/calling-spells-with-setcursor

PR #1093, opened 2017-06-09
Rewrote Dagon and Added Spell Amp

Chronophylos:
Added more Stats to Modifier Generic Bonus


game/scripts/vscripts/items/dagon.lua
```diff
@@ -0,0 +1,47 @@
...
+
+  local particleName = "particles/items/dagon_oaa.vpcf"
+  local particleThickness = 300 + (100 * level) --Control Point 2 in Dagon's particle effect takes a number between 400 and 2000, depending on its level.
+
+  local damage = self:GetSpecialValueFor("damage")
+
+  local particle = ParticleManager:CreateParticle(particleName,  PATTACH_POINT_FOLLOW, caster)
+  ParticleManager:SetParticleControlEnt(particle, 0, caster, PATTACH_POINT_FOLLOW, "attach_attack1", caster:GetOrigin(), true)
+  ParticleManager:SetParticleControlEnt(particle, 1, target, PATTACH_POINT_FOLLOW, "attach_hitloc", target:GetOrigin(), true)
+  ParticleManager:SetParticleControl(particle, 2, Vector(particleThickness))
```
Trildar: I don't know if it's actually required, but I've always done `ParticleManager:ReleaseParticleIndex` after I'm done with a particle.

Trildar: There also seems to be some extra particle that Dagon 5 applies to its target or something, but I'm not sure how to use it.

game/scripts/vscripts/items/dagon.lua
```diff
@@ -0,0 +1,47 @@
...
+
+
+  caster:EmitSound("DOTA_Item.Dagon.Activate")
+  target:EmitSound("DOTA_Item.Dagon5.Target")
```
Trildar: Only doing the `Dagon5` sound for level 5 and up may be more accurate, but it's not really that important.

Chronophylos: In Pizzalol's Spell Library they were using both

game/scripts/vscripts/items/dagon.lua
Chronophylos: I've never used it
but it may be better to use

game/scripts/vscripts/items/dagon.lua
Trildar: Well, the Bristleback implementation wasn't entirely accurate either with the damage storage, though to be fair, that one had a comment explicitly saying the author wasn't sure if that's how Bristleback worked.

Chronophylos: Maybe `DOTA_Item.Dagon.Activate` is always played and `DOTA_Item.Dagon5.Target` for Level 5 and above as a additional sound

Chronophylos: Cast Range is increasing too and both sounds have a limited distance

PR #1102, opened 2017-06-09
Refresher Core and modifier_octarine_vampirism_applier fixes

### LINKED ISSUE #1072: Bug Arc Warden can duplicate his illusions infinetely by using Refresher Core

Trildar:
Fixes #1072.
Some restructuring of `modifier_octarine_vampirism_applier`. Also made it do the typical force refresh I've been putting on practically every aura modifier to prevent spell lifesteal possibly not upgrading properly.


chrisinajar:
Conflicts. Also we never ever use this as an aura, that's an artifact of where we copy and pasted it from.

Trildar:
It still needs to be an aura anyway. I can't really think of a better way to make spell lifesteal non-stacking. I think even now it actually stacks with Octarine Core 1.

Trildar:
Actually, I guess something like Heart Transplant's method for not stacking Health Regeneration would work.

Trildar:
Rebased. Will probably rework the spell vamp at some later time.

chrisinajar:
`modifier_intrinsic_multiplexer` is so good for code clarity, i love it. Gonna merge this and the other so it can start getting tested on local instances.

PR #1107, opened 2017-06-09
Prevent Lazor Particle from Growing

Chronophylos:


Trildar:
Pretty sure this actually affects the smoke. You probably want to look at messing around with `fountain_lazor_sphere`.

Chronophylos:
In the particle editor the white sphere stopped growing

Trildar:
Probably incidental. I just checked, and disabling that operator in the parent particle makes the smoke particles really tiny because they don't get their radius set. Try disabling the children particles if you want to check for yourself since the smoke can be hard to see with everything else going on.

Chronophylos:
I'll look into it

PR #1120, opened 2017-06-11
Handle reconnecting players during duels better

chrisinajar:


Chronophylos:
Is it `keys.userid` or `keys.playerid`?

chrisinajar:
It's `keys.PlayerID`!

Chronophylos:
Here it's `keys.userid` https://github.com/OpenAngelArena/oaa/blob/master/game/scripts/vscripts/events.lua#L16

PR #1128, opened 2017-06-11
Shielder Shield

warpdragon:
Hex particle based shield for shielder boss
Has 3 different sizes for 3 phases (still to be implemented, but particle is ready)
Shield particle system represents the exact angles of the damage reduction shield.
Decay animation leaves something to be desired. Still I think it's good enough.
Video related: https://youtu.be/4MWXK95VxQc


allanvbhg:
This is just phase 1 particles right?

PR #1132, opened 2017-06-12
Refactor calls to math.random to use Dota API random functions instead

Trildar:
If I did this right, most of this should be largely inconsequential. The main exceptions being the Charger tower locations, which looks like it was using math.random in such a way that the location vector could only be 1 of 9 (or 8 rather, since one of the possible vectors was (0,0), which would get rejected) locations, and ward_invisiblity which was using math.random in a way it's not really meant to be used.

### DIFF
--- a/game/scripts/vscripts/abilities/stopfightingyourself/dupe_heroes.lua
+++ b/game/scripts/vscripts/abilities/stopfightingyourself/dupe_heroes.lua
@@ -143,7 +143,7 @@ function boss_stopfightingyourself_dupe_heroes:OnSpellStart()
 
       -- Randomly play sound to player (10% chance)
       -- NOTE this doesn't seem to work
-      if math.random(100) <= 10 then
+      if RandomFloat(0, 1) < 0.1 then
         --EmitAnnouncerSoundForPlayer('sounds/vo/announcer_dlc_rick_and_morty/generic_illusion_based_hero_02.vsnd', unit:GetPlayerID())
         illusion:EmitSound('sounds/vo/announcer_dlc_rick_and_morty/generic_illusion_based_hero_02.vsnd')
       end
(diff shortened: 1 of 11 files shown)


Haganeko:
So Charger was taking a random spot in a square then protecting that onto a circle? That explains the excessive spawning of towers in the corners...

Trildar:
More like it was choosing from 1 of 8 points on an 800 radius circle around it for each tower spawn.

Haganeko:
Oh. I assume math.random only does integers? Makes sense. Could theoretically have reduced the number of rand calls by randomizing the angle instead.

Trildar:
Depends on how it's called. `math.random` does either a floating point number [0, 1) or an integer in some given interval. Calling it as `math.random(-1, 1)` makes it just return -1, 0, or 1.

Haganeko:
I see. If you don't mind me asking, what is that rand call in wards?

Trildar:
It was using it to generate a unique ID for each ward basically. There's a much better function for that in the Dota API, `DoUniqueString`.

Now I'm just wondering why the Charger is apparently able to spawn pillars closer than 700 units to itself. That ought not to happen...

Trildar:
Right. It's probably just a case of trying to spawn pillars near where other pillars already are, thus making them spawn wherever there's space in that area.

Haganeko:
Should this get marked as WIP, since you're still pushing to it?

Trildar:
Nah, should be fine. Just some stuff I missed. Should be good now.

chrisinajar:
The pillars have massive hulls and wont spawn on top of each other. That's why the 8 points still appeared pretty damn random, it would almost never manage to get a tower into the exact spot, they always push eachother around. There's a natural randomness to the distance of the tower because of this, varrying an extra ~300 units or so from the distance desired.

PR #1135, opened 2017-06-12
Set Flags later to make sure they actually get send

Chronophylos:


chrisinajar:
Gotta add `statCollection` to .luacheckrc

Chronophylos:
I'm not even sure if this changes anything
I was told the reason the Version doesn't get send is that is set too late

PR #1141, opened 2017-06-12
Blueberry!!!

warpdragon:
implemented blueberry. first successful model texture modification.


warpdragon:
To be clear, this is only the creep camp unit. This has no effect on the Boss "Killer Tomato"

warpdragon:
Oops forgot pic 
bloob

PR #1147, opened 2017-06-13
VTEX Update

warpdragon:
changed a line in vtex. don't know if this will have any effect.


chrisinajar:
Seems legit, that file exists and the other didn't.

PR #1148, opened 2017-06-13
Master

l34Um1:


Trildar:
Heh. Not sure that works though. I remember reading something about in-built abilities not respecting KV damage types properly. Don't know if that's true in this case.

Haganeko:
That... Seems effective (if it works), albeit crude.

PR #1154, opened 2017-06-14
Slight buff to arc warden

KarmaticNeutral:
This is a slight buff to arc warden in hopes that the hero will become more viable.
Make a viable arcanes strat where you can farm the creeps and get the gold or send the illu off to get assist gold else where
Plus the thing is then alive longer. obviously a buff.

If this is still not enough we can lower cooldown, but I don't like the idea of making the duration any longer than this

Also this is an attempted nerf to silencer as arc is one of his worst matchups in vanilla.

### DIFF
--- a/game/scripts/npc/abilities/arc_warden_tempest_double.txt
+++ b/game/scripts/npc/abilities/arc_warden_tempest_double.txt
@@ -31,7 +31,7 @@
       "01"
       {
         "var_type"                                        "FIELD_FLOAT"
-        "duration"                                        "12 14 16 18 20"
+        "duration"                                        "12 14 16 20 26"
       }
     }
   }
--- a/game/scripts/npc/abilities/arc_warden_tempest_double.txt
+++ b/game/scripts/npc/abilities/arc_warden_tempest_double.txt
@@ -31,7 +31,7 @@
       "01"
       {
         "var_type"                                        "FIELD_FLOAT"
-        "duration"                                        "12 14 16 20 26"
+        "duration"                                        "12 14 16 19 22"
       }
     }
   }


Trildar:
You currently never get creep assist gold from yourself, and I'm not sure we really want to make an exception to that, especially considering how much creep assist gold gives right now. I think the Tempest Double should already carry the aura though, but that needs confirmation.

imaGecko:
This would also need the implementation of creep assist gold being shared

Trildar:
Shared? Between who?

imaGecko:
between everyone who is in range to get the creep assist gold.
So if there's 2 people with ggg in range of a creep they both only get half

Haganeko:
Between all heroes with Arcanes within range of a same creep. Since different tiers are possible, it basically means each player's assist gold should be proportional to 1/n, where n is the number of Arcanes holders in range.

Haganeko:
Well, we might amend the formula to be a bit less punishing (much like kill assist), but for now it's that.

Trildar:
I'm not sure what this has to do with allowing self-assist gold. I assume that's what Gecko was talking about.
I also really dislike that idea. It discourages multiple Arcanes on a team.

imaGecko:
Could be something like (0.8+0.2n)/n
Everyone in range would get 100/55/40/32.5/28

It's exactly to discourage multiple arcanes in the same team

KarmaticNeutral:
I appreciate the discussion of how arcanes should work. Does anyone have an issue with a longer duration on double?

KarmaticNeutral:
Zarnotox - Today at 11:08 AM
your slight buff to arc warden just made him able to spawn a double from his double then xD
since he has a 10% cdr talent

KarmaticNeutral:
That new commit should make it so he can't have a clone summon a clone

PR #1157, opened 2017-06-14
Attempt to Nerf Glaives

KarmaticNeutral:
If this doesn't work I would suggest a damage cap. This should go a long way. 15% of his int to damage reduced at max and less damage along to way making it harder to get kills. This should be BIG

### DIFF
--- a/game/scripts/npc/abilities/silencer_glaives_of_wisdom.txt
+++ b/game/scripts/npc/abilities/silencer_glaives_of_wisdom.txt
@@ -12,7 +12,7 @@
     "AbilityUnitTargetTeam"                               "DOTA_UNIT_TARGET_TEAM_ENEMY"
     "AbilityUnitTargetType"                               "DOTA_UNIT_TARGET_HERO | DOTA_UNIT_TARGET_BASIC"
     "SpellImmunityType"                                   "SPELL_IMMUNITY_ENEMIES_NO"
-    "AbilityUnitDamageType"                               "DAMAGE_TYPE_PHYSICAL"
+    "AbilityUnitDamageType"                               "DAMAGE_TYPE_PURE"
     "HasScepterUpgrade"                                   "1"
 
 
@@ -36,7 +36,7 @@
       "01"
       {
         "var_type"                                        "FIELD_FLOAT"
-        "intellect_damage_pct"                            "35 50 65 80 85 90"
+        "intellect_damage_pct"                            "35 45 55 65 70 75"
       }
       "02"
       {


l34Um1:
This doesn't fix anything. Silencers damage with pure glaives is grossly too high, you have turned him from three shotting people to maybe sometimes four shotting them instead.

Everything in Dota and especially OAA is based on resistances. Usually damage gets reduced by 60-80% in OAA but Silencer gets to bypass it. If he has 300 Int (which is very common) then previously he added 270 pure damage, with this change it's 225. Assuming the enemy has 60% physical resistance the effective physical damage added by glaives was 675, now it is 562. This is still completely OP for a non ultimate ability. Same scenario with physical glaives adds 108 dmg with 90% and 90 dmg with 75%. 

My change is probably too harsh but we can always adjust it to be more reasonable and feel good for the players. They'll level up the spell and be like "fuck yeah 150% dmg" but it doesn't break the game. If we keep it pure we need to reduce it to like 40% which is only slightly more than level 1 of regular Dota. 

It also messes with the base dota values which was the main argument against my change of turning it to physical damage.

KarmaticNeutral:
You're right, this nerf is definitely not enough. But changing these base dota values does significantly less toward making the hero feel and play differently than changing his damage to physical does. When you change damage types from pure to physical silencer stops wanting to target high Armor tanks in a fight and wants to pick off supports that have little armor. The entire mindset and entire playstyle of the hero changes. I have nothing against having alternative builds on heroes but this doesn't do that. It removes one build and puts another in its place. You aren't nerfing a hero, you are reimplementing it. 

Silencer needs to be nerfed. I don't think physical damage is the answer. This is what I can see that can be done. I would love to hear alternative _solutions_. I don't think these changes will be enough but it's something and the heroes key mechanic is preserved.

KarmaticNeutral:
We could make it scale 35/40/45/50/55/60 if you like. Still very strong due to increased int levels in this game mode but caps it at a 40% reduction of int to damage. I would support that.

Chronophylos:
He gains permanent Intelligence, right? Could we make it run out so it has a duration?

Minnakht:
I don't think that'd matter much, Chrono - he essentially has "Flesh Heap but 2 intelligence", which over a game would stack up to... 40 stacks, maybe? I guess it's a fair amount... 

That said, I'm not too sure how is that alterable.

Haganeko:
Getting the cap seems best, past this.

game/scripts/npc/abilities/silencer_glaives_of_wisdom.txt
```diff
@@ -12,7 +12,7 @@
     "AbilityUnitTargetTeam"                               "DOTA_UNIT_TARGET_TEAM_ENEMY"
     "AbilityUnitTargetType"                               "DOTA_UNIT_TARGET_HERO | DOTA_UNIT_TARGET_BASIC"
     "SpellImmunityType"                                   "SPELL_IMMUNITY_ENEMIES_NO"
-    "AbilityUnitDamageType"                               "DAMAGE_TYPE_PHYSICAL"
```
chrisinajar: I don't think the damage nerfs are enough to make up for changing the damage back from physical to pure.

Trildar: On that note though, I don't think the change to physical worked. I tested hitting a unit with Stoneskin on and off with a Silencer with max level Glaives and some 400 int, and Stoneskin hardly changed the damage.

Haganeko:
The change to physical, as has been repeatedly pointed out on discord, changes how the hero is played. By a lot.
Honestly, I believe this is partly a consequence of us going back on the initial idea of making attributes on items scarce. 
In any case, this reduces the amount of pure damage he deals from 180% of int to 150%. That's a 17% nerf, relatively. The one Karmatic proposed would reduce it further to 120%. At that point we've essentially removed the aghs, in that the aghs value is nearly the old value without. 
I believe we should get a damage cap on glaives. It's the only way to control the ability without nerfing it into the ground or changing the way it works drastically.

Haganeko:
The current changes may look simple, but they drastically alter how the hero functions. Nerfing numbers is a less invasive nerf.

warpdragon:
A different balance solution was chosen.

Haganeko:
The PR does not predate the chosen solution. It was opened as an alternative to it, as it is more invasive and less elegant; also, we aren't quite sure it even worked.

Trildar:
I think we're fairly certain at this point that changing the KV didn't do anything.

warpdragon:
No reason to keep it open.

Haganeko:
It's the best nerf we have until someone rewrites the thing in lua.

warpdragon:
Some of us have standards.

Haganeko:
Call them, then. They would probably tell you that the less invasive (this one, plus cap) option is best.

warpdragon:
The underscape will not be kind.

PR #1160, opened 2017-06-14
Initial WK aghs nerf

Minnakht:
Apparently wraith delay was being problematic in duels so let's cut down on its duration.


chrisinajar:
Should we just try 7 first?

warpdragon:
7

Minnakht:
The discussion in #balance-help was led under the assumption it was 7 at all ranks, and it was deemed too strong even then. @chrisinajar - you're free to reject this PR and merge in your own where you change it to 7 at all ranks if you'd prefer. It's not like we're making final decisions here, just public testing.

Haganeko:
#1185

Issue #1121, opened 2017-06-11
Upgraded Guardian greaves and Charge bkb


CynicalDemon:
Guardian greaves don't give the correct amount of armor when upgraded to farming greaves or higher (aura and other stats still work fine).

Charged bkb can be upgraded without needing to spend upgrade cores on the magic wand upgrades by clicking on the next bkb recipe instead.

Haganeko:
I believe the second part is intended, if you mean that charge BKB upgrade only costs one core once the item is made.

CynicalDemon:
Yeah that's what I meant. It just seemed off since it said the magic wand needed upgrade cores as well.

Trildar:
The recipe displayed is to go from a regular BKB to charge BKB, which I think actually isn't right because it costs more cores to combine later due to needing high level Magic Wand. Sidegrade probably also shouldn't be the primary recipe.

Trildar:
As for Guardian Greaves, turns out the item was giving the armor bonus the aura was supposed to. And the aura was probably giving nothing.

Thanks for the reports, but in future, it would be better to make separate reports for unrelated issues such as these.

Trildar:
Closed by #1123 and #1163.

PR #1163, opened 2017-06-15
Charge BKB reworks

### LINKED ISSUE #1127: New Charge BKB upgrade ladder

Trildar:
Now a Lua item. Now with a more proper tooltip. Now with more sensible build path.

More importantly though, the charge mechanic has been reworked to reset like Fury Swipes, with a reset time of 10 seconds, rather than decay individually. I'm not sure if this is necessarily what we want, but Haganeko and Minna felt it was sensible enough.

Closes #1127 and #1121.


JRJML:
omg did you really fix the build path?? cuz that thing was so frustrating! :blueberry:

Haganeko:
Dummy items are missing tooltips. Had the same issue with greaves.

Trildar:
Oh right. Those shouldn't even be purchasable in this case though (though it does appear in shop search). Guess I need tooltips anyway just to appease Travis.

imaGecko:
The original recipe should cost 0 and include the dummy recipe in it which has the price. The dummy recipe exists just as an empty item that doesn't do anything and doesn't result in anything as you're not allowed to have 2 recipes result in the same item

Haganeko:
The dummy recipe is purchasable - it's there in place of the actual recipe, which costs zero.

Trildar:
Are you sure? It works in local testing. The only issue I had was that the second recipe wouldn't show build path in shop. The combining still worked. I fixed build path by making the normal recipe also have the same combination as the dummy. Means you have weirdness like BKB3 + Wand + CBKB1 Recipe + CBKB2 Recipe = CBKB2, if for some reason you buy both recipes, but that's strictly worse than the intended combination that doesn't include CBKB2 Recipe.

imaGecko:
You could take a look at ggg

Trildar:
I know how the GGG build path works. I'm specifically trying to do it a different way that doesn't make the Recipe tooltips look different from normal.

Trildar:
@chrisinajar Do we want to keep enforcing the one recipe item for one result thing? What I have here seems to work fine in local testing. I can change it to be like the GGG recipes, but I don't really like that because it makes the recipe tooltips different.

chrisinajar:
define `makes the recipe tooltips different`

Trildar:
image
vs
image

Haganeko:
Oh. That's a question of writing the tooltip - I was lazy and wrote it empty (also I didn't remember that recipes had tooltips)

chrisinajar:
yeah, easily fixable. also any time you copy a tooltip exactly from dota it doesn't send it out for translations, just uses the dota ones

Trildar:
Um. How? The recipes won't have reference to the item AbilitySpecial. Unless I suppose you want to copy that whole block to every dummy item as well. It'd be kind of annoying to have item values repeated some more whenever it comes to changing them.

chrisinajar:
The recipe can just have the ability special values, and we can add a test case that if they exist they must be exactly equal to the item they create. Seems simple, and travis maintains it for us.

Trildar:
Rebased and changed the recipes to pass current Travis standards. I honestly think the amount of duplication that has to be done for the tooltips is silly though. You need the KVs for cooldown, mana cost, item level, item quality, AbilitySpecial, and then you need the localisation KVs for the AbilitySpecial in the tooltip.

PR #1164, opened 2017-06-15
Implement Visage Gravekeepers Cloak in lua

chrisinajar:
Didn't gain stacks passed 4, making upgrading the ability worse. This fixes that.

### DIFF
--- a/game/scripts/vscripts/abilities/visage_gravekeepers_cloak.lua
+++ b/game/scripts/vscripts/abilities/visage_gravekeepers_cloak.lua
@@ -1,4 +1,8 @@
 
+if Debug then
+  Debug.EnabledModules['abilities:*'] = true
+end
+
 LinkLuaModifier("modifier_visage_gravekeepers_cloak_oaa", "abilities/visage_gravekeepers_cloak.lua", LUA_MODIFIER_MOTION_NONE) --- PETH WEFY INPARFANT
 LinkLuaModifier("modifier_visage_gravekeepers_cloak_oaa_aura", "abilities/visage_gravekeepers_cloak.lua", LUA_MODIFIER_MOTION_NONE) --- PETH WEFY INPARFANT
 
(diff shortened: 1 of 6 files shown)


chrisinajar:
I'll do tooltips later

game/scripts/vscripts/abilities/visage_gravekeepers_cloak.lua
```diff
@@ -0,0 +1,109 @@
...
+
+function visage_gravekeepers_cloak_oaa:GetIntrinsicModifierName()
+  return "modifier_visage_gravekeepers_cloak_oaa"
+end
+
+if IsServer() then
+  function visage_gravekeepers_cloak_oaa:OnUpgrade()
+    local caster = self:GetCaster()
+    local mod = caster:FindModifierByName("modifier_visage_gravekeepers_cloak_oaa")
+    mod:SetStackCount(self:GetSpecialValueFor("max_layers"))
```
Trildar: Should only do this when first learned. i.e. when level is 1.

game/scripts/vscripts/abilities/visage_gravekeepers_cloak.lua
```diff
@@ -0,0 +1,109 @@
...
+  end
+end
+
+---------------------------------------------------------------
+
+modifier_visage_gravekeepers_cloak_oaa = class(ModifierBaseClass)
```
Trildar: Needs `IsPurgable` false and `RemoveOnDeath` false.

game/scripts/vscripts/abilities/visage_gravekeepers_cloak.lua
```diff
@@ -0,0 +1,109 @@
...
+  }
+end
+
+function modifier_visage_gravekeepers_cloak_oaa:GetModifierTotal_ConstantBlock(keys)
+  local ability = self:GetAbility()
+  local caster = self:GetCaster()
+  local stackCount = self:GetStackCount()
+  local damageReduction = math.min(100, self:GetAbility():GetSpecialValueFor("damage_reduction") * stackCount)
+  local damageThreshold = self:GetAbility():GetSpecialValueFor("minimum_damage")
+  if keys.attacker:GetTeam() ~= caster:GetTeam() and (keys.attacker:GetTeam() == DOTA_TEAM_GOODGUYS or keys.attacker:GetTeam() == DOTA_TEAM_BADGUYS) then
```
Trildar: I think there are ways to properly check if the attacker is player-owned, but this works too I suppose, though it doesn't technically check for player-owned.

game/scripts/vscripts/abilities/visage_gravekeepers_cloak.lua
```diff
@@ -0,0 +1,109 @@
...
+function modifier_visage_gravekeepers_cloak_oaa:IsAura()
+  return true
+end
+
+function modifier_visage_gravekeepers_cloak_oaa:GetAuraSearchType()
+  return bit.bor(DOTA_UNIT_TARGET_ALL)
+end
+
+function modifier_visage_gravekeepers_cloak_oaa:GetAuraSearchFlags()
+  return bit.bor(DOTA_UNIT_TARGET_FLAG_PLAYER_CONTROLLED)
```
Trildar: Not sure this is all that necessary since the aura only works on the Familiars anyway.

game/scripts/vscripts/abilities/visage_gravekeepers_cloak.lua
```diff
@@ -0,0 +1,109 @@
...
+function modifier_visage_gravekeepers_cloak_oaa:GetAuraEntityReject(entity)
+  return string.sub(entity:GetUnitName(), 0, 24) ~= "npc_dota_visage_familiar"
+end
+
+---------------------------------------------------------------
+
+modifier_visage_gravekeepers_cloak_oaa_aura = class(ModifierBaseClass)
+
+function modifier_visage_gravekeepers_cloak_oaa_aura:GetDuration()
+  return 1
```
Trildar: Is this for aura stickiness? Cause `GetAuraDuration` on the applying modifier is probably better, and the wiki also says that the aura lingers for 0.5 seconds, which is the default aura stickiness.

game/scripts/vscripts/abilities/visage_gravekeepers_cloak.lua
Trildar: Oh wait, doing it only on level 1 would break the higher stacks... Seems weird to just refresh all stacks when upgrading though. Maybe it should check if the max stacks has increased compared to last level and add that many.

chrisinajar: It should probably do something like ```lua
local currentLevel = self:GetLevel()
if currentLevel > 1 then
  local oldMaxStacks = self: GetLevelSpecialValueFor("max_layers", currentLevel - 1)
  local stackDiff = oldMaxStacks - self:GetStackCount()
  self:SetStackCount(self:GetSpecialValueFor("max_layers") - stackDiff)
else
  self:SetStackCount(self:GetSpecialValueFor("max_layers"))
end
```

chrisinajar:
I forgot i was working on this....

Chronophylos:
Are you?

chrisinajar:
Todo: 
* [ ] Add `IsPurgable false and `RemoveOnDeath` false.
* [ ] Add tooltips

PR #1169, opened 2017-06-16
just a particle idea for charger

warpdragon:


Haganeko:
Much ado about warpdragon's commit.

Chronophylos:
can we have a image or a gif?

warpdragon:
image

allanvbhg:
And the blue changing places all the time right? And it emits glowing pieces arouns the cirkels right?

PR #1171, opened 2017-06-16
Swap farming abilities of arcane/ggg and travels

chrisinajar:
Also made test cases run in windows / crappy nix envs


Haganeko:
When was this decided?

Chronophylos:
During a Townhall meeting i think

Minnakht:
@Haganeko I think I recall Baumi saying something to that effect this week, not entirely sure though. 

I'm not sure if I'm entirely happy with this change, but I guess I can be

imaGecko:
I think this makes much more sense, get ggg for fighting and stay back with tps and farm until you can tp in

Haganeko:
The new TP boots tooltip doesn't include a description of Commensalism?

PR #1176, opened 2017-06-17
Tweak the items that the damage loadout command gives

Trildar:


Haganeko:
MKB?

Trildar:
Now gives max level Daedalus, Desolator, Mjollnir, and MKB.

PR #1177, opened 2017-06-17
Implement Splinter Shot Greater Phase Boots

Trildar:
Essentially Drow Aghanim's effect for anyone.
Interacts in a bit of a weird way with Drow Agh's. It basically doesn't stack in any way; only the main attack will produce Greater Phase splinters, and the Greater Phase splinters also don't trigger Drow Agh's effect. So effectively, Greater Phase Boots just increases splinter count on Drow, though the Phase Splinters don't avoid targeting the same units as Drow Agh's, so they can hit the same targets.

Requesting @l34Um1 and the balance team to provide some initial numbers for the effect. Currently uses the same values as Drow and scales damage percentage the same amount that Greater Phase Boots previously scaled.


l34Um1:
Initial numbers:

Split count: 1 2 3 4 5
Split damage: 50%
MS: 50 55 60 65 70
Bonus Damage: 35 70 105 140 210
MS Bonus Melee: 26 29 32 35 38
MS Bonus Ranged: 22 25 28 31 34
Phase duration: 2.5

Trildar:
What about split radius?

Also, due to details of how Drow Splinter procs, Greater Phase splinter can actually be made to proc Drow Splinter, but it would require changing to something more like Medusa's split shot where the projectiles originate from the hero rather than the target.

l34Um1:
I actually think it's fine that the split shot from drow aghs doesn't trigger the one from the boots. 375 search radius at all levels seems fine.

PR #1179, opened 2017-06-17
Fixed Bug where players get stuck in the cave

Chronophylos:
This was caused when players finished the cave and a duel started before the teleport (3 seconds) was finished

Made `MoveCameraToPlayer` a standalone util function


Trildar:
My Git hasn't been ignoring the compile sound events file either. Not sure what's happening there.

PR #1182, opened 2017-06-18
Manta Style Lua Rewrite

Chronophylos:
I have a small Problem with my current code,
for some reason I can't figure out how to define a radius where the illusions spawn.
They spawn at the right angle, but not at the right distance from where the caster was.


game/scripts/vscripts/items/manta.lua
```diff
@@ -0,0 +1,213 @@
...
+    local image = ability.images[imageIndex]
+
+  -- The formation of the owner and the illusions is always the same. One spawns on the owner's cast location and the others randomly on north, east, south or west side each.
+  -- Though the formation is always the same, the owner and the illusions take a random position in the formation and have all the same facing angle.
+    local position = GetImageLocation(origin, casterIndex, false, imageIndex, images_count)
+
+  -- Recasting this replaces the illusions from the previous cast which are currently under the owner's control.
+    if image ~= nil and not image:IsNull() then
+      if image:IsAlive() then
+        position = image:GetAbsOrigin()
```
Trildar: Wouldn't setting `position` to the old illusion's position cause the new one to spawn there? They're supposed to spawn next to you.

game/scripts/vscripts/items/manta.lua
```diff
@@ -0,0 +1,213 @@
...
+   local distance = Vector(360, 360, 0)
+   local theta = (360 / (imageCount - 1)) * imageIndex
+   local cosTheta = math.cos(theta)
+   local sinTheta = math.sin(theta)
+   local position = Vector(0, 0, 0)
+
+   position.x = (origin.x * cosTheta) - (origin.y * sinTheta)
+   position.y = (origin.x * sinTheta) - (origin.y * cosTheta)
+
+   return (origin + position):Normalized()
```
Trildar: I'm pretty sure `Normalized` makes the vector unit length, i.e. length of 1. I'm not sure why this isn't just spawning the illusions right next to (0,0), considering that if it's normalizing (origin + position) it should just return a vector of unit length.

game/scripts/npc/items/item_manta.txt
```diff
@@ -120,12 +122,12 @@
       "14"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "images_take_damage_percent_ranged"               "300"
+        "tdamage_incoming_ranged"                         "300"
```
Chronophylos: yes

game/scripts/vscripts/items/manta.lua
Chronophylos: > Recasting this replaces the illusions from the previous cast which are currently under the owner's control.
from: http://dota2.gamepedia.com/Manta_Style

game/scripts/vscripts/items/manta.lua
Chronophylos: I added `Normalized()` in the hopes of it actually working

game/scripts/vscripts/items/manta.lua
DoctorGester: Again IsNull is a useless function, use IsValidEntity

game/scripts/vscripts/items/manta.lua
```diff
@@ -0,0 +1,213 @@
...
+
+   if imageIndex >= blockedIndex then
+     if not ignoreBlock then
+       print('index is above blocked index')
+       imageIndex = imageIndex + 1
+     end
+   end
+
+   local distance = Vector(360, 360, 0)
+   local theta = (360 / (imageCount - 1)) * imageIndex
```
Chronophylos: I'm trying to distribute all illusions (and the caster) around the previous position of the caster.
This gives me the angle by which each position differs from the previous
for example I have 4 units so i get `360 / 4` and i multiply that with `imageIndex` to get the `imageIndex`th position on my circle

game/scripts/vscripts/items/manta.lua
DoctorGester: The hell is that code even

```
local origin = whatever
local distance = 360
local theta = (math.pi * 2 / (imageCount - 1)) * imageIndex
return origin + Vector(math.cos(theta), math.sin(theta)) * distance
```

game/scripts/vscripts/items/manta.lua
Trildar: Maybe not that well worded. What it means by replace is just that the old ones die and new ones spawn next to you. They don't spawn in place of the old ones.

game/scripts/vscripts/items/manta.lua
Trildar: Why `imageCount - 1` though? I also don't really see how this then relates to the angle positioning, but maybe I'm just failing at trigonometry.

Chronophylos: I don't know either
the angleing works
the distance is weird tho

game/scripts/vscripts/items/manta.lua
Chronophylos: copied from https://en.wikipedia.org/wiki/Rotation_%28mathematics%29
but I'll try your code

DoctorGester: Why rotate a vector if you can create a rotated one already? That's polar coordinates 101.

Chronophylos: Because I've never heard of the polar coordinate system

Chronophylos:
It should work now.
Thanks for your help

game/scripts/vscripts/items/manta.lua
```diff
@@ -0,0 +1,204 @@
...
+
+      image:MakeIllusion()  --Without MakeIllusion(), the unit counts as a hero, e.g. if it dies to neutrals it says killed by neutrals, it respawns, etc.  Without it, IsIllusion() returns false and IsRealHero() returns true.
+
+      ability.images[imageIndex] = image
+    end
+
+    -- Allow Render of Caster
+    caster:RemoveNoDraw()
+
+    ParticleManager:DestroyParticle(self.particle, false)
```
Trildar: Should also release index: `ParticleManager:ReleaseParticleIndex(self.particle)`

chrisinajar:
level 1 manta should be renamed, thoroughly test each level. i expect to see problems with this...

Chronophylos:
Yep there are problems
Level 1 doesn't delete the effect correctly

Chronophylos:
Effects get deleted correctly

Chronophylos:
now positioning is bugged for level 4

PR #1183, opened 2017-06-18
Move Simple Boss 2 to T2

Chronophylos:
on Baumis Request


chrisinajar:
pls use .editorconfig

Chronophylos:
I did that in the Browser

PR #1189, opened 2017-06-18
Add cooldowns to Time Lock and Bash of the Deep

SquawkyArctangent:


imaGecko:
How would this interact with the pseudo random distribution?
would it count the hits during the cd as not bashing, increasing the chance?

Haganeko:
Shouldn't do that.

SquawkyArctangent:
It only increments the chance to proc each time it had the potential to proc, but didn't.

Chronophylos:
Travis fails; You do not have new IDs for your abilities

SquawkyArctangent:
Is there a reason why it demands IDs for this stuff? They're not exactly used for anything besides Valve's stat tracking ( which we don't have access to and probably isn't even fully enabled for custom game ) or items.

Trildar:
I actually feel like I've read somewhere that the PRD on Entangling Claws counts hits during CD as well, but I have no idea if that's true or not.

Haganeko:
A lot of people seem to want to know about PRD and CD... Here's what I could find:
http://dev.dota2.com/showthread.php?t=88281
Also, "Note that for instances that would not trigger the effect, the probability counter does not increase. So a hero with critical strike attacking buildings does not increase its chance to critical strike on its next attack, since critical strike does not work against buildings. The chance is not reset/increased as long as the ability is on cooldown as well (e.g. Bash)." from http://dota2.gamepedia.com/Random_distribution#Example
Also http://dev.dota2.com/archive/index.php/t-64931.html (look at last answer)

chrisinajar:
Great work!

PR #1191, opened 2017-06-19
Make Team win if opposing team has left

Chronophylos:
Triggers on player disconnect
Checks 5 mins later if opposing team player count is still 0

It's not a very good solution, but someone kept asking for this

### DIFF
--- a/game/scripts/vscripts/events.lua
+++ b/game/scripts/vscripts/events.lua
@@ -11,10 +11,26 @@ function GameMode:OnDisconnect(keys)
   DebugPrintTable(keys)
 
   local name = keys.name
-  local networkid = keys.networkid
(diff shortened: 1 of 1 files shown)


Haganeko:
Can you check if the opposing team is empty at start of game and disable this in that case? Would help testers not get problems.

Chronophylos:
It only executes on a the player disconnect event
a player can't disconnect when he doesn't exist

Haganeko:
So, if, say, two people are testing something (say, shield boss) on the same team, and one has to leave, the other one will need to rehost. I guess it'll be rare enough not to matter.

Chronophylos:
> It's not a very good solution

PR #1196, opened 2017-06-21
Martyr's Mail fixes

Trildar:
The return damage sound was playing for everyone when it should only play for the player getting return damage. Martyr's Mail was also reflecting damage after reductions, which made it incredibly weak at reflecting damage.

Question: Should Martyr's Mail heal for damage before or after reduction?


Haganeko:
@chrisinajar this has been discovered to need further work; I don't know whether we'd like to use the same PR or a new one. 
@Trildar thoughts?

chrisinajar:
Well, i merged it, so new PR :P

Haganeko:
It'll delete a lot of what was changed here, but I guess it works. The further work is removing reflection from MM...

PR #1202, opened 2017-06-23
Icons.exe

warpdragon:
Have some more icons!


chrisinajar:
In the future can you itemize the icons in the PR description so that I don't need to do it myself when I do the changelogs? Saves me a bunch of time. Thanks!

PR #1204, opened 2017-06-23
Delete complied charger.vsndevts_c file

Trildar:


Trildar:
Missed the fact that this was done already.

PR #1209, opened 2017-06-24
Fix attributes of Greater Travel Boots modifier

Trildar:


game/scripts/vscripts/items/farming/greater_travel_boots.lua
```diff
@@ -139,6 +127,18 @@ function modifier_item_greater_travel_boots:IsHidden()
...
+function modifier_item_greater_travel_boots:IsDebuff()
+  return false
+end
+
+function modifier_item_greater_travel_boots:IsPurgable()
+  return false
+end
+
+function modifier_item_greater_travel_boots:GetAttributes()
+  return MODIFIER_ATTRIBUTE_MULTIPE
```
Haganeko: MULTIPE instead of MULTIPLE.

chrisinajar:
Why should travel boots stack, and why aren't they hidden anymore?

game/scripts/vscripts/items/farming/greater_travel_boots.lua
chrisinajar: `    game/scripts/vscripts/items/farming/greater_travel_boots.lua:139:10: (W113) accessing undefined variable MODIFIER_ATTRIBUTE_MULTIPE`
<3 travis

Trildar:
Doesn't actually need to stack I suppose, but non-stacking boots movespeed is enforced by the modifier property already anyway, and passive item modifiers ought to stack in general. Probably less chance for it interacting problematically with the intrinsic multiplexer that way too.

PR #1217, opened 2017-06-25
Cheap armlet

### LINKED ISSUE #1215: Wrong Pricing on Armlet
It has been in the game for a while, but all the upgrades of armlet after the path of stoneskin cost 3500

Haganeko:
Fixes #1215.


chrisinajar:
I love the item cost test cases...

Haganeko:
I actually did the update prior to opening the PR. test case averted!

chrisinajar:
yeah but i don't need to double check your math, since the test cases passed :+1:

Haganeko:
True.

Issue #1221, opened 2017-06-25
Manta broken


prsnnami:
Dont know the exact reason but it is really broken, has more illusions(3 at lvl 1), gives way more damage , idk , but get manta and gg, dont know why because i havent used it.

Haganeko:
This isn't very clear. If you could look at the bug report submission docs over at the bug report repository (https://github.com/OpenAngelArena/Bug-Reports), it would probably help.

Trildar:
Sort of known. Though I hadn't made an issue for it. #BlameChrono.
Not sure what's up with the damage though. I assume you mean the illusion damage is more than it should be?

prsnnami:
in a game right now, there were 10 units , 2 heros 8 illusions , i am earth shaker , lvl 3 dunk , illusions took less damage than the heros

prsnnami:
i could test specific things if you want , playing against a lvl 1 manta illusions should not be that hard

prsnnami:
and they tell me the manta cant be upgraded

prsnnami:
and deal too much damage, acc to them

Haganeko:
Dunk doesn't count illusions since shaker aghs was changed. Could be that.
Will try to look at Manta KVs.

Trildar:
The extra illusion is a code issue, just FYI so you don't have to look in KV for that. Upgrading is probably the only thing purely KV related.

Haganeko:
Was looking at the KV for both problems. I could find no obvious issues, though.

prsnnami:
tried it myself , doesnot look like they take more damage , its probably not 350%, and also 4 illusions are too op for early game

Trildar:
Fixed by #1222.

PR #1222, opened 2017-06-25
Fix Manta Style

Trildar:
So many changes... commit logs pretty much describe them all. Most notably fixes the extra illusion and lack of outgoing damage reduction or incoming damage amplification mentioned in #1221.

Creating illusions of most heroes this way seems to cause lag when they spawn. Alchemist was particularly laggy for me, while Jakiro and Io both had no noticeable lag at all. Not sure this is fixable unfortunately.


Haganeko:
Wow. Trildar really is the anti Chrono...

Trildar:
To be fair, most of it is a bunch of small stuff, just broken up into a lot of commits so that the messages can be meaningful.

Chronophylos:
The reason for my stupid calculation it that I wanted to allow a theoretical infinite number of illusions

Haganeko:
Wait; will this break if we decide to go up to five images for some reason?

Trildar:
It won't break, but the illusions won't have a nice formation anymore. I don't really mind too much if we place them evenly in a circle, but making that circle 256 radius is kind of silly. It's far too spread out. Practically becomes like Doppelganger rather than Manta.

PR #1231, opened 2017-06-26
Rewrite Glaives of Wisdom and Intelligence Steal in Lua

Trildar:
Now respects the damage type in KV.

This implementation kind of relies on some assumptions about the numbers in the event tables. Unfortunately, I have no real way of knowing if those assumptions will hold true aside from the fact that stuff seemed to work properly throughout my time working on this. If stuff breaks horribly, changing the ability to a toggle might be one way of making this easier.


Haganeko:
Could this help stop silencer from farming Uriel illusions? Or is that already fixed?

Trildar:
Fairly sure it wasn't. We could in theory do something through this, but I don't think there are any good flags to check for, other than the Uriel heroes being on neutral team I suppose. But really, they just need to not be real heroes. It messes with too many things.

chrisinajar:
I think the other abilities have `_oaa` at the end instead of `oaa_` at the beginning, right? Can you double check how the others are (tiny grow, etc) and match the naming style of those? Either way make the filename and ability name match.

Trildar:
It's a little inconsistent I believe. The other abilities I did before have `oaa` at the beginning, but everything else has it at the end. The ones I rewrote before weren't even put in separate files. They either overwrite the vanilla ability or just replaced it in the file.

Trildar:
Changed to suffix naming convention.

chrisinajar:
Cool, we can standardize the stragglers on that. I like it.

PR #1235, opened 2017-06-27
Make a bunch of ability and item sounds respect visibility

Trildar:
I discovered while working on Glaives of Wisdom that `EmitSoundOn` doesn't seem to respect visibility. Hence why Bottle drinking sounds could be heard through Fog of War. In the first place, the Bottle drinking sound should only play for the player using the Bottle, so that was changed to do that. The others were just changed to use `entity:EmitSound` instead, which seems to respect visibility for when the sound can be heard.

List of items and abilities modified by this:
- Bristleback
- Call of the Wild (both boar and hawk)
- Demonic Conversion
- Rearm
- Charge BKB
- Hand of Midas
- Martyr's Mail
- Satanic Core
- Shiva's Cuirass (though that code isn't currently used anymore)
- Shroud of Shadows
- Stoneskin Armor
- Greater Arcane Boots
- Greater Guardain Greaves
- Greater Boots of Travel
- Enrage Crystal
- Invisibility Crystal
- Bubble Orb
- Teleportation Shard
- Infinity Bottle

### DIFF
--- a/game/scripts/vscripts/abilities/oaa_bristleback.lua
+++ b/game/scripts/vscripts/abilities/oaa_bristleback.lua
@@ -64,7 +64,7 @@ function bristleback_takedamage(params)
         -- This is the actual "damage reduction".
         params.unit:SetHealth((params.Damage * back_reduction_percentage) + params.unit:GetHealth())
         -- Play the sound on Bristleback.
-        EmitSoundOn(params.sound, params.unit)
+        params.unit:EmitSound(params.sound)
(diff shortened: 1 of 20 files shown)


Chronophylos:
I think entity:EmitSound is server onlz

Trildar:
The function exists only on serverside, but all of these instances of `EmitSoundOn` were being run only on serverside anyway. Unless I made a mistake, which I might have as I didn't pay that much attention going through these.

PR #1236, opened 2017-06-27
Fix errors when attacking items with Glaives of Wisdom on autocast

Trildar:


game/scripts/vscripts/abilities/oaa_glaives_of_wisdom.lua
```diff
@@ -57,7 +57,7 @@ end
 function modifier_oaa_glaives_of_wisdom:OnAttackStart(keys)
   local parent = self:GetParent()
   local ability = self:GetAbility()
-  if keys.attacker == parent and (keys.gain == 0 or ability:GetAutoCastState()) and ability:IsOwnersManaEnough() and (not keys.target:IsMagicImmune() or parent:HasScepter()) then
+  if keys.attacker == parent and (keys.gain == 0 or ability:GetAutoCastState()) and ability:IsOwnersManaEnough() and keys.target.IsMagicImmune and (not keys.target:IsMagicImmune() or parent:HasScepter()) then
```
Haganeko: Won't this only return true if the target is magic immune and silencer has scepter?

Trildar: `keys.target.IsMagicImmune` and `keys.target:IsMagicImmune()` are different things. The former is the function handle, if it exists, if `keys.target` doesn't have a `IsMagicImmnue` method then it's `nil`, which is considered `false` for logic. The latter calls the method and returns true if `keys.target` is Spell Immune.

Haganeko: Oh! Thank you.

PR #1237, opened 2017-06-27
Switch Manta back to using synchronous CreateUnitByName

Trildar:
Also switch from caster:GetOwner() to caster:GetPlayerOwner(). I'm just being rather undecisive on that.
The asynchronous CreateUnit didn't really accomplish anything, and based on the fact that Manta caused a crash while I was playing around with it, it might actually be less stable.


Chronophylos:
You did not change any line where caster gets set

Trildar:
Is there some reason I have to?

PR #1238, opened 2017-06-27
Allied Cyclone fixes

Trildar:
- Add cast sound
- Add purge when casting on enemies
- Tweak tooltip to make it clear that it can target Spell Immune allies

Unsure if that last one is actually the behaviour we want, but it is how it works currently, so might as well make the tooltip accurate.


VilkkuV:
Thanks @Trildar

Issue #1186, opened 2017-06-18
Hero tweaks


Haganeko:
- [x] Make bosses immune to Reaper's Scythe.
- [x] Give bash abilities a cooldown. (#1189)

Maroxis:
Don't forget about killing bosses with axe's dunk

Chronophylos:
His dunk is not strong enough. It's not percentage based.

Maroxis:
but a lot of bosses have little health and a lot of armor, so you can 1 shoot 1/3 of their health

Haganeko:
Scythe kills bosses at 70% health if you have it at max level. And that's because we nerfed it.

Trildar:
We actually already had #607 for giving CD to bashes. Was never done though.

Chronophylos:
Culling Blade threshold:
`250 325 400 550 1150`
Most T1 bosses have less that 2K HP
Later bosses all have more HP
And when you have the `1150` threshhold you'll most likely encounter bosses with >10K HP

Haganeko:
Finishing a boss also puts Culling Blade on CD, which is not ideal for Axe.

l34Um1:
I disagree with The Swarm. Swarm is a nice alternative way of tanking bosses, I don't think we should take it away. Scythe is kinda broken though.

Haganeko:
Weaver is doing very very well, though... Wanted to nerf him somehow. If we're not doing this, then what?

l34Um1:
Implement more detection. Reveal on scan is still missing.

allanvbhg:
What about making gem edible, cause no one wants to carry a gem now because an item is way more important?

Chronophylos:
No
we talked way too often about this
the final answer was always no

l34Um1:
Implement Dust effect on upgraded Mekka. Make Scan reveal. Give Urn offensive active detection (and nerf its damage). Some ideas.

Haganeko:
I was thinking we could put dust on an offensive area item, instead of a defensive one. A Shiva upgrade, or urn of Sorcery if that's still a thing

allanvbhg:
How does that work? Detection on urn? Add a dust effect?

l34Um1:
I would like to have the detection on support items. Supports still need a bit of love (although they are super underrated atm).

allanvbhg:
Yes but how does urn reveal invis?

Minnakht:
[...]

PR #1246, opened 2017-07-01
Make bosses immune to Reaper's Scythe

VilkkuV:
#1186 This is the easiest way to do it, but it prevents targeting [other creep heroes](http://dota2.gamepedia.com/Summons#Creep-heroes). Is that OK?

EDIT: Now it doesn't affect other creep heroes.


Trildar:
I don't really think we actually need this anymore, since Valve patched Reaper's Scythe to be properly affected by damage block, which means bosses take 85% less damage from Reaper's Scythe. I mean, it's still pretty powerful at higher levels, but level 1 can barely do any damage against bosses now. Need confirmation from someone else though.

VilkkuV:
Okay, thanks for info.

Haganeko:
@Trildar @VilkkuV we actually do. Veil + eblade makes it possible to kill bosses from 40% HP late.

PR #1248, opened 2017-07-02
Cave overhaul

Haganeko:
Last attempt to make cave worth it while new comeback mechanics aren't implemented.
Marked WIP to await review.


chrisinajar:
LETS TRY IT

PR #1252, opened 2017-07-07
Fixed the Map

warpdragon:
Prettied things up.


warpdragon:
Radiant

warpdragon:
Dire

warpdragon:
I had to move the boss spawners (some further, some closer) so that they would be in the center of its area. I also kept the approximate position of the terrain baumi created as closely as possible. Enjoy!

warpdragon:
This tree

PR #1255, opened 2017-07-08
KV Fixes

imaGecko:
Abilities should now work again
Also cleaned up duplicate values where I came across them
Clean up the custom hero file, had entries for files that no longer exist


imaGecko:
I believe travis is giving out false errors this time

Trildar:
Probably can't parse LinkedSpecialBonusField since that's a new key name.

chrisinajar:
test cases need to white list `LinkedSpecialBonusField` as a valid value

PR #1262, opened 2017-07-10
Crimson 5 levels

hallelsegel:
finally finished this


Haganeko:
What on earth is that second commit?

PR #1267, opened 2017-07-11
Fixed the degeneracy with lua bashes and illusions

SquawkyArctangent:
one line fixes woo


chrisinajar:
This seems important, hahaha

PR #1268, opened 2017-07-11
Added a level 4 to Elder Dragon Form

SquawkyArctangent:
Based on Baumi's idea, it allows EDF to be triggered on attack with
heavily reduced duration.

Also fixes a KV mistake which accidentally gave splash attack too much
range.

and for some reason the bash thing got dragged along bleh


game/scripts/vscripts/abilities/oaa_elder_dragon_form.lua
```diff
@@ -131,10 +192,58 @@ if IsServer() then
...
+
+--------------------------------------------------------------------------------
+
 	function modifier_dragon_knight_elder_dragon_form_oaa:OnRespawn( event )
 		local parent = self:GetParent()
+		local spell = self:GetAbility()
 
-		if event.unit == parent then
+		if event.unit == parent and spell:GetLevel() >= 5 then
 			local spell = self:GetAbility()
```
chrisinajar: Since you added this line above, you need to remove this one. The test cases prevent you from defining the same variable twice.

PR #1269, opened 2017-07-11
Hotfix for the cave

Haganeko:
Should reimplement exponential experience as a separate thing cave won't hook into using clears, and reapply it to cave using minutes.


chrisinajar:
Proper fix in #1272

Haganeko:
Thanks.

PR #1272, opened 2017-07-12
Remove exponential xp gain from farming cave

chrisinajar:


chrisinajar:
This effectively does the same as just commenting out the exponential parts. Only creep camps use that multFactor now, it's always 1 for the cave.

The cave has never taken current time into account, we could add that as a separate unit of work if it's desired.

That said, I don't think that or the other changes are the problem with the farming cave. We shall see.

Haganeko:
The problem was the cave "going nuts" after the fifth clear. 
Given that the farming cave considers twelve minutes per clear, that would be precisely the onset of exponential XP.

chrisinajar:
Yeah you're right, I was reading the code thinking the 4 was multiplied by
the k value, but it's actually 12, so yeah 5th clear.

PR #1301, opened 2017-07-21
Fixed a bug introduced like three months ago

Haganeko:
With one character. Amazing, right? feat. chronophylos, chrisinajar.


warpdragon:
what the...

Haganeko:
The code was fetching the XP value instead of the gold one. The bug has been there since the cave was added, but since I was the only one who meddle into the numbers I was the only one who noticed.

PR #1307, opened 2017-07-23
Fixed Bug where players get stuck if they finish cave right before Duel

Anemolo:
* Checks if player is in duel right before teleportation. Instead of checking before timer starts.

* QuickTeleportAll function(line 366) is not being used inside this file anymore.

* Fixed the Bug where i get triggered when Baumi says that this is a "very significant bug, really really really difficult to deal with".


Chronophylos:
>  Fixed the Bug where i get triggered when Baumi says that this is a "very significant bug, really really really difficult to deal with"

hahaha

About the whitespaces: Your editor probably doesn't respect [EditorConfig](http://editorconfig.org/), we're using it.

Honeth:
did you play test this?

Anemolo:
I did play tested, works as intended.

chrisinajar:
Yup, did the same fix myself, although you can also remove the QuickTeleport function since it's not used

PR #1310, opened 2017-07-25
Balance

l34Um1:
rip haganeko when he sees the bosses


Haganeko:
I'm actually okay with the magic resistance thing, though I'd probably have stuck to a more conservative 0.

l34Um1:
I spent a long time thinking about it and I think -25% is better but I wouldn't be surprised if I am wrong. Luckily it's an easy fix if it needs to be changed again.

Haganeko:
True. Getting a bunch more bosses where just right clicking away doesn't work should help!

SquawkyArctangent:
Don't forget to change AbilityCastRange on the Underlord ult - there's a reason why the keyvalue has the "_tooltip" part at the end

PR #1317, opened 2017-07-28
Update Highway to Hell

SquawkyArctangent:
Precaches the portal particle so it properly shows up, as well as adds a
"Close Highway" subspell to prematurely close the portals.

As a result, there's now a status buff on Underlord that shows him the
duration of all his portals, and units are interrupted ( Nether Swap
style ) when they go through a portal.


chrisinajar:
if you use an editor that supports `.editorconfig` files you never need to worry about travis formatting. You'll also be happier. I recommend sublime text 3.

SquawkyArctangent:
it's the principle of the matter

Issue #1321, opened 2017-07-31
Lone Druid's bear doesn't teleport to duel arena.


Ynscription:
Lone Druid's bear does **not** get teleported to the duel arena like other summons.

PR #1322, opened 2017-07-31
Fix LD's Bear not getting tped into duels

Chronophylos:
Include Heroes in search parameter for duel tp
Fixing #1321


Minnakht:
I wonder how will this interact with Brewmaster. Duel starting during Primal Split caused some janky behavior that ultimately ended up with the elementals being in the duel - wonder if this'd make it smoother.

Doesn't really matter, I suppose.

Issue #1229, opened 2017-06-25
Disassembling Farming Boots


Trildar:
Unsure what current consensus might be, but right now Greater Arcanes can be disassembled as an artifact of copypasta, but none of the other Farming Boots can be. Do we want to make all Farming Boots disassembleable, make Greater Arcanes not disassembleable, or possibly just leave it as is?

Trildar:
Melon suggested making all level 1 Farming Boots disassembleable, and Haganeko agreed. Probably also normalise Greater Arcanes to this so that only level 1 can be disassembled.

PR #1327, opened 2017-08-11
Disassemble lvl1 Farming Boots and do not greater one

ProstoSanja:
Hello, hopefully someone remembers me. 
Decided to return back to contributing, and started with easier tasks to remember how everything went.
This is the solution to issue #1229


Chronophylos:
Welcome back!

Issue #610, opened 2017-04-20
Limit "room x cleared"-text to only people in the farming cave


l34Um1:
Title.

Haganeko:
Should this display at all? It sounds like a debug thing.

Trildar:
Maybe when we have the visible doors we can consider taking it out, but right now it's good to inform people not familiar with how the farming cave works.

warpdragon:
Limit teleporting to fountain to players in the cave or implement teleporter pad

warpdragon:
Take a look in the assets browser for a suitable model. I bet theres some cool ones in there

Chronophylos:
There are some good particles as well

warpdragon:
I know. Either way. I dont much like the duct tape used at the entrances of the individual boss arenas

warpdragon:
Has anyone done this? Seems simple enough.

PR #1333, opened 2017-08-13
Show room 1,2,3 cleared to only people in room!

ProstoSanja:
#610

### DIFF
--- a/game/scripts/vscripts/components/cave/cave.lua
+++ b/game/scripts/vscripts/components/cave/cave.lua
@@ -182,13 +182,17 @@ function CaveHandler:CreepDeath (teamID, roomID)
         FIND_ANY_ORDER, -- order
         false -- can grow cache
       )
+
+      local hasSeenNotification = {}
+
       for _, unit in pairs(result) do
-        if CaveHandler:IsInFarmingCave(teamID, unit) then
+        if CaveHandler:IsInFarmingCave(teamID, unit) and hasSeenNotification[unit:GetPlayerOwner()] == nil then
           -- inform players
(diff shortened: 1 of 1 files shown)


game/scripts/vscripts/components/cave/cave.lua
```diff
@@ -171,11 +171,26 @@ function CaveHandler:CreepDeath (teamID, roomID)
...
+        DOTA_UNIT_TARGET_TEAM_FRIENDLY, -- team filter
+        DOTA_UNIT_TARGET_ALL, -- type filter
+        DOTA_UNIT_TARGET_FLAG_NONE, -- flag filter
+        FIND_ANY_ORDER, -- order
+        false -- can grow cache
+      )
+      for _, unit in pairs(result) do
+        if CaveHandler:IsInFarmingCave(teamID, unit) then
+          -- inform players
+          Notifications:Top(unit:GetPlayerOwner(), {
```
chrisinajar: This will go off for every unit a player has in the farming cave. You should make sure to only send the notification once per player ID. simple map and check would work, `hasSeenNotification[playerID] = true`

ProstoSanja: Wait, but like before this, I get all players from friendly team, which are in this cave `local result`, and this second check is even redundant, isn't it?

chrisinajar: You're fetching all *units* from the team, not just heroes. A venomancer will get the notification for every plague ward they have in the cave.

ProstoSanja:
Done

PR #1344, opened 2017-08-22
Mana rebalance example

imaGecko:
Applied the changes from balance discussion to some example heroes.
Abaddon, Juggernaut, Lina, Skywrath, Underlord, Zeus.
Overall ability mana costs go up, ult costs go down.
Abilities that scaled their mana cost super hard and as such became expensive to the point of not being usable will be cheaper.
Abilities that didn't scale their mana cost at all will now scale in the last 2 levels.

Instead of going the typical 1/1/1/1/2/8 scaling the mana cost from the highest vanilla level now will be multiplied by 1.5 and 2.5


imaGecko:
https://images-ext-2.discordapp.net/external/49mPG54ejtpQQPsgdRH1bqKCuA2_owox2iCtpIRapco/https/puu.sh/xB7Ww/f6f80662d8.jpg?width=1202&height=677
The rebalance would put the manacost at 750 plus more expensive regular spells. Would still punish for not building any mana but at least he could cast his spells even if not a full rotation

PR #1345, opened 2017-08-22
Cleaning up Item files

imaGecko:
The mix of tabs and spaces in some item files were bugging me so I converted tabs to 2 spaces and reindented


chrisinajar:
tldr

PR #1347, opened 2017-08-23
Tranquil Boots

ICryALot:
I think giving lifesteal and lower cooldown is better.


Haganeko:
That should be debated before implementation. Also, I don't think changing the KV without editing the Lua would work.

chrisinajar:
what haga said

PR #1351, opened 2017-08-24
Make item upgrades not disassemblable

imaGecko:
Items that can no longer be disassembled:
Any tier 2+ item

Items that can now still be disassembled:
Reflex items
Tier 1 farming boots
Tier 1 items that can be disassembled in vanilla dota


chrisinajar:
tldr

PR #1355, opened 2017-08-24
Improve item searching in shop

imaGecko:
Added a bunch of aliases to the items so they can be properly searched for in the shop
Does not include farming boots


imaGecko:
Conflicts are solved now ._.

PR #1360, opened 2017-08-25
Tranquil Boots

ICryALot:
I think giving lifesteal and lower cooldown is better.


Trildar:
Duplicate of #1347. Closed for same reasons.

PR #1362, opened 2017-08-25
Duel fix

ArchOAA:
Commented out the GetAuraEntityReject. Was thinking this can be replaced with a separate duel modifier zone that applies the modifier_duel_rune_enemy aura but is a larger area so that there is more control if this fix has the reverse effect on duels.


chrisinajar:
The first 2 fixes are already in, and removing that function would make the duel rune hill not work correctly.

PR #1363, opened 2017-08-25
Cave Creeps True Sight

Yzanas:
all cave creep KVs are now seperate from their non cave counterparts (like the Tomato before)

all cave creeps now have the necro3 true sight passive


Yzanas:
finished, tested, and implementable btw

PR #1367, opened 2017-08-26
Make cores killable

imaGecko:
Sometimes you just will not have any use for cores
Dropping them and killing them with an attack gives a way to dispose of them that will not cause new players to just get rid of their cores. on accident, from misinformation or otherwise.
Less clutter, same gameplay


chrisinajar:
These were originally designed this way to help players understand how important the cores were, I think for many reasons this is much more obvious now, I approve.

Haganeko:
I also support making them sellable. It's not a major income source, but if a team can  get more than six (five in some cases if they're using reflex items) cores of a tier they should get *something* for it besides the points.

imaGecko:
Killing bosses still have the rewards of being worth a kill plus denying an easy boss kill to the enemy since it will now spawn a harder one

Issue #1381, opened 2017-08-30
Fix Fountain Pull from Cave Room 1 Exploit


warpdragon:
I'm just gonna probably check that the entities are symmetrical and that no team has the advantage over the other while also removing the current Cave pulling exploit.

warpdragon:
Update: I'm also gonna take a look at some of those weird visual bugs that happen with select circles near cliff edges. I understand the problem and its relationship to the tileset, but have yet to solve it. I'm gonna take a closer look. I'd like to find a solution that doesn't require using a different tileset.

PR #1382, opened 2017-08-30
Map tweaks

warpdragon:
Did <https://github.com/OpenAngelArena/oaa/issues/1381>. Fixed some fountain exploits and visually separated aura and non-aura zones using materials.


warpdragon:
Rotated and flipped the radiant shop and fountain so their relative distance other map assets, particularly the cave creeps, would match dires.
Also changed the material in the main area where the aura is to "stone"

PR #1388, opened 2017-09-02
Added text to tell players about sv_cheats

Chronophylos:
Thought this was a good idea especially with the upcoming tournament.


chrisinajar:
It'd be cool if it appeared statically up with the version number so that it's apparent in any clip as well

PR #1398, opened 2017-09-10
Implement capped Death Pact

SquawkyArctangent:
Current cap is based on a maximum of 1500 creep health, since that's the
highest you'd find on a creep in vanilla DOTA.

'Course, the cap is set on the health and damage gained from it, so it's
entirely possible to e.g. have Clinkz only benefit from up to 1500 creep
health for the health, but benefit from up to 3000 creep health for the
damage.


Chronophylos:
> bb18e47213a275581f7b2a53c1913bf307eca01f
LuL

PR #1400, opened 2017-09-12
Pause game during pick screen, teach bots to use captains mode

chrisinajar:


Haganeko:
How do you teach bots to use CM?

PR #1402, opened 2017-09-12
Recipe leveled Icons

imaGecko:
Adds the dots to the recipes of the different item tiers. 

A common complaint is that "items don't combine" when in reality they just bought the wrong tier recipe. Also if you have a recipe in your inventory the only way to look up the tier is looking at the sell cost.
This should make it a lot more clear


warpdragon:
image

PR #1403, opened 2017-09-13
Wrap all event handlers in try expressions

chrisinajar:


game/scripts/vscripts/components/duels/duels.lua
```diff
@@ -177,12 +177,18 @@ function Duels:CheckDuelStatus (hero)
   end
 
   if not player.assigned or not player.duelNumber then
+    DebugPrint('Player died who isnt in a duel?')
```
Chronophylos: I think we should make this an `error`.

game/scripts/vscripts/internal/eventwrapper.lua
```diff
@@ -0,0 +1,17 @@
...
+-- ListenToGameEvent('dota_rune_activated_server', Dynamic_Wrap(GameMode, 'OnRuneActivated'), self)
+function ListenToGameEvent (eventName, fn, obj)
+  return _ListenToGameEvent(eventName, function (...)
+    local data = {...}
+    local status, err = pcall(function()
+      fn(unpack(data))
+    end)
+    if err then
+      print(err)
+      error(err)
```
Chronophylos: why `print` and `error`?

chrisinajar: just in case one of the wrappers breaks, since error also throws an exception

game/scripts/vscripts/components/duels/duels.lua
chrisinajar: i'd rather have it logged that it happens then see the behavior after it than to hard fail out, error keeps the rest of the code from running

Issue #1249, opened 2017-07-02
Add Cave Comeback Factor


Haganeko:
Add a comeback factor which multiplies gold given by the cave. Suggested value is:
`exp(ln(m)t/(1+|t|))`
Where m is the maximum multiplier and t is a scaling factor times the relative difference in wealth between the teams.
`t = k*(MyTeamWealth-YourTeamWealth)/LeastTeamWealth`
Exposing the bound m and the scaling factor k somewhere would be helpful.

Haganeko:
Currently, best estimate for initial m and k are:
m = 2
k = 1/10000

Zarnotox:
Wouldn't it be better to make 1/k increase over time?

Haganeko:
Actually, since I made the t-value relative, no... But then that is now a terribly low value. Updated!

warpdragon:
afaik this has been taken care of. Closed.

Haganeko:
AFAIK it hasn't. If you say so, however... 
@chrisinajar?

warpdragon:
really? I thought u did it like 2 patches ago...

Haganeko:
I have the numbers... Turns out I can't code.
All I did was change the cave from two minutes per room to three, and include a factor which makes its increase grow faster than normal creeps'

l34Um1:
So is this or not? It says merged but I don't believe it's actually in the game.

Haganeko:
It is not in the game.

Haganeko:
No longer relevant since all of the cave will be reworked.

PR #1404, opened 2017-09-13
Post Tourney Cave

Haganeko:
Need to implement drake as a custom creep to give it the slow resistance.
Also, if someone could tell me how to hook into each team's current net worth, I'll try to implement #1249 here.


Haganeko:
So I forgot to update my repo. Duh.

PR #1425, opened 2017-10-27
Level 5 skills rework

l34Um1:
This is a massive rework of almost all heroes. Previously level 5 spells were very weak in comparison to level 6 which often lead to casters trying to rush to level 40 as fast as possible. They were useful level 1-25 and then starting at 40 again. This change evens the gap between levels 4-5 and 5-6, hopefully fixing this issue. 

Overall this buffs almost all heroes in the game but casters benefit much more.

There are also quite a few balance changes to various heroes in here. Listing everything isn't possible inside the Dota client, you will have to go to the GitHub for the detailed information.


warpdragon:
Wow thats a lot of work! Nice job. @chrisinajar will travis check for syntax?

chrisinajar:
Yup! You'll notice a commit ago it caught the LC error. It reads every ability and does some basic sanity checks.

PR #1429, opened 2017-11-02
Various technical fixes for 7.07, nearly no KV work

chrisinajar:


chrisinajar:
I don't want to fix these.

PR #1430, opened 2017-11-02
Activate the two new heroes

Honeth:
already enabled still need to give them ability upgrades


chrisinajar:
Add the upgraded levels in this PR so we can merge em in all at once.

PR #1441, opened 2017-11-03
[WIP] 707 Abilities

imaGecko:

PR #1442, opened 2017-11-03
Add more test cases around special values

chrisinajar:
Checks that any ability or item that is based solely on a dota one (no custom stuff, just numbers) has all of the special values listed in it's parent, as well as any special values in them for things like talent links.


chrisinajar:
Will be merged with https://github.com/OpenAngelArena/oaa/pull/1441

PR #1450, opened 2017-11-04
Fix Greed giving GPM on illusions

Trildar:


game/scripts/vscripts/items/farming/modifier_passive_gpm.lua
```diff
@@ -19,14 +19,19 @@ function modifier_passive_gpm:GetTexture()
...
     if not PlayerResource then
       -- sometimes for no reason the player resource isn't there, usually only at the start of games in tools mode
       return
     end
     local caster = self:GetCaster()
     local gpm = self:GetAbility():GetSpecialValueFor('bonus_gold_per_minute')
-    Gold:ModifyGold(caster:GetPlayerID(), gpm / 60, true, DOTA_ModifyGold_GameTick)
+
+    -- Don't give gold on illusions or Meepo clones
+    if caster:IsIllusion() or caster:IsClone() then
```
chrisinajar: Should you check if they're a tempest double too?

Trildar: It's been a while. Forgot that Tempest Double requires a separate check. Good catch.

PR #1453, opened 2017-11-04
small fixes

imaGecko:
Disabled custom abilities that aren't updated yet (fv and slardar bash)
fixed herolist
od arcane orb kv fix, changing the damage type is tooltip only, it deals pure damage
gave all items shared cooldowns
cyclone still had 100+ mana regen
made refresher core and sanctity euls have constant mana regen


imaGecko:
should probably get a lookover

PR #1454, opened 2017-11-05
Fix Dagon KVs

Trildar:
7.07 removed the bonus_damage AbilitySpecial, which broke our Dagon due to non-matching AbilitySpecial.

More test case tweaks?


chrisinajar:
Could be dota changing small things in the background, they have a few errors we noticed (like necro heartstopper aura shows a different radius than it has because they didn't update the ability range, only the kv range for the aura)

chrisinajar:
Hmm, I wonder why the test cases didn't catch this item as bugged before. I'll check into that.

PR #1459, opened 2017-11-06
Fix for trees blocking vision (mergeable edition)

Satsaa:
5aa46c70569fed55c99bc462304a02d0


chrisinajar:
Woot!

Before you start your next patch, run this
```
git fetch upstream
git reset --hard upstream master
```
that'll completely wipe your local history and reset you back to OAA master. Then you can do work from there.

If you don't want to need to delete your old work every time, you can do things in a branch. That also allows you to have more than one pull request at a time!

PR #1468, opened 2017-11-07
Fix Covfefe interaction with health regen amp from Strength

Trildar:
Turns out the health regen % increase from Strength also applies to negative regen. This combined with changes in health regen during Covfefe's duration (or possibly even without) caused unstable oscillation in the regen amount that Covfefe applies. And turns out Dota doesn't like things applying insanely large numbers of negative health regen.

So this rewrites Covfefe to not use a health regen modifier property. Probably causes health regen to not be smooth in cases, but until Valve actually makes Spirit Vessel's effect accessible from Lua, this is probably the best that can be done.

Side note: This effect seems to stack multiplicatively with Spirit Vessel now. Though that only matters for lvl 1 Covfefe anyway.


chrisinajar:
lol, old covfefe was a piece of shit xD xD xD

Issue #1443, opened 2017-11-03
Implement an updated logging (Spec layout and discussion)


IBurn36360:
This is a list of general thoughts and a base spec layout for a logging API both internal to the game/server client side and the server side where the API sends the event payloads to.

# General Thoughts

## Game Side
- Internal API should be flexible and allow payloads to be generated for the following cases:
    - Informational payloads, such as general number or status based events not necessarily associated with any other event or error
    - Status payloads, such as a start of game event, duel start event etc.
    - Sync payloads, such as sync errors or anything specific with the network handling and updating before the main game loop takes control of the data (These are very often, so we would quantify what counts as a sync error and likely throttle them)
    - Exception payloads, such as caught exceptions that may or may not contribute to game breaks or errors.  In this case, exception should be reserved for any error that is caught and handled.
    - Error payloads, such as uncaught exceptions that may or may not contribute to game breaks or errors.  The alternate case of the above
    - Fatal payloads, sent on game hard crash, if possible.  In the event a game sends no events for a reasonable amount of time to the API, the server should consider the game as fatal for one reason or another.  A host disconnect or lobby disconnect should be flagged as such if possible.
- The API CAN NOT collide with game performance, meaning all API interaction and payload sending SHOULD be as light as possible and SHOULD attempt to allow the game to proceed before its own processing.  Any API that affects the gameplay in a detrimental manner will drive users away and should be avoided at all costs.
- The API should be available in all scopes as a global or superglobal and should present a simple and unified interface for developers to use.
- The API should collect backtraces for any and all errors, when available, on all event of severity Exception and higher (Sync might be a candidate for this as well).  Events may request a trace on behalf of the payload for events who do not already get backtraces.
- The API format MUST allow for dynamic allocation of data, that the view on the server side can decide on how to display later.  This is to prevent an event from being sendable due to design limitations and allow for deep level data collection where it makes sense.

## Server Side (Non-DOTA)
- Ingest point for log payloads should be resolvable for all people (Ideally should be a domain name pointed to a server or set of servers)
- Endpoint should validate requests are either a chain creation or are part of an existing chain (Prevent malicious or orphaned events from going into the system)
- Endpoint should be as fast as possible to release the connection so a new one can be made by the HTTP daemon to prevent load issues when events like a tournament are going on.
- Views should understand how to bucket and display the log events in a dashboard for easy at-a-glance vieweing
- Views may or may not require authentication, so allow all views to have permission controls
[...]

PR #1486, opened 2017-11-13
Implement a new logging system to allow OAA games to report to an external server

IBurn36360:
This PR aims to address Issue #1443 

This is being opened up to allow for discussion on the manner in which this code is being developed for the OAA game side of things in case anything needs to change.

- Initial code shows:
    - inclusion placement (Very early in game init to allow the real game start to be captured as an event (This should allow games that hard crash or have full bails to be identified)).
    - One example call (Would capture and dispatch the game start event)
    - Full module for the internal to external reporting

### DIFF
--- a/game/scripts/vscripts/addon_game_mode.lua
+++ b/game/scripts/vscripts/addon_game_mode.lua
@@ -1,6 +1,10 @@
 -- This is the entry-point to your game mode and should be used primarily to precache models/particles/sounds/etc
 
 GAME_VERSION = "2.21.0"
+
+-- Setup the main logger
+require('internal/logging')
+
 CustomNetTables:SetTableValue("info", "version", { value = GAME_VERSION })
 -- lets do this here too
 local mode = ""
--- a/game/scripts/vscripts/internal/logging.lua
+++ b/game/scripts/vscripts/internal/logging.lua
@@ -1,0 +1,162 @@
+-- Though we shouldn't encounter a case where this gets initialized twice, don't allow it to happen anyway
+if (D2CustomLogging == nil) then
+    D2CustomLogging = {
+        -- TODO: Make this validate based on what kind of game is being launched if there is an care on not logging dev games
+        --[[
+            @var bool isEnabled - State true if logging is to be enabled
+        ]]--
+        isEnabled = true,
+
+        --[[
+            @var string gameClientVersion - The string game client version detected when the game started up
+        ]]--
+        gameClientVersion = CLIENT_VERSION,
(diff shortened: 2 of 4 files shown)


chrisinajar:
we have print hooks in vconsole.lua, those might be useful

chrisinajar:
You can find example usage of the HTTP request object in the stats collection library in our codebase

IBurn36360:
This is at a stage where a Loggly account needs to be set up and bound (Takes a matter of minutes to do).  I created one quickly just to test, and can likely delegate access out if no one wants to setup an account.

chrisinajar:
just make the account tokens configurable. if they're private then make it a dedicated config file, otherwise just put them into `settings.lua`

IBurn36360:
The instance has an endpoint configuration internally at this time (I can move the actual references to come from the `settings.lua` file without an issue).

https://github.com/OpenAngelArena/oaa/pull/1486/commits/136d4be112b3bc68e4e64d2f721f33aee342e3d6#diff-027e60dec048b1b02390d99f12eba77cR29
https://github.com/OpenAngelArena/oaa/pull/1486/commits/136d4be112b3bc68e4e64d2f721f33aee342e3d6#diff-027e60dec048b1b02390d99f12eba77cR33

If that is all, then we might be good for a run on the test client.  I have, since the screencap shown in #general, added a players array with status and account name to help with debugging state at that point in time.

Since there shouldn't, ideally, be any errors in the normal release code, we might need to put in a testing error message just to see it working outside of the tools client.  I just added a test error call to the duel start function to test when in tools mode.

chrisinajar:
yeah it'd be great if it was boiled down to a `LOGGLY_ACCOUNT_ID` value in `settings.lua`

IBurn36360:
Done and rebased over master so the merge conflict is handled

IBurn36360:
Rebased to resolve version merge conflict

chrisinajar:
Yay! Super excited for this.

IBurn36360:
Just before you go to merge this, I'll resolve the merge conflict (Is a side effect of the version constant being updated, so happens every version update).

IBurn36360:
Since you could be merging this over the weekend, it is rebased over master to resolve the merge conflict (Also adds in a bail out case for if the Loggly account UID is not provided, allowing it to be merged before complete configuration without throwing errors when something goes to be sent).

chrisinajar:
Great work, thanks for adding the error checking code so I can merge this. Makes things easier :+1:

PR #1491, opened 2017-11-26
Terrorblade boss fix

Nortlen:
Changed Terrorblade boss to be an ancient, so wraith king cant one shot him.


chrisinajar:
Thanks! It's amazing how long stuff like this can sit in the repo without anyone fixing it....

PR #1494, opened 2017-11-28
Weird lighting in hero select

Zarnotox:
This will fix the weird lighting that is in the heroselection screen right now.


chrisinajar:
Already done in other PR

Issue #1455, opened 2017-11-05
Dagon does not interact with illusions the same as vanilla


Trildar:
Dagon was rewritten in Lua so that it could be given spell amp at the levels above vanilla, but it wasn't written to interact with illusions the same way as vanilla Dagon. Vanilla interaction is apparently to instantly kill illusions, except for if they have Shallow Grave, in which case it just brings the illusion to 1 HP.

l34Um1:
This was recently fixed.

PR #1495, opened 2017-11-28
Make dagon deal 13371337 pure damage to illusions

Zarnotox:
As issued in #1455 
Shallow grave will stop this damage

> Instantly kills illusions, regardless of any status effects, except for Shallow Grave.
> Dota2 wiki

`This commit also contains the lighting bug fix since I forgot to create a new branch for that one.`


Haganeko:
Wouldn't it be cleaner to just kill the illusion?

chrisinajar:
i assume shallow grave wouldn't work?

Haganeko:
Oh. True.

PR #1499, opened 2017-11-29
Small Map Update. READY

warpdragon:
RIP Pig-Goyle


game/scripts/vscripts/abilities/fountain_attack.lua
```diff
@@ -58,12 +58,20 @@ modifier_fountain_attack_aura = class(ModifierBaseClass)
 
 function modifier_fountain_attack_aura:OnCreated(keys)
   local caster = self:GetCaster()
+  local teamID caster:GetTeamNumber()
```
Trildar: Missing an `=` there. Should be `local teamID = caster:GetTeamNumber()`. Classic Chrono. :P

Chronophylos: dota didn't threw any errors

Trildar: I dunno. I just tried putting a line like that in my editor with the linter enabled. Apparently it's not a syntax error, cause Lua. And Dota apparently doesn't have any issues with you calling CreateParticle with a `nil` name. You just don't get a particle.

game/scripts/vscripts/abilities/fountain_attack.lua
```diff
@@ -58,12 +58,20 @@ modifier_fountain_attack_aura = class(ModifierBaseClass)
...
   local target = self:GetParent()
-  local attackEffect = "particles/abilities/tesla_coil_radiant.vpcf"
+  if teamID == DOTA_TEAM_GOODGUYS then
+    local attackEffect = "particles/abilities/tesla_coil_radiant.vpcf"
```
Trildar: Um. Also just realised that this is scoped to the `if` block. The variable needs to be declared outside.

warpdragon:
Ok so this thing is basically ready. A modification I made to the fountain resulted in a change to the fountain regen range and that's the only thing left to fix.
We need to either:
1. Create a new fountain. We could have it only apply the regen aura in a rectanglular trigger box so that we control exactly where the regen does and doesnt' apply and we can make sure that range is the exact same range as the laser and the shop.
2. Simply figure out a way to move the fountain to its original position without having to also move the laser.

I think Chrono has tried #1 and failed. Dota is being an asshole. Does someone else wanna take a look?

warpdragon:
PUNCH IT

chrisinajar:
remove your WIP tags noob :P

warpdragon:
Im on my fone

warpdragon:
There. Happy?

game/scripts/vscripts/abilities/fountain_attack.lua
```diff
@@ -32,7 +32,9 @@ function modifier_fountain_attack:GetModifierAura()
 end
 
 function modifier_fountain_attack:GetAuraRadius()
-  return self.trigger:GetBoundingMaxs():Length2D()
+  local caster = self:GetCaster()
+  local distance = caster:GetAbsOrigin() + self.trigger:GetAbsOrigin()
+  return self.trigger:GetBoundingMaxs():Length2D() + distance
```
chrisinajar: This function doesn't make any sense.

It takes caster location + trigger location, which is some spot very far away (`-` would give the distance as a vector, + is the two combined)
then it takes the bounding max's, which are the largest extremities of the given zone, calculates the 2d length of that vector, and adds that float value to the vector position...

warpdragon: @Chronophylos i dunno. it works though.

chrisinajar: What are you trying to do here?

warpdragon: that's the bounding box for target selection for the fountain lazor

warpdragon: something i changed broke it. this is what fixed it.

chrisinajar: if i had to guess, the trigger box is no longer at the exact same spot as the entity. `return self.trigger:GetBoundingMaxs():Length2D()` seems correct to me.

warpdragon: ok fine i'll see what happens i guess....

warpdragon:
i suspect it will work. some kinda bug got into our files causing symlinks to break. that might have made it look like things weren't working when they actually should have.

yup. it worked

PR #1503, opened 2017-11-30
Add tests for abilities and talents

chrisinajar:


chrisinajar:
There were 2 talents that needed replacing, i based both on existing talent replacements in those exact same slots,
 * Oracle's gold income "15" (??) replaced with special armor 4
 * wyren's gold income "15" (??) replaced with bonus intelligence 12

PR #1514, opened 2017-12-01
Debug overlay from 2nd day of Frostivus

Chronophylos:
### TODO:
add more values


Haganeko:
Why close it? Is the work deprecated?

PR #1515, opened 2017-12-01
Refactor Glaives of Wisdom code

Trildar:
- Make cast range equal attack range
- Add special attack sound
- Removed hacky detection of manual casts
- Added a bunch of checks for things Glaives shouldn't proc on


chrisinajar:
Awesome, thanks! :+1:

PR #1523, opened 2017-12-02
Temporary Covfefe rebalance

Yzanas:


chrisinajar:
Why?

l34Um1:
I disagree with this.

chrisinajar:
Seems like a strict downgrade to me, and balance team agrees.

PR #1525, opened 2017-12-03
Charger New Custom Model

Anyangas:
This Charger custom model was made only for testing. The vmdl file "charger_test.vmdl" contains material, animations, mesh, mesh physics, and hitboxes.


warpdragon:
what happens when you open it in the tools?

Anyangas:
When you open it with the model editor you can see it properly, the mesh and the animations work fine there. I tried changing the "spirit_breaker.vmdl" model from "npc_dota_boss_charger.txt" (I also had to change the scale of the model to 300, in future models this can be fixed easily) and I could see the mesh properly in game. The charger boss functioned perfectly with the new custom models, but the animations weren't working, probably need to change some coding for that. Right now I can't do that, but I wanted to make PR so anyone that wishes to can start working on implementing the new custom animations.

carlosrpg:
Anyangas, your model is too small, you may want to make it a lot bigger. the joints are larger than the model itself.
Regarding animation. they are grouped (anim) but they must be on the root I think.
Also you need to assign Sequences like this:
image
You should aim to get something near that structure.
The Activity is how the lua script know how to call the animation, so it must match the common ones and the abilities.

Anyangas:
Yeah, the model is really small, but if you set ModelScale to 210 in the "npc_dota_boss_charger.txt" file it works fine. I could get the "stunned" animation to work perfectly, but the other ones are really weird, some play really slowly, others  stop at the middle of it. If anyone could get them to work it would be great, I've been trying for like 3 hours and can't get to make all the anims work.

Anyangas:
I fixed the animations and made the model bigger. The only thing not working is the channel animation, but the activity properly attached to the animation. A workflow document has also been added. Here are some images of the model in-game. 

test4

test3

PR #1527, opened 2017-12-03
Make aghs give both talents as it upgrades

chrisinajar:
Also makes it droppable again because that works fine now.


Zarnotox:
YES!

PR #1533, opened 2017-12-04
Rework Arcane Orb

Trildar:
Now does physical damage and can target Ancients.

Also now counts as cast on attack rather than on projectile impact. No idea why the vanilla one even does that. Especially since the note on the ability says it spends the mana before calculating damage.

Apparently the int steal in the vanilla also does some weird things with enemy mana rather than simply preserving percentage like most stat modifications. I couldn't figure out how to reuse the in-built modifiers though, as it seems to expect some table value for the int steal amount. So now the int reduction works like any other with respect to mana.


Haganeko:
Isn't not targeting ancients part of how it works in vanilla? Also, was OD still overpowered?

chrisinajar:
Is there a ticket associated with this? What's the reason for this change?

Haganeko:
It was an old ticket we never implemented and closed once it became clear Silencer was really the only problematic one.

Trildar:
Oh? Guess I should have asked first. Still though, is OD considered fine as is? I know there were some minor complaints about the fact that he struggles to do any damage to bosses since he can't use Arcane Orb on them.

Haganeko:
Similar to how he can't do anything to buildings in Vanilla

PR #1535, opened 2017-12-04
Fix Shadow Shaman +60 Serpent Ward Damage Talent being interpreted as +60 Serpent Wards

Trildar:


chrisinajar:
and the serpent wards already respect the talent and gain the damage?

Issue #1544, opened 2017-12-06
Make Unstable Current respect cooldown


Haganeko:
It doesn't.

Trildar:
There's a need for a general KV parity update really. Apparently Valve updated more stuff with tooltip values and such. Hence why Travis is having a fit.

chrisinajar:
Still? I really think this is fixed, and if it's not it's probably already detected by Travis. We should avoid opening tickets for stuff Travis already is forcing us to fix anyway.

PR #1545, opened 2017-12-06
Attempted to Fix Razor

Haganeko:
#1544 
Would be great if someone could check it works. (Still don't know how to lol)


imaGecko:
this should really be part of a wholesale fix of everything

chrisinajar:
gecko did it first and better :P

Haganeko:
Yeah. I was first, though! XP

PR #1546, opened 2017-12-06
Fix travis

imaGecko:
Valve finally decided to actually put linkedspecialbonus in for their talents


imaGecko:
technically you don't need to include lines that are not overwritten, could put em all in for completions sake

Haganeko:
Then how come the linked special bonus operation broke? We're not replacing it with anything.

imaGecko:
we are replacing the value of the abilityspecial key, and within that block we are replacing a vanilla value with a linked bonus with just a custom value without linked bonus

PR #1552, opened 2017-12-07
Lucius LongClaw

carlosrpg:
Lucius LongClaw from SiltBreaker


chrisinajar:
I suspect the summons need better "go home" code....
but.....
image

PR #1553, opened 2017-12-07
Fixed Scepter of Sanctity

Haganeko:
Tooltip was wrong on all levels, and the last one was different from the others. Plus it didn't trip linkens.


imaGecko:
active name would also need to be changed

PR #1554, opened 2017-12-07
Fixed Ursa ultimate lacking duration

Haganeko:
It was missing a linked special modifier.


Haganeko:
Apparently it already has it. No idea why it's erroring, then.

PR #1555, opened 2017-12-07
Bloodstone now starts with 14 charges (7.07)

Yzanas:
on this episode on literally unplayable


chrisinajar:
edit bloodstone.lua as well

Issue #1521, opened 2017-12-01
Add HotS style capture points after defeating a boss


chrisinajar:
Instead of rewarding the team who gets the last hit on a boss, require them to hold the arena for a short moment to capture it. This makes boss sniping much more reliably go in the favor of the team who can win a fight, instead of who gets lucky / axe cheese...

Should look like HotS camps or like a much larger color coordinate revive particle from the dota tombstones.

l34Um1:
Here are some screenshots of how it works in HotS.

First you kill the camp (boss). Then the capture point unlocks. 

![2](https://user-images.githubusercontent.com/24982519/33502120-2bf41b7a-d6df-11e7-84d0-4359c82606d0.jpg)

If a hero enters the capture point the point starts being captured for the team. This process is halted if someone from the enemy team enters the capture point. If all heroes leave the capture point it resets to starting state. Capturing a point should take 10s. This is not sped up by having multiple heroes in the circle.

![unbenannt-4](https://user-images.githubusercontent.com/24982519/33502052-f76abe22-d6de-11e7-8a75-b19ae4b66738.jpg)

If a team manages to fully capture the point they get the reward. In our case cores.

![unbenannt-3](https://user-images.githubusercontent.com/24982519/33502250-aa6e5a10-d6df-11e7-9e94-3b08cdadd482.jpg)

warpdragon:
lol i forgot baumi had started playing HOTS. Now it all adds up.

warpdragon:
so we were thinking of using the rosh timer for this purpose. Basically it would be a dial that could turn right or left, and then once it reached the end of the dial it would be "captured". it's very similar to HOTS but not quite. I wonder if the little HOTS status bar would be easier?
![image](https://user-images.githubusercontent.com/13878439/33502586-b0800aba-d6ae-11e7-846a-7ee40edb2fd4.png)

l34Um1:
Status Bar makes it much more obvious which team is capturing. Can we color the Roshan timer to indicate which team is capturing atm? If so then that's perfect.

warpdragon:
theoretically

Zarnotox:
Would this also be necessary if only one team damaged the boss?

On the one hand I think it would be annoying to capture it every time even without being contested.
On the other hand it might be harder for people to understand why you have to capture the point if it never showed up on earlier uncontested boss kills.

l34Um1:
Yes it always shows up.

warpdragon:
[...]

PR #1567, opened 2017-12-08
Boss capture point and precache restructure

Trildar:
Implementing #1521. I'm currently assuming the capture pauses when it's contested rather than resetting. Is that correct? Also, how big should the capture point be?

Current state:
Spawns on death of boss. Seems to correctly animate to indicate capturing.
Does not yet actually do anything. Probably want to change how and when the colour changes to better indicate capturing team.

Also, the main ring particle (the only one that shows when the CP is empty) seems quite flaky. It flickers a bunch for some reason. No idea if it's just my computer. Would be great if someone could check that for me.

Credits to @carlosrpg for hacking together the initial particle.

### DIFF
--- a/game/scripts/vscripts/modifiers/modifier_boss_capture_point.lua
+++ b/game/scripts/vscripts/modifiers/modifier_boss_capture_point.lua
@@ -60,7 +60,7 @@ if IsServer() then
     self.radius = keys.radius or 300
     self.captureTime = keys.captureTime or 10
     self.captureProgress = 0
-    self.thinkInterval = 0.025
+    self.thinkInterval = 0.02
     local parent = self:GetParent()
     self.captureRingEffect = ParticleManager:CreateParticle(particleDirectory .. "capture_point_ring.vpcf", PATTACH_ABSORIGIN, parent)
     -- Particle colour
--- a/game/scripts/vscripts/modifiers/modifier_boss_capture_point.lua
+++ b/game/scripts/vscripts/modifiers/modifier_boss_capture_point.lua
@@ -1,0 +1,155 @@
+modifier_boss_capture_point = class(ModifierBaseClass)
+
+local particleDirectory = "particles/capture_point_ring/"
+
+function modifier_boss_capture_point:IsPurgable()
+  return false
+end
+
+function modifier_boss_capture_point:IsHidden()
+  return true
+end
+
+-- Looks up the name on self and cleans up the particle and index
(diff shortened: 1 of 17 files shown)


Trildar:
Also made separate precache list file since the list of things we're precaching is getting pretty long.

game/scripts/vscripts/modifiers/modifier_boss_capture_point.lua
```diff
@@ -0,0 +1,155 @@
...
+    -- Controls how much of the dial to spawn. 1 is the full circle
+    ParticleManager:SetParticleControl(self.captureClockEffect, 11, Vector(0, 0, 1))
+  end
+  ParticleManager:SetParticleControl(self.captureClockEffect, 3, self:GetColor())
+end
+
+function modifier_boss_capture_point:GetColor()
+  local neutralColor = Vector(255, 150, 0)
+  local radiantColor = Vector(0, 162, 255)
+  local direColor = Vector(5, 155, 255)
```
carlosrpg: The dire color is wrong in think. This is should be (241, 5, 5 ) or some sort of red

warpdragon:
image
soemthing like this

carlosrpg:
I figure out how to fix the particle ring problem, and how to attach the control point.
Still I could not figure out how to send a callback to the modifier.

Trildar:
Ready for merge. @l34Um1 Current size is 300 units radius and the capture pauses without resetting when contested. It also does not care about units taking damage. Is that all as intended?

l34Um1:
It should reset. Ideally it goes down at half the speed it went up when nobody is capturing. Size is cool, saw the video, looks great

Trildar:
So it does a slow reset both when contested and when there's nobody in the capture point?

l34Um1:
No when it is contested, meaning both teams have players in the circle it stays at the same it was before. If the circle is empty it slowly goes down

Trildar:
And when it's being captured by the opposite team from previous? e.g. Radiant was inside for 5 seconds, then left, then Dire stands in it.

l34Um1:
In that case it should go down faster. Maybe 1.5 the regular speed. Once the team starts capturing it for themselves it goes back to normal speed though

Chronophylos:
I'm having problems displaying the Particle in the Asset Browser

chrisinajar:
`    game/scripts/vscripts/components/boss/ai.lua:82:10: (W113) accessing undefined variable IsPlayerTeam`

PR #1569, opened 2017-12-09
Make fountain target wards

Chronophylos:


Trildar:
This doesn't seem to actually work for some reason.

chrisinajar:
I bet `fountain_attack.lua` does it's own scan with hardcoded flags instead of using the getter... also damn you chrono, test your code :P

Trildar:
I checked. The code should use the ability's flags. So I dunno.

Trildar:
`DOTA_UNIT_TARGET_OTHER` is invalid in KV for some dumb reason. Fixed in #1686 by adding the target type in Lua instead.

PR #1572, opened 2017-12-09
Trying to make Greevils Greed do anything

Nortlen:
Because our creeps give much more gold than regular ones I think it is needed to increase how much gold you get from greevils greed. I mean maxed out greevils greed gave you less gold then the weakest creeps at the beginning of the game.


l34Um1:
I hope to make it be % based. Just a flat number increase is not the way to go.

warpdragon:
Make it an issue for the coders and close this

PR #1576, opened 2017-12-10
Bind Music to InGame Volume Slider

carlosrpg:
Update music .vsndevts to KV3 and change music type to battle so Dota interprets it as music and allow control its volume on InGame Music volume slider


chrisinajar:
No need for lua changes? That's awesome. Good fucking work homie!

PR #1593, opened 2017-12-15
Update contributors list with new folks (and Zarnotox)

chrisinajar:


Zarnotox:
(and Zarnotox) lmao

But apart from great memes, I don't think I contributed to OAA art :smile:

chrisinajar:
You made a really really early version of the map.

PR #1600, opened 2017-12-15
Allow referencing other tooltip values with a tooltip

chrisinajar:
This should reduce the amount of effort required to maintain accurate tooltips as DoTA updates to near negligable. The only thing we must continue to maintain is when keys are added / removed. When keys are removed travis will throw errors.

PR #1610, opened 2017-12-16
MoM Description fix

OTAJumperjay:
It still said MoM is an UAM even though those were removed in 7.07


imaGecko:
This fix will be irrelevant/wrong with the changes from #1600

chrisinajar:
@OTAJumperjay want to change these to the new style tooltips?

`"DOTA_Tooltip_Ability_item_mask_of_madness_5_Description"  "#{DOTA_Tooltip_Ability_item_mask_of_madness_Description}"`

OTAJumperjay:
As i said yesterday i'll do them.. can i somehow put multiple things into a pull request so i don't make a pr for every single item

imaGecko:
just change multiple files before commiting, also you're already doing it? I just started :D

OTAJumperjay:
imaGecko i just wanted to start doing these we should VC so we don't make additional unnesseccary work

PR #1615, opened 2017-12-18
"Fix" aura upgrade items

carlosrpg:


carlosrpg:
This is actually a workaround, since all aura items have some sort of thinker with 1 sec think time. (you can drop and pickup again and will not lose the aura if it was shorter than 1 sec

game/scripts/vscripts/modifiers/modifier_aura_item_upgrade.lua
```diff
@@ -0,0 +1,90 @@
...
+    if itemOnUpgradeSpot ~= nil then
+      hero:AddItemByName( itemOnUpgradeSpotName )
+    end
+
+    UTIL_Remove( self:GetParent() )
+	end
+
+end
+
+function modifier_aura_item_upgrade:OnDestroy()
```
Trildar: Why is there an empty function?

game/scripts/vscripts/modifiers/modifier_aura_item_upgrade.lua
```diff
@@ -0,0 +1,90 @@
...
+    local upgradeItemSlot = self:FindItemSlot(upgradeItem)
+    if upgradeItemSlot ~= self.ItemSlot and  upgradeItemSlot > -1 then
+      hero:SwapItems(upgradeItemSlot, self.ItemSlot)
+    end
+
+    if itemOnUpgradeSpot ~= nil then
+      hero:AddItemByName( itemOnUpgradeSpotName )
+    end
+
+    UTIL_Remove( self:GetParent() )
```
Trildar: I'd personally use `ForceKill`, but I don't know that it really matters.

game/scripts/vscripts/modifiers/modifier_aura_item_upgrade.lua
```diff
@@ -0,0 +1,90 @@
...
+    -- Only remove the item if it has aura upgrade
+    for _, value in ipairs(self.AuraItems) do
+      if string.find(self.ItemName, value) then
+        -- only remove the item if it is in a active slot
+        for i = DOTA_ITEM_SLOT_1, DOTA_ITEM_SLOT_6 do
+          local item = hero:GetItemInSlot(i)
+          if item then
+            if item:GetName() == self.ItemName then
+              self.ItemSlot = i
+              hero:RemoveItem(item)
```
Trildar: Should probably save the handle to the item, use `TakeItem` then use `AddItem` to give it back again so that you don't destroy the item and create a new one. Example of this can be found in `courier.lua`.

game/scripts/vscripts/events.lua
```diff
@@ -409,6 +410,9 @@ function GameMode:OnItemCombined(keys)
 
   -- The cost of the item purchased
   local itemcost = keys.itemcost
+
+  local hero = player:GetAssignedHero()
+  local hthinker = CreateModifierThinker( hero, hero , "modifier_aura_item_upgrade", { ItemName = itemName, PlayerId = plyID}, hero:GetOrigin(), hero:GetTeamNumber(), false )
```
Trildar: I'm not sure why you're using a modifier thinker in the first place. Also, the second argument is the ability the modifier is attached to, so you shouldn't be putting `hero` there. Use `nil` if there's not related ability.

Trildar: I don't think you should be using a modifier thinker in the first place, but regardless, the second argument is the ability the modifier is attached to, so you shouldn't be putting `hero` there. Use `nil` if there's not related ability.

game/scripts/vscripts/modifiers/modifier_aura_item_upgrade.lua
SquawkyArctangent: ForceKill kills the unit, which will remove it eventually, while UTIL_Remove flat-out removes it without killing it.

With thinkers, the latter is far more preferable than the former.

Trildar: The modifier (and thus the thinking) would be destroyed on death anyway though, unless you specifically made the modifier persist through death. Though there admittedly isn't much reason to have the thinker unit entity hanging around either.

SquawkyArctangent: Yes, but it would also die. Which would mean it would trigger on-death effects and such, which could require reworking some of our code to accommodate for that. Heck, Valve's code might not even be designed to accommodate thinkers dying.

carlosrpg: I copied the pattern from SiltBreaker. UTIL_Remove( self:GetParent() )  is used there on thinkers. IF there is a reason for that I don't know

game/scripts/vscripts/modifiers/modifier_aura_item_upgrade.lua
carlosrpg: I was tracking destruction but forgot to remove

game/scripts/vscripts/modifiers/modifier_aura_item_upgrade.lua
carlosrpg: Valve documentation sucks....
I saw the TakeItem but on API it just says 

TakeItem | void TakeItem(handle hItem) | Hidden item

game/scripts/vscripts/events.lua
chrisinajar: Put this into a component and observe the event with `GameEvents`.

carlosrpg:
I tried to make it work with modifiers and I couldn't do it.

1 - there are usually multiple modifiers per aura
2 - could not make the modifiers be bounded to the item. (if I remove the auras a parent modifier dies, if I readd all modifiers they become binded to the hero even having the same parent, caster and ability)

This was supposed to be a workaround to avoid the problem (today is much worse having the item and not having the correct values)

Anyway, I could not manage to make it work messing with the modifiers only. Do you guys want to add this as a workaround? if so let me know and I'll add the suggested changes, If not you can comment and close the PR.

carlosrpg:
I need feedback on this. Will we keep the way it is now or do this workaround?
I agree that the modifier thing would be better but it seems to be out of my league

PR #1621, opened 2017-12-18
Change boss drop to use PlayerResource:GetSelectedHeroEntity

Trildar:
This avoids errors with nil players caused by disconnects. Also allows the items to actually be granted to disconnected players. Hopefully this doesn't introduce some other new and interesting bug.

### DIFF
--- a/game/scripts/vscripts/components/boss/ai.lua
+++ b/game/scripts/vscripts/components/boss/ai.lua
@@ -69,8 +69,7 @@ end
 
 function BossAI:GiveItemToWholeTeam (item, teamId)
   PlayerResource:GetPlayerIDsForTeam(teamId):each(function (playerId)
-    local player = PlayerResource:GetPlayer(playerId)
-    local hero = player:GetAssignedHero()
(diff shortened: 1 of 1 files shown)


Haganeko:
Does this use the unit the player has selected? Does this mean I can select another hero and give them my core?

Trildar:
The name would make you think that. But I'm fairly certain it doesn't. Given, I only tested with a bot that I spawned in and kicked. I believe the server actually doesn't even have keep track of what players have selected though (or at least no Lua function seems to expose it), much to my annoyance as it means I can't get it to easily print modifiers on whatever unit I have selected.

Haganeko:
Well, it sort of has to, since it's possible to see it when on player vision spectator mode

Trildar:
Good point. Valve probably just didn't see a point in making an API function for it. To be fair, I don't think there would be a huge amount of use for it anyway.

Also, the function does say "HeroEntity", and you don't necessarily have to have a hero in your unit selection all the time.

PR #1625, opened 2017-12-18
Discord Link

Chronophylos:


chrisinajar:
It should either be clickable or use an url shortener to make it memorable

Chronophylos:
I couldn't figure out how to make it clickable.

PR #1627, opened 2017-12-18
Make each extra hero worth .5 more when capturing

chrisinajar:


game/scripts/vscripts/modifiers/modifier_boss_capture_point.lua
```diff
@@ -120,6 +123,7 @@ function modifier_boss_capture_point:OnIntervalThink()
     -- Point is being captured by a team
     captureTick = self.thinkInterval
   end
+  captureTick = captureTick * math.max(1, heroMultiplier/2)
```
Trildar: Doesn't this only increase the capture rate for 3 heroes or more?

chrisinajar: yeah, i guess it really shoulda been `1 + hm/2`

Trildar: `0.5 + hm/2` 1 would make the first hero capture at 1.5x speed. :P

game/scripts/vscripts/modifiers/modifier_boss_capture_point.lua
```diff
@@ -99,12 +99,15 @@ function modifier_boss_capture_point:OnIntervalThink()
...
   local captureTick
+  local heroMultiplier = 0
 
   -- Start capturing from neutral
   if radiantUnits[1] and self.capturingTeam == nil then
     self.capturingTeam = DOTA_TEAM_GOODGUYS
+    heroMultiplier = #radiantUnits
   elseif direUnits[1] and self.capturingTeam == nil then
     self.capturingTeam = DOTA_TEAM_BADGUYS
+    heroMultiplier = #direUnits
```
Trildar: Also just realised this is actually the wrong place to count the heroes as this block only runs when capture starts from neutral.

PR #1639, opened 2017-12-20
Redo gold and experience

Haganeko:
Added more experience early.
Added a lot more gold, especially after the first ten minutes.
Fixed a math mistake which made cave give less gold.
Would like to (but dare not) put boots in progression with the gold:
Arcanes 670/1830/4000/7200/11400
Tranquils 1000/3000/6700/12000/19000
I hope no one will find these values unreasonable.

### DIFF
--- a/game/scripts/vscripts/components/cave/cave_types.lua
+++ b/game/scripts/vscripts/components/cave/cave_types.lua
@@ -42,12 +42,12 @@ CaveTypes = {
   [1] = { -- 1 "Howl's it Going?"
     {                                                  --HP  MANA  DMG  ARM   GOLD   EXP  RESIST
       units = {
-        {"npc_dota_neutral_custom_cave_big_pupper",    600,    0,  30,   1,   141,   307,  30},
-        {"npc_dota_neutral_custom_cave_big_pupper",    600,    0,  30,   1,   141,   307,  30},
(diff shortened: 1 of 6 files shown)


Haganeko:
Note one of the commits is misnamed - I actually updated passive gold on e8a6dfc

Haganeko:
This is the current arcanes vs new gold:
arcanes
This is what I'm proposing:
arcanesproposed

Haganeko:
Equivalent charts for tranquils. Current:
tranquils
And proposed:
tranquilsproposed

Haganeko:
For ease of visualization, here's the expected changes in projected level and gold:
gold
level
Do note that late game levelling will be a bit slower, but the "one level per minute" trend is maintained more closely.
I'll probably still tinker with this tomorrow.

l34Um1:
The boots shouldn't be messed with until we have the new mechanics in.

As for xp the old graph does not match playtesting. You have to rethink the graphs, people do not cave before 10 minutes. So add that missing xp in somewhere. Maybe all passive?

Haganeko:
I can go back to the very old model where I front loaded the experience. Can we try the other stuff, at least? Especially correcting the math mistake I made.
Used to be people would leave one hero at base to grab first cave room. Wonder if that could come back?

Haganeko:
Values all updated. Would have liked permission to update boots, but this is as ready as it's getting this side of New Year's.

PR #1647, opened 2017-12-21
Boss Balance

l34Um1:
Removed spider lifesteal, made ogre less tanky, nerfed lycan wolves and hp, buffed lycan damage


Haganeko:
The lifesteal was too much even with the reduction?

Yzanas:
spider lifesteal was already rebalanced in #1644

PR #1659, opened 2017-12-26
Added far sight

jb1361:
Still needs a icon for the item.


warpdragon:
icons eh?

warpdragon:
icons eh? Something like this maybe 
image

l34Um1:
It needs to level up as other items do. Recipe costs need to match as well. I added in stats for all 4 levels of the item.

PR #1660, opened 2017-12-27
Reflex items

Derk00:
These are the old reflex items but without reflex cores


imaGecko:
doesn't return any errors on my local files

imaGecko:
Tested the items, seem to work just fine.
Regen Crystal gives an error when upgraded while the active is still running

Cannot reproduce the travis error

chrisinajar:
Once this is done we get to schedule the tournament :D

imaGecko:
Did the nuking option of just copying everything over to a clean branch. Should be 100% exactly the same changes

Derk00:
I see that it works now, but why would that fix it tho? You are copying everything, so if there were any errors, they are copied as well right?

PR #1664, opened 2017-12-29
some fancy assed coding right there.

warpdragon:
Where my free points!?


Haganeko:
Dammit warp you don't need free points.

PR #1674, opened 2017-12-30
Some Far Sight fixes

Trildar:
* MODIFIER_ATTRIBUTE_MULTIPLE for modifier_item_far_sight
* Add true sight within Far Sight's vision radius


chrisinajar:
Should there be a generic modifier for revealing? we now do it in two different spots, here and scan

Trildar:
Possibly. Scan is not tied to an ability though, which could complicate trying to make it generic over both cases. It should probably be applying the vanilla `modifier_truesight` rather than having its own modifier though.

PR #1676, opened 2017-12-30
Add a bunch of people to contributor bottle icon list

Trildar:
List as follows:
* Chronophylos
* Trildar
* SquawkyArctangent
* imaGecko
* Haganeko
* Warpdragon
* Honeth
* Minnakht
* Allan vbhg
* Apisal
* carlosrpg

Hardly complete, but it's a start.

Also disable debug printing in `bottle.lua` and modify the id lookup to use `or`.


chrisinajar:
Logging is intentional so that people can send us their steam id's super super super easy

PR #1681, opened 2017-12-30
Fix duels item cooldown abuse

chrisinajar:
Currently you can move around items to refresh cooldowns from duels, this is especially bad with midas and refresher. This fixes it so that any item whose cooldown was refreshed for a duel has its cooldown set again after the duel.


Haganeko:
What happens if one gets to courier to take the item from them? Or uses it in a recipe?

chrisinajar:
courier = restored correctly, upgraded = yeah you get your free refresh WHATEVER

PR #1688, opened 2017-12-31
Fix Giant Form interaction with other model scaling effects

Trildar:
The scaling is possibly slightly smaller than it was previously. Not entirely sure.


chrisinajar:
HAH, i thought that modifier just didn't work at all, good to know it's just a different unit

PR #1690, opened 2018-01-01
Small tooltip update

imaGecko:
Added notes, fixed typos. Rather random tooltip fixes, just went through the files and fixed a few small things.


Trildar:
The bonus attack speed tooltip for Echo Sabre is missing the closing brace. Could add fixing that here.

PR #1714, opened 2018-01-04
Fix git ignore

carlosrpg:


carlosrpg:
I think this would be a merge, It will make dealing with that compiled of files on git so much better!

chrisinajar:
We've been burned by doing this before so i'm hesitant, but i know how to recompile things locally a lot better now. i'm merging this because the OAA philosophy has always been "move fast and break things, someone will fix it"

PR #1716, opened 2018-01-04
Replace Underlord, Shadow Fiend and Viper's building dmg talents

Yzanas:
balance discussed with JumperJay, Naeil and Chogex


Haganeko:
4 MP regen? Wouldn't that be by far the highest MP regen talent? That's the same as level 4 Maiden aura self regen!

Yzanas:
1) Viper has really low Intelligence compared to heroes with similar mana regen talents (Techies, Storm,...) so it doesn't get amped as much. Also the 4 mana regen is taken from Techies' lvl 10 talent of all things.
2) You can't compare a talent to a teamwide aura at all. If anything CM aura gives a net 20 regen

Haganeko:
Nah. Maiden aura gives everyone else 1.4
Only she gets the 4
In any case, techies is a very mana intensive hero with terrible stats. Give viper this, and he'll never have mana problems again.

PR #1717, opened 2018-01-04
Fix item upgrade modifier persist on death

carlosrpg:


game/scripts/vscripts/modifiers/modifier_aura_item_upgrade.lua
```diff
@@ -8,8 +8,12 @@ function modifier_aura_item_upgrade:IsPurgable()
   return false
 end
 
-function modifier_aura_item_upgrade:GetAttributes()
-  return MODIFIER_ATTRIBUTE_MULTIPLE
```
Trildar: Why was this removed? I'm pretty sure it is 100% necessary. Without it, it won't allow multiple instances of the modifier on the hero and it will probably not work properly when items are combined in quick succession.

game/scripts/vscripts/modifiers/modifier_aura_item_upgrade.lua
```diff
@@ -50,6 +54,9 @@ if IsServer() then
         end
       end
     end
+
+    self:Destroy()
```
Trildar: Was this tested? I would think it would cause the modifier to destroy itself before it has a chance to run its think function.

Trildar: Oh. Duh. It returns early when it finds an item.

game/scripts/vscripts/modifiers/modifier_aura_item_upgrade.lua
carlosrpg: You are right

game/scripts/vscripts/modifiers/modifier_aura_item_upgrade.lua
carlosrpg: Without the DestroyOnExpire it was getting destroyed after the first think

game/scripts/vscripts/modifiers/modifier_aura_item_upgrade.lua
chrisinajar: Truth, I only tested with 1 item at a time.. Easiest way to test is get everything ready for combining in stash then walk into the other well. Items will instant combine and you die a quarter second later.

PR #1722, opened 2018-01-05
Strafe Balance

Nortlen:
Increased cooldown of Strafe, so it isn't zero.


imaGecko:
Can very easily agree with the change

Haganeko:
Seems okay.

l34Um1:
keep at 15

Haganeko:
It does already progress the AS value.

Nortlen:
Changed the cooldown to stay at 15

Issue #1603, opened 2017-12-15
Scan doesn't Scan


l34Um1:
Currently Scan only detects but doesn't have the original Scan effect anymore. It should still Scan.

carlosrpg:
Fixed

PR #1723, opened 2018-01-05
Make Scan do the vanilla Scan plus reveal

Trildar:
Also added a tooltip and icon for the debuff the reveal adds.

Turns out it wasn't actually that hard to put vanilla Scan back in as apparently it just uses a modifier thinker. The duration is currently just like vanilla, which makes it a little shorter than the reveal effect (8 vs 14).


Trildar:
Closes #1603

PR #1732, opened 2018-01-07
Small Temple Guardians fixes

carlosrpg:
Fix Spawns for Temple Guardian. Fix Deaggro animation, reduce vision (aggro range) and reduce leash


imaGecko:
doesn't 100% status resistance still interrupt the current action?

PR #1738, opened 2018-01-09
Add recipe tooltips

imaGecko:
This should remove the "UNKNOWN" crash when alt clicking recipes
the crash will still happen for any other thing that you can ping that doesn't have tooltips, e.g. modifiers missing a tooltip


imaGecko:
done and done

PR #1740, opened 2018-01-11
Language file for Afrikaans Language in South Africa

devorama:
Create a Addon language file for Afrikaans Language in South Africa


chrisinajar:
That's not how translations work. Join discord and ask around in `translation-help`, we do all our translations on a 3rd party website.

PR #1745, opened 2018-01-11
[WIP] New Boss Tiers

imaGecko:
Move all high tier bosses to t4 at most
Create t5 version of every boss

To do: 
Separate Ability stats for t5 version
Balancing values


Haganeko:
So tier 6 is not a thing anymore?

l34Um1:
serves no purpose

l34Um1:
I think we can close this, seeing as the other PR is the same thing but finished.

PR #1751, opened 2018-01-13
Glyph/Scan Icons and CooldownOverlay

carlosrpg:


Trildar:
Unfortunately, the TypeScript compiler doesn't have options for indentation and string quote style, so it seems like the options are pretty much: A) Set up a JS formatter. B) Change the semistandard settings. C) Don't keep compiled JS in git.

Why is semistandard configured for single quote strings anyway? And speaking of, maybe we should get luacheck to enforce quote style if it can. It's currently inconsistent.

carlosrpg:
A) Set up a JS formatter: What do you mean by that?
B) Change the semistandard settings: I don't think it is possible
C) I don't like that, someone will forget to compile and since the normal tools does not complile it It will cause errors.

I change the styles myself this time, but we need to decide how are we going to do this in the future

chrisinajar:
The long term solution is to have all ts files compile into js files that have a common symbol in their name or a common directory, just a `dist/` or something. Then you make semistandard ignore that pattern.

The proper build system stuff can be done in another PR. Also, fwiw, `semistandard -- fix`.

PR #1760, opened 2018-01-16
Balance

l34Um1:


Haganeko:
May I also have a look at the boots? I'll be good, I promise!

PR #1765, opened 2018-01-17
Fix gold amount display in kill toast message

Trildar:
Lots of other changes in the process as well. One major one being refactoring hero_kills.lua and fixing some bugs; details in commit message.

Also changed the gold bounty popups to show the last hit gold and assist gold separately. It's now more like vanilla Dota, where the last hit gold appears on the killed hero, and assist gold appears on the player's own hero for every player that gets assist gold. Dunno if maybe we just want to combine last hit gold and assist gold anyway, since it is easier to see that way.

~~Still TODO (though the current state is ready for merging): Fix gold distribution and toast message when a non-hero kills a hero. For us, that basically means the fountain. For some reason it's all screwy, and even the non-custom parts don't act anything like vanilla Dota. It seems to half think that the fountain is a player hero, while at the same time not.~~

P.S. Chris, why were you using the `each` function as basically just `for` loops? :P


Trildar:
Now also fixes kill gold for non-player kills, e.g. the fountain.

PR #1769, opened 2018-01-18
I don't really know whats going on

samuraiJak:
Will delete if some shit happens


Trildar:
Um. What are you trying to do exactly?

chrisinajar:
neither do i

PR #1770, opened 2018-01-19
Make capture point actually accelerate for additional heroes

Trildar:
Previously, it was only checking the number of heroes on the very first tick of a team starting a capture. Subsequent ticks would have no multiplier for additional heroes, effectively meaning the capture rate never changed.

The acceleration only applies for capturing and not reversing capture.

Do we even still want this? Seeing as it's been sitting non-functional for so long. Maybe at a reduced multiplier? Additional 0.5 per hero is quite fast.


Haganeko:
Maybe reduce capture rate for one hero to 80% current speed, then make extra heroes worth .3 extra multiplier?

imaGecko:
Personally from having played with it I'd say lower the multiplier and make the current base speed speed equal to 2 or 3 heroes

Haganeko:
My idea would make two heroes 104% current speed, three 128%, four 152% and five 176%

PR #1771, opened 2018-01-19
Fix non-working uses of UseResources method

Trildar:
Apparently Valve went and broke this method for most types of abilities. Not really sure what cases still work.

- Fix Time Lock and Bash of the Deep not having a cooldown
- Fix Glaives of Wisdom not using mana


Trildar:
Apparently, the proper way to handle this is to give the items/abilities a `ShouldUseResources` method that returns true at the point you call `UseResources`. Using `StartCooldown(GetCooldown)` stuff results in the cooldown ignoring cooldown reduction cause for some reason, there's no API to get the cooldown after reduction.

PR #1773, opened 2018-01-20
Fix Hand of Midas giving 0 XP

Trildar:
It was setting the XP bounty to 0 before then querying it to calculate the XP to give. Why did I even reorder it in the first place...


Haganeko:
Does it work with bear?

Trildar:
I dunno... not sure I ever tested that. That wouldn't have changed since it was first implemented though.

Haganeko:
If you didn't code for it to give the XP to the druid when bear uses it, I assume it doesn't.

Trildar:
Vanilla Midas doesn't seem to do that. So not sure if it's something we should change.

PR #1778, opened 2018-01-21
Refactor pseudo-random distribution

Trildar:
Added a function to calculate PRD C values for arbitrary target probabilities, and changed stuff to use that instead of copying a lookup table everywhere.


chrisinajar:
I probably woulda put `cfinder.lua` in `library`

also NERD!!!

chrisinajar:
Can you move the code into libraries and also make it cache results since there's no reason to constantly recalculate them.

Trildar:
It does cache them. The function is memoized.

PR #1782, opened 2018-01-21
Make Cave consider four minutes clear

Haganeko:
As opposed to the previous eight.

### DIFF
--- a/game/scripts/vscripts/components/cave/cave_types.lua
+++ b/game/scripts/vscripts/components/cave/cave_types.lua
@@ -42,12 +42,12 @@ CaveTypes = {
   [1] = { -- 1 "Howl's it Going?"
     {                                                  --HP  MANA  DMG  ARM   GOLD   EXP  RESIST
       units = {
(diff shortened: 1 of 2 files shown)


chrisinajar:
Doesn't this make cave slower? That'd be bad for the balance of it since then a hero can more likely farm it indefinitely..

Haganeko:
It essentially means you'll need to clear it twice for the same reward as clearing once now. Difficulty will also grow at half rate.

Melongod:
I'm not sure if we want people to have to dedicate more time to farming the cave while it's as boring as it currently is.

Haganeko:
Clearly not going in.

chrisinajar:
I think at this point we should base the models on the cave rather than vice versa. If we can define specific problems with the cave then we can discuss solutions to them, but I think unless a change is fixing a specific problem then we should wait until after the tournament.

Should we get a staging branch for 3.0 for changes like this?

Haganeko:
Sure? I won't really be able to really do any real work until the tournament is over, though, and getting the rest of gold right (more early, less late) is my first priority.

PR #1800, opened 2018-01-22
Made Cave Rune

ThatMusicWriter:
I posted this as an issue like a numpty earlier. This is the actual pull request. The bottle needs a proper ID number, and I may need to add more dispell data in the LUA, unless the base DOTA modifiers for the runes are sufficient. Also, I'm using the BKB sound and empty bottle model as placeholders. I don't know the ins and outs of the DOTA sound and model library, but it will serve for now.


game/scripts/vscripts/items/cave_rune_hasted_doubledamage.lua
```diff
@@ -0,0 +1,26 @@
...
+function item_cave_rune_hasted_doubledamage:GetIntrinsicModifierName()
+
+  return "modifier_rune_haste", "modifier_rune_doubledamage"
+
+end
+
+function item_cave_rune_hasted_doubledamage:OnSpellStart()
+	local caster = self:Getcaster()
+
+	caster:AddNewModifier(caster, self, "modifier_rune_haste", {duration = 30.0})
```
carlosrpg: Does that adds the particle effects?
if not we have to add them

game/scripts/npc/items/custom/item_cave_rune_haste_doubledamage.txt
```diff
@@ -0,0 +1,38 @@
...
+	"Model"				"models/props_gameplay/bottle_empty.vmdl"
+
+	//Stats
+	//----------------------------------------------------------------------------------------------------
+	"AbilityCooldown"		"0.5"	
+	"AbilityCastPoint"		"0.0"
+
+	//Item Info
+	//----------------------------------------------------------------------------------------------------
+	"ItemCost" 			"1300"
```
carlosrpg: I think adding item cost may affect our Networth values. Maybe set it to 0?

game/scripts/vscripts/items/cave_rune_hasted_doubledamage.lua
```diff
@@ -0,0 +1,26 @@
+LinkLuaModifier("modifier_rune_haste", LUA_MODIDIER_MOTION_NONE)
+LinkLuaModifier("modifier_rune_doubledamage", LUA_MODIDIER_MOTION_NONE)
+
+item_cave_rune_hasted_doubledamage = class(ItemBaseClass)
+
+function item_cave_rune_hasted_doubledamage:GetIntrinsicModifierName()
```
carlosrpg: does it work? I think we can't set multiple modifiers with this.
Since you add them on spellStart you can remove this.

game/scripts/npc/items/custom/item_cave_rune_haste_doubledamage.txt
```diff
@@ -0,0 +1,38 @@
...
+	"ItemPermanent" 		"1"
+	"IsTempestDoubleClonable" 	"0"
+	"ItemInitialCharges"		"1"
+	"ItemRequiresCharges" 		"1"
+	"ItemShareability" 		"ITEM_FULLY_SHAREABLE_STACKING"
+	"ItemPurchaseable" 		"0"
+
+	//Special
+	//-----------------------------------------------------------------------------------------------------
+	"AbilitySpecial"
```
carlosrpg: this file is not properly closed.
Check your curly brackets.

game/scripts/npc/items/custom/item_cave_rune_haste_doubledamage.txt
```diff
@@ -0,0 +1,38 @@
+"DOTAItems"
+{
+//============================================================================================================
```
carlosrpg: Why?

ThatMusicWriter: It's in every other file of the type and looks pretty

game/scripts/vscripts/items/cave_rune_hasted_doubledamage.lua
ThatMusicWriter: According to LUA, it should work, but I am not sure about DOTA. This needs a lot of review atm

game/scripts/vscripts/items/cave_rune_hasted_doubledamage.lua
ThatMusicWriter: I cannot find the rune modifiers in the game files, so I have no idea if it includes the particles. Any idea where all the game files can be found so I can find more information?

carlosrpg: particles/generic_gameplay/rune_doubledamage.vpcf
particles/generic_gameplay/rune_haste.vpcf

carlosrpg: just use the asset browser to search select particles and search for rune_<effect>

game/scripts/vscripts/items/cave_rune_hasted_doubledamage.lua
carlosrpg: I does not work, Dota expects only one string ad result with the modifier, If it worked we would not have to call AddNewModifier on spell start because dota would do that for us.

game/scripts/npc/items/custom/item_cave_rune_haste_doubledamage.txt
carlosrpg: So pretty we need to add 2 sections of them? :D

game/scripts/vscripts/items/cave_rune_hasted_doubledamage.lua
```diff
@@ -0,0 +1,26 @@
+LinkLuaModifier("modifier_rune_haste", LUA_MODIDIER_MOTION_NONE)
+LinkLuaModifier("modifier_rune_doubledamage", LUA_MODIDIER_MOTION_NONE)
```
Trildar: Unnecessary for in-built modifiers.

game/scripts/vscripts/items/cave_rune_hasted_doubledamage.lua
Trildar: IntrinsicModifier is the wrong thing for this in any case. For items, the intrinsic modifier is a modifier that is added whenever the unit has the item active in inventory and removed whenever the item is backpacked or otherwise leaves the inventory.

game/scripts/npc/items/custom/item_cave_rune_haste_doubledamage.txt
```diff
@@ -0,0 +1,38 @@
...
+
+	//Stats
+	//----------------------------------------------------------------------------------------------------
+	"AbilityCooldown"		"0.5"	
+	"AbilityCastPoint"		"0.0"
+
+	//Item Info
+	//----------------------------------------------------------------------------------------------------
+	"ItemCost" 			"1300"
+	"ItemStackable" 		"1"
```
Trildar: Are these really supposed to be stackable?

game/scripts/vscripts/items/cave_rune_hasted_doubledamage.lua
```diff
@@ -0,0 +1,26 @@
...
+  return "modifier_rune_haste", "modifier_rune_doubledamage"
+
+end
+
+function item_cave_rune_hasted_doubledamage:OnSpellStart()
+	local caster = self:Getcaster()
+
+	caster:AddNewModifier(caster, self, "modifier_rune_haste", {duration = 30.0})
+
+	caster:AddNewModifier(caster, self, "modifier_rune_doubledamage", {duration = 30.0})
```
Trildar: Durations need to be read from KV. See `GetSpecialValueFor` method.
The double damage will likely need a custom modifier, so that the damage increase amount can be exposed to KV.

game/scripts/npc/items/custom/item_cave_rune_haste_doubledamage.txt
ThatMusicWriter: Unsure, but I thought about how Baumi does things, and it seemed like a valid strategy to clear the cave twice and double the buff duration in the next duel. Plus, it means that only one person on the team has to carry the bottles, and can just hand them out before the duel. So, with that in mind, I thought stacking would enrich gameplay.

Plus, Baumi always says to make something OP initially and then nerf it as needed.

l34Um1: Stackable in this context just means the duration extents right? In that case, yes it should be stackable.

Haganeko: It means that multiples of a single bottle will stack like tangos. Or infinity bottles.

l34Um1: Oh. Then no.

ThatMusicWriter:
I will edit the item now to remove stackability

ThatMusicWriter:
OK, the thing should work. It just needs a proper ID, and I do not know how to determine which IDs are taken.

game/scripts/npc/items/custom/item_cave_rune_haste_doubledamage.txt
Trildar: It can be changed to consume all stacks in a single use and extend duration, if that would be desirable.

l34Um1: Nono, that's unintuitive. Just unstackable is fine. They are not really bottles, they are single use consumables. Think of them like a Mango.

game/scripts/vscripts/modifiers/modifier_rune_haste_doubledamage
```diff
@@ -0,0 +1,23 @@
+modifier_rune_haste_doubledamage = modifier_rune_haste_doubledamage or class({})
```
chrisinajar: This isn't a `.lua` file

ThatMusicWriter:
I'm in over my head with the modifier; someone else will have to do that. The rest of the code is solid.

ThatMusicWriter:
I'm in over my head with the modifier; someone else will have to do that. The rest of the code is solid.

PR #1805, opened 2018-01-22
fix underlord ult hotkeys

imaGecko:
tiniest possible pull request, the second


Haganeko:
Chris's new test cases, surely?

imaGecko:
not that new, the comment must contain what was there previously and since i moved the ability slots the comments were no longer correct
just forgot about them

PR #1819, opened 2018-01-23
Rework Charger resistances, some general Charger cleanup

### LINKED ISSUE #1574: Charger and Heartpiercer
Heartpiercer disables all of the Charger's armour, making it easy to right click him without having to bother with baiting him into pillars. So the armour needs to be replaced with some other method of damage mitigation.

Trildar:
Changed Charger to used damage block like Boss Resistance instead of relying on high armor and magic resistance that gets rekt'd by Heartpiercer. Closes #1574.

Also gave the Cliff Walk modifier a tooltip, fixed the misspelling of "glancing" in the Charger files, and made Charge have half the cooldown when it's interrupted.


Trildar:
Hm. It's failing because of the icon, which is an unused ability icon in vanilla Dota files. Do the test cases only check for it matching Dota ability names? It should do the search against the folder in GameTracking.

chrisinajar:
I think the whitelist is only hero abilities because we haven't run into anything else yet, haha...

PR #1828, opened 2018-01-24
Log when networth can't be ranked properly

Trildar:
Hopefully some data will shed some light on why this might be happening.

The `GameMode` key is optional right?


Trildar:
Ugh. Having a bunch of folders not read by Luacheck is rather annoying when it comes to these globals.

PR #1829, opened 2018-01-24
Balance

l34Um1:


Haganeko:
Might want to compress the Underlord progression a bit. There's barely any time to teleport with the new level 1

PR #1838, opened 2018-01-25
Move some strings for translation integration, update game help info

Trildar:
The updating game info part needs doing. Could use some help from @warpdragon or somebody for that. Currently it still talks about Reflex Cores.


chrisinajar:
This seems dead. Reopen if not.

PR #1872, opened 2018-01-28
Prevent Doom from getting True Sight from Cave creeps

Trildar:
With the level 20 Talent, Doom can Devour the Cave creeps and thus get True Sight from some of them. He won't get the ability if it's not in the first 2 slots, so the centaur's abilities were reordered to give the attack speed aura instead. Dragon currently just has an empty ability, which looks slightly odd if you examine the creep. Dunno if we might want to give them another ability.


Haganeko:
Should keep him from getting the AS ability as well.

PR #1874, opened 2018-01-28
Fix exception found during test

carlosrpg:


game/scripts/vscripts/items/transformation/baseclass.lua
```diff
@@ -133,7 +133,7 @@ if IsServer() then
 
   function modifier_transformation_item_watcher:Stop()
     if self.mod then
-      if not self.mod:IsNull() then
+      if self.mod ~= nil or not self.mod:IsNull() then
```
chrisinajar: This is redundant since the previous conditional checks if self.mod is truthy.

PR #1876, opened 2018-01-29
Rebalance Boot Gold

Haganeko:
Let's try this. If they're overnerfed we can try something else
/Should I do the fools gold increase here too? Would be more thematic than in the middle of Baumi's balance patch.

### DIFF
--- a/game/scripts/npc/items/farming/item_greater_arcane_boots.txt
+++ b/game/scripts/npc/items/farming/item_greater_arcane_boots.txt
@@ -85,7 +85,7 @@
       "05"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "bonus_gold_per_minute"                           "1200 2100 3600 6200 9600"
+        "bonus_gold_per_minute"                           "500 1350 3000 5250 8300"
       }
       "06"
       {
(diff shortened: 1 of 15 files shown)


chrisinajar:
Wow, so fools gold is going down to 30% *and* the gold from boots is being reduced by over half? How are supports going to be able to afford anything? All the level 1's are reduced by well over half, and the level 2's are pretty close... It'll take supports well over double the amount of time to acquire their items. I was thinking like a 20-30% nerf and then a buff to travels...

Haganeko:
This is experimental. I have another set of values ready to go depending on feedback.

chrisinajar:
Lets save experiments for after the tournament.

chrisinajar:
Perhaps we should start a branch for post-tournament to start merging stuff like this into it?

Haganeko:
Are you really going to block something we discussed extensively in discord?

chrisinajar:
Not block, merge into the correct branch. We shouldn't be making radical (reducing GPM by over half is definitely radical) changes right now. Anything you consider "experimental" should wait until after. We'll *massively* destabilize the meta right after the tournament and add in all kinds of crazy shit, but we shouldn't be tableflipping the meta leading up to the tournament.

Haganeko:
There is a month till, and inscriptions are still open. Plus, this is just correcting an egregious balance oversight. This is the correct power level for the boots. The only question is whether all gold is wrong and needs to be increased.

chrisinajar:
Define "correct", how are you measuring that?

chrisinajar:
How can something be both "correct" and "experimental" with backup numbers ready to deploy in case they're wrong?..... I think maybe we need to rethink how we're measuring what these values need to be.

PR #1877, opened 2018-01-29
Boot Update in case the first one was terrible

Haganeko:
Just getting ahead of possible need.


Haganeko:
Unnecessary after all.

PR #1885, opened 2018-01-30
Vampire SFX sound effects

carlosrpg:
- Kudos to arrowpack! thanks


arrowpack:
So does it fit for Vampire Theet?

PR #1887, opened 2018-01-30
How to get away with boot removal

Haganeko:
Made some farming boots an offer they couldn't refuse. Also fixed boots being balanced XP


Haganeko:
Ha! Nailed it on the first attempt!
Now let's see how it breaks the game.

PR #1888, opened 2018-01-31
Make all boss sounds respect Fog of War

Trildar:
Also refactored a bunch of repeated calls to `GetCaster` and `GetParent` to use local variables instead.

Hopefully I didn't mess anything up in the process.

P.S. @chrisinajar I wonder if we could have automatic checks for `EmitSoundOn`. There are only a handful of places where it should actually be used since it plays sounds without respecting Fog of War.


imaGecko:
is this also possible for particles? temple guardians and ogre boss have some visible particles

Trildar:
My guess is that it's because those particles weren't attached to an entity. A little harder to search for. In the case of Guardians, I'm also a little concerned if it would become possible for players fighting them to not to see the warning circles.

PR #1894, opened 2018-01-31
Change back to vanilla manta, scale illusions

chrisinajar:
No more lag!


game/scripts/npc/items/item_manta.txt
```diff
@@ -3,36 +3,17 @@
...
   //=================================================================================================================
   // Manta Style
   //=================================================================================================================
-  "item_manta"  "REMOVED"
-  "item_manta_1"
+  "item_manta"
   {
     // General
     //-------------------------------------------------------------------------------------------------------------
-    "ID"                                                  "4048"                            // unique ID number for this item.  Do not change this once established or it will invalidate collected stats.
```
zelding: you should write back the old id `147`

PR #1898, opened 2018-02-01
Fix Meepo Death behaviour

carlosrpg:
I hate meepo.
 - Meepo talents upgrade is still broken


Haganeko:
Don't hate the hero, hate the designer. And maybe whoever it was that thought "enable meepo" was a good stretch goal XP

PR #1910, opened 2018-02-01
Check for AFK / Fountain idling players

Melongod:


carlosrpg:
It seems like this flag is broken now (causing massive UI lags) . Let's wait a little bit before check this in

PR #1923, opened 2018-02-03
Make a bunch of attack modifiers respect process_procs flag

Trildar:
List includes:
* Arcane Orb
* Elder Dragon Form
* Glaives of Wisdom
* Covfefe
* Greater Phase Boots
* Greater Power Treads
* Giant Form

The process_procs flag is what indicates when attacks like Medusa's Split Shot should not have attack modifiers applied to them.

### DIFF
--- a/game/scripts/vscripts/items/farming/greater_phase_boots.lua
+++ b/game/scripts/vscripts/items/farming/greater_phase_boots.lua
@@ -73,8 +73,11 @@ modifier_item_greater_phase_boots_splinter_shot.OnRefresh = modifier_item_greate
 
 function modifier_item_greater_phase_boots_splinter_shot:OnAttackLanded(keys)
   local parent = self:GetParent()
-  if keys.attacker == parent and not self.doReduction then
+  if keys.attacker == parent and keys.process_procs and not self.doReduction then
     local ability = self:GetAbility()
+
+    Debug:EnableDebugging()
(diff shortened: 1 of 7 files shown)


Haganeko:
So covfefe and Giant Form don't stack? Or does Medusa not get covfefe even with the talent? Or both?

Trildar:
I'm not sure if you mean with or without this change, but that's all off mark in either case.
Not checking for the flag means that all these effects currently apply to things like Medusa Split Shot without the Talent and even the splinters of Greater Phase Boots (though obviously Phase Boots already doesn't proc on itself).

game/scripts/vscripts/items/farming/greater_phase_boots.lua
```diff
@@ -73,9 +73,12 @@ modifier_item_greater_phase_boots_splinter_shot.OnRefresh = modifier_item_greate
 
 function modifier_item_greater_phase_boots_splinter_shot:OnAttackLanded(keys)
   local parent = self:GetParent()
-  if keys.attacker == parent and not self.doReduction then
+  if keys.attacker == parent and keys.process_procs and not self.doReduction then
     local ability = self:GetAbility()
 
+    Debug:EnableDebugging()
+    DebugPrintTable(keys)
```
chrisinajar: Can you remove this? gonna spam....

PR #1924, opened 2018-02-03
Fix Assault Cuirass 5 cast range

### LINKED ISSUE #1917: Assault cuirass 5 has wrong cast range
it has `AbilityCastRange` 25 instead of 900

filalex77:
Assault Cuirass 5 should have a cast range of 900, not 25. Fixes #1917


imaGecko:
fix your titles

PR #1928, opened 2018-02-03
Make Vampire Tooth respect process_procs flag

Trildar:
Also disable the passive lifesteal during the active.

And some minor tweaks to the visual effects.


carlosrpg:
Baumi requested the return projectile speed to be faster

Trildar:
I wouldn't know about that. But I'm pretty sure he said it'd be nice if it moved slightly slower after it was changed to 1200.

PR #1929, opened 2018-02-03
Use Lua's bit32 library

filalex77:


Trildar:
Dota does not have `bit32`, it has `bit`, the LuaJIT BitOp library: http://bitop.luajit.org/

filalex77:
@Trildar sure, thanks! Will patch soon.

chrisinajar:
Yup, this is why I chose that specific implementation of sha over the others.

Is there a problem with the current hashing function that you're trying to fix? It works fine currently, confirming checksums in node.

Closing this for now since I believe it was opened under a false assumption. I'll reopen if there's something to fix here, so just let me know.. Thanks!

filalex77:
@chrisinajar yes, just an assumption. sorry for bothering

PR #1930, opened 2018-02-03
Add Chat on Picking Screen

carlosrpg:
-DOTA_GAMERULES_STATE_GAME_IN_PROGRESS event seems to be broken (running while still in picking screen)
 - Added new event to oaa_picking_finished


carlosrpg:
This is not working with Captains Mode yet. I don't know how to test it locally and we need to figure out a new layout to get fit the chat box, otherwise this will block Player selection on CM

chrisinajar:
There already is an event, OnGameInProgress runs when all the heroes are done spawning into the game.

PR #1938, opened 2018-02-04
Add wrapper for hero:AddExperience

Trildar:
This is to make it go through the ModifyXP filter since for some reason Valve apparently didn't make it do so.

Currently denies any call to `AddExperience` on bots due to their connection state never being "connected". Not sure whether there's a need to fix that.

Also could use some testing to see if granting XP on reconnect still works properly with this change.

### DIFF
--- a/game/scripts/vscripts/components/heroselection/ardm.lua
+++ b/game/scripts/vscripts/components/heroselection/ardm.lua
@@ -50,19 +50,6 @@ function ARDMMode:Init (allHeroes)
   self:ReloadHeroPool(DOTA_TEAM_BADGUYS)
 end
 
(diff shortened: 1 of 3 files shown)


chrisinajar:
ardm.lua, line 34 and the method at 56. This change is mutually exclusive with that code, so remove the code there. It'll make sense when you see it, i did the same thing but as a super shitty instance-level shim.

chrisinajar:
PS i think there should be a hardcoded flag for when it's XP from a method as opposed to from "dota"

Trildar:
I assume you meant to have a flag in the XP filter data table. One has been added.

P.S. `GetCurrentXP` is totally a thing. :P

PR #1942, opened 2018-02-04
Balance

l34Um1:


Haganeko:
On meteor hammer, why not double level 1 as well?

l34Um1:
actually yeah good point.

Haganeko:
You redid the nerf you undid, in addition to the new one.
Also, you sure we should alter vanilla values?

l34Um1:
You know I don't like altering vanilla values but Naga is incredibly oppressive. I am also aware I redid the nerf. She would be overpowered with vanilla level 3 ultimate. I honestly expect her to still be very strong, just not stupidly broken.

Haganeko:
Ah, okay. Not refreshing the Core for duels would also reduce her power level, no?

l34Um1:
Yes but she would still be strong. Naga is super op atm.

PR #1957, opened 2018-02-06
They made a good point

l34Um1:
And I am a doofus for not seeing it earlier.

Nerfed Refresher Core.


Haganeko:
What was the point made? Please share.

Melongod:
I pointed out that long cooldowns on huge spells are completely pointless if the refresher core cooldown is shorter than them.

Haganeko:
Melon, I love you.

Akkaren:
To be fair, in original dota reincarnation and blackhole got a higher CD at lvl 1 then refresher. But yes, i believe these are the only ones.

PR #1959, opened 2018-02-07
Fix Scan triggering on bosses

### LINKED ISSUE #1832: Make Scan not trigger on Bosses
Currently, if you Scan a location with a Boss then the Scan shows red, meaning there is an enemy hero there. That shouldn't happen, Scan should only show as red when there actually is a hero.

Trildar:
Turns out Scan isn't actually very hard to implement in Lua.

Closes #1832.


Haganeko:
Where in the code does it show that green/red circle on the map? Trying to learn to read these better.

SquawkyArctangent:
It's lines 38 and 41.

Haganeko:
Oh duh. Managed to skip those lines somehow.

PR #1972, opened 2018-02-10
make aether lens scale better

OTAJumperjay:
Update tooltip_lycan_boss.txt - can i somehow remove this change ? i want to do the tooltips in a seperate pr... i used this to open a new branch ... could someone also tell me how to open new branches without having to change a file on the github server.. i'm sure there is a better way to open one

### DIFF
--- a/game/scripts/npc/items/item_aether_lens.txt
+++ b/game/scripts/npc/items/item_aether_lens.txt
@@ -61,7 +61,7 @@
       "03"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "cast_range_bonus"                                "250 375 400 575 700"
+        "cast_range_bonus"                                "250 325 425 550 700"
       }
     }
   }
--- a/game/scripts/npc/items/item_aether_lens_2.txt
+++ b/game/scripts/npc/items/item_aether_lens_2.txt
@@ -66,7 +66,7 @@
       "03"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "cast_range_bonus"                                "250 375 400 575 700"
+        "cast_range_bonus"                                "250 325 425 550 700"
       }
     }
   }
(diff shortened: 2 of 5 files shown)


OTAJumperjay:
stupid me okay i just found the big AF "new" button ... and i do work locally i just didn't know how to open a different branch lul

Haganeko:
Should also take a look at the shapeshift tooltip. It's pretty bad.
Cast range should be an integer. Try
250/325/400/550/700

Haganeko:
Ah. Switch to master before creating new branches.

OTAJumperjay:
master is horribly outdated... how can i update my master ?

OTAJumperjay:
shapeshift doesn't have a cast range.. what are you talking about

OTAJumperjay:
also i changed the cast range from INTEGER to FLOAT .. that should do it

Haganeko:
That's not the solution. First, the item shouldn't scale perfectly linearly. Later levels can and should be worth more.
Second, float cast range might do weird things.

OTAJumperjay:
okay well didn't work anyway ... i'll do your suggestion then i guess

OTAJumperjay:
250/325/400/550/700 that is not good either i think.. like this it's 250 + 75 + 175 + 150 + 150

l34Um1:
Items shouldn't scale 100% linearly. I don't see much of an issue with the current scaling. Yes it's a great value pickup and doesn't need to be maxed out right away but if it wasn't I don't think it would see play.

OTAJumperjay:
250 375 400 575 700... that is the current one from lvl 2 to lvl 3 it is +25

OTAJumperjay:
then +175 to lvl 4 and +125 to lvl 4

Haganeko:
Jumperjay
250/325/400/550/700 is
+75/+75/+150/+150
In fact, 250/325/400/525/700 might be even better.

OTAJumperjay:
oh ye .. i guess.. what do you think @l34Um1

Haganeko:
250/325/425/550/700
+75/+100/+125/+150

l34Um1:
Sure, let's go with that last one. Current early scaling is actually a typo, my b

OTAJumperjay:
oh i need to update the other aether lens files as well .. guess i learn every day

PR #1983, opened 2018-02-12
Remove Bloodstone recipe to make it like vanilla

MasterGordon:


Haganeko:
Need to correct the cost of all Bloodstones now.

PR #1987, opened 2018-02-13
Fixed pl juxtapose tooltip

Melongod:
Fixed juxtapose having 6 levels of illusion_damage_out_pct and tooltip_illusion_damage


chrisinajar:
This isn't just a tooltip problem, the numbers of lvl 4 and 5 were both wrong.

PR #1995, opened 2018-02-16
Scepter upgrade talents on Panorama

carlosrpg:


Melongod:
Works. 10/10

chrisinajar:
At some point this should be changed to use a nettable instead of an event so that it works automatically upon reconnect. It works well enough to add in now, great work :+1:

PR #1996, opened 2018-02-16
Fix Enchantress Enchant Affects Ancients Talent not working

Trildar:
I was actually supposed to put this fix in a week ago, but apparently I forgot to actually make the PR...


Melongod:
Works.   10/10

PR #1998, opened 2018-02-18
SimpleBoss model

carlosrpg:
Found while messing with Siltbraker models


chrisinajar:
Did you intend to add this to all the simple bosses or just the geostrike one? There's also cleave and fury swipes (although fury swipes is tier 2 if you wanted to take that into consideration)

PR #2000, opened 2018-02-18
10v10 redesign

warpdragon:
Map's ready.


warpdragon:
CHECK MY TRIPS!
get

warpdragon:
CHECK MY TRIPS! #2000
get

PR #2002, opened 2018-02-20
Delay initial cave spawn until after first duel

chrisinajar:


Melongod:
Why was this committed now? This should be post tournament.

l34Um1:
no this should be now

Issue #2004, opened 2018-02-20
Make Final Duels not spawn more normal Duel timers on non-victory


Trildar:
I'm fairly certain this is what causes there to be Duel spam whenever a Final Duel isn't won by the triggering team.

chrisinajar:
Yes, it is.

PR #2005, opened 2018-02-20
Fix final duel and disconnect/reconnect behavior

chrisinajar:


Haganeko:
Does this fix #2004?

Trildar:
I think it does close #2004. Though not really in the way I was thinking of, since this restarts the Duel timer at the end of any Duel as opposed to having the normal Duel timer independent of Final Duels.

Haganeko:
I guess this helps prevent duel spam...?

game/scripts/vscripts/components/duels/duels.lua
```diff
@@ -95,10 +95,11 @@ function Duels:Init ()
         -- player is not in a duel, they can just chill tf out
         return
       end
-      player.disconnected = false
       if player.killed or not player.disconnected then
         return
       end
+      player.disconnected = false
```
Trildar: What is this `player.disconnected` even for? I doubt it's persistent since I think player instances are pretty much just deleted when players disconnect.

game/scripts/vscripts/components/player/connection.lua
```diff
@@ -5,6 +5,9 @@ if PlayerConnection == nil then
 end
 
 function PlayerConnection:Init()
+  GameEvents:OnPlayerDisconnect(function(keys)
+    PauseGame(true)
```
Trildar: Wouldn't having auto-pause active in all modes, all the time annoy people?

chrisinajar: Probably, yes. I should check if it's CM.

game/scripts/vscripts/components/duels/duels.lua
chrisinajar: The player variable here is a reference to the player object within the currentDuel. It uses the disconnected vs killed flag to tell the difference between a player that died before/after disconnect to keep track of the score correctly. There was an error in the code that always counted reconnecting players as having not had disconnected so it wasn't un-counting and unfreezing them (also I had forgotten to add the unfreeze code).

Trildar: Oh wow. It's not a player instance. I saw `Duels:AllPlayers` returning a player instance in one branch and didn't look further than that. This code confuses me so hard.

chrisinajar: It's bad code. One of the first features written, there's a ticket to refactor it open for like 8+ months xD

PR #2007, opened 2018-02-20
Deleting old items

GymLeaderGiovanni:
Removed files for items no longer in the game or items that wont be added. Was advised by Haga on what to remove.


Haganeko:
As is widely known, nothing has ever gone wrong under my watch.

PR #2012, opened 2018-02-20
Removing penalties

Melongod:
as per request by baumi.


Haganeko:
Uh... Why? "Baumi said so", yeah, but why.

warpdragon:
cuz

Trildar:
When safe to leave lands, it might be worth testing if the `Gamerules:SetSafeToLeave` function actually affects the timeout leaver penalty.

PR #2014, opened 2018-02-21
Update Panorama Typescript declaration files

Trildar:
I wonder if there's some fancy git wizardry that would let this stuff be linked to a fork of the ModDota repo containing these files rather than maintaining it completely decoupled from its source.

### DIFF
--- a/content/panorama/dota.d.ts
+++ b/content/panorama/dota.d.ts
@@ -19,2304 +19,2317 @@ type particleID = number;
 type scheduleID = number;
 
 interface CDOTA_PanoramaScript_GameEvents {
-    /**
-     * Subscribe to a game event
-     */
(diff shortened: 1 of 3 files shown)


chrisinajar:
Yes, yes there is, although the moddota repo is wildly out of date and should be completely ignored.

Lets figure out some scripts to generate this, perhaps it's time i have CI start committing to the repo (this and translations)

Trildar:
Buh? Wildly out of date? Are we talking about the same repo here? Pretty much this entire commit was just updating for parity with the repo. And changing the indentation to 2 spaces since the original ones use 4 spaces.
https://github.com/moddota/api

chrisinajar:
you're right i was thinking dotabbuff

PR #2016, opened 2018-02-22
Sohei models (Warp/Carlos)

warpdragon:
Got some pants using FollowEntity(). Walking animation is bugged for some reason.


Haganeko:
Couldn't reuse the wearables library from Frostivus?

warpdragon:
We're going to need to do that eventually, but i'm tired. Can someone look at that walking animation that'd be so fantastic

warpdragon:
Someone should definitely add the CosmeticLib into our repo. Without this  I am unable to do anything outside of Juggernaut-only cosmetics.

warpdragon:
https://youtu.be/3580gaK6ZK4

Haganeko:
Wouldn't a staff look better than a polearm, especially with an overhead swing?

warpdragon:
meh. i asked baumi and he wanted this. the good news is we can change it to anything we want. it could be a giant lolly

OTAJumperjay:
🤔 could be worth thinking about for an april fools update

l34Um1:
I don't mind a staff. Actually someone suggested something like this and I think it would be cool: http://www.nihonzashi.com/bokken_shinai_pictures/purpleheartwoodtetsubo.jpg

FelPac:
An inspiration Model I have made with blender
https://imgur.com/a/Yldpu

warpdragon:
Sohei is ready to roll!

carlosrpg:
I approve this message!

PR #2020, opened 2018-02-22
nullifier custom icons

imaGecko:
nullifier didn't use custom icons


chrisinajar:
The images already exist?

PR #2025, opened 2018-02-23
hugeass sohei refactor + recode

SquawkyArctangent:
prepare for conflicts

- Refactored basically the entire code, including organizing ability
code into separate files.
- Dash and Guard should no longer be castable when they shouldn't; they
were previously, but they tried to fake not being casted by refunding
cooldown and mana cost, but stuff like Nether Ward and Magic Wand would
still react to them.
- Guard now blocks more than one spell, no longer makes Sohei
unkillable, and Sohei will now properly avoid ranged attack damage
entirely instead of healing it back ( previously ranged attack damage
would stick during Ice Blast / Covfefe due to the method used )
- Modifiers that move units ( Dash, Guard talent knockback, Momentum
knockback ) are now proper motion controllers and thus will interact
properly with other motion controllers.
- Momentum now checks for unpathable terrain instead of just terrain
that's of a different height.
- Momentum attacks will always do both the critical damage and the
knockback, instead of sometimes only performing the knockback in rare
situations.
- Momentum now has cooldown support; it will neither build charges while
on cooldown nor can it proc while on cooldown. This means
trigger_distance can be set to 0 and it'll operate strictly on its
cooldown, while also allowing for its cooldown and trigger_distance to
operate in conjuction.
- Dash now has a "momentum_cd_refund" AbilitySpecial, which will cause
value% of momentum's current cooldown to be immediately refunded when
Dash is cast.
- Flurry of Blows is now a channel.
- New Aghs effect: Adds a new ability, Palm of Life, which uses a Dash
charge to dash directly to an ally at 600 range. If Momentum is ready
upon cast, it is discharged to heal the target for 10% of Sohei's
current health when he reaches them.


warpdragon:
daaaaaaaaaaaaaaaaaaaaaaaaaaaaamn

game/scripts/vscripts/abilities/sohei/sohei_momentum.lua
```diff
@@ -0,0 +1,410 @@
...
+		end
+
+		local spell = self:GetAbility()
+
+		if nonHeroTarget then
+			self:SlowAndStun( parent, caster, spell )
+			self:SlowAndStun( nonHeroTarget, caster, spell )
+			self:Destroy()
+		-- why do these mean two different things
+		elseif not GridNav:IsTraversable( tickOrigin ) or GridNav:IsBlocked( tickOrigin ) then
```
chrisinajar: This made heroes instantly collide with themselves (but creeps don't, only heroes), but i didn't test it with the modifier making them MODIFIER_STATE_NO_UNIT_COLLISION... Definitely worth testing extra, if it doesn't work checking Z diff and IsBlocked should probably work.

game/scripts/vscripts/abilities/sohei/sohei_palm_of_life.lua
```diff
@@ -0,0 +1,233 @@
...
+function sohei_palm_of_life:OnHeroCalculateStatBonus()
+	local caster = self:GetCaster()
+
+	if caster:HasScepter() then
+		self:SetHidden( false )
+		if self:GetLevel() <= 0 then
+			self:SetLevel( 1 )
+		end
+	else
+		self:SetHidden( true )
```
chrisinajar: Shouldn't you also unlevel it? Just in case, also makes it copypastable to a generic baseclass

game/scripts/vscripts/abilities/sohei/sohei_palm_of_life.lua
```diff
@@ -0,0 +1,233 @@
...
+if IsServer() then
+	function sohei_palm_of_life:OnSpellStart()
+		local caster = self:GetCaster()
+		local modifier_charges = caster:FindModifierByName( "modifier_sohei_dash_charges" )
+
+		if modifier_charges and not modifier_charges:IsNull() then
+			-- Perform the dash if there is at least one charge remaining
+			if modifier_charges:GetStackCount() >= 1 then
+				modifier_charges:SetStackCount( modifier_charges:GetStackCount() - 1 )
+			end
```
chrisinajar: else refund mana and return

game/scripts/vscripts/abilities/sohei/sohei_palm_of_life.lua
```diff
@@ -0,0 +1,233 @@
...
+		local modMomentum = caster:FindModifierByName( "modifier_sohei_momentum_passive" )
+		local spellMomentum = caster:FindAbilityByName( "sohei_momentum" )
+
+		if ( modMomentum and modMomentum:IsMomentumReady() ) and ( spellMomentum and spellMomentum:IsCooldownReady() ) then
+			doHeal = 1
+		end
+
+		caster:RemoveModifierByName( "modifier_sohei_palm_of_life_movement" )
+		caster:RemoveModifierByName( "modifier_sohei_dash_movement" )
+		caster:EmitSound( "DOTA_Item.ForceStaff.Activate" )
```
chrisinajar: ForceStaff sound can be removed

game/scripts/vscripts/abilities/sohei/sohei_palm_of_life.lua
```diff
@@ -0,0 +1,233 @@
...
+		parent:SetAbsOrigin( tickOrigin )
+		parent:FaceTowards( targetOrigin )
+
+		GridNav:DestroyTreesAroundPoint( tickOrigin, self.tree_radius, false )
+
+		local distance = parent:GetRangeToUnit( self.target )
+
+		if distance <= self.endDistance then
+			if self.doHeal then
+				-- do the heal
```
chrisinajar: Can you throw a sound event for the heal in there too, and we'll find one to fit it

game/scripts/vscripts/abilities/sohei/sohei_momentum.lua
SquawkyArctangent: Works beautifully from my tests. Might've been also caused by the fact that the modifier wasn't a motion controller, so it caused nonsense in terms of blocking.

game/scripts/vscripts/abilities/sohei/sohei_palm_of_life.lua
SquawkyArctangent: This seems pretty unnecessary, as vanilla Aghs added abilities actually start at level 1 no matter what. This is really more of a quick-and-dirty way of doing it as opposed to just starting the ability already leveled as soon as the hero spawns in.

game/scripts/vscripts/abilities/sohei/sohei_guard.lua
```diff
@@ -0,0 +1,370 @@
...
+		local caster = self:GetCaster()
+		local target = self:GetCursorTarget() or caster
+
+		-- Check if there are enough charges to cast the ability, if the caster is stunned
+
+		if caster:IsStunned() then
+			local modifier_charges = caster:FindModifierByName( "modifier_sohei_dash_charges" )
+
+			if modifier_charges then
+				modifier_charges:SetStackCount( modifier_charges:GetStackCount() - 2 )
```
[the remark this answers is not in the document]
SquawkyArctangent: Like I said in the comments ( scattered though they may be ), OnSpellStart is really much too late to be doing these types checks. Too much stuff has already happened at this point in the logic chain to say no to the ability happening, so imo it's really better to just let the ability fire off if the spell somehow reached this point even if the conditions aren't right to avoid stuff like what happened with the previous Dash talent bug.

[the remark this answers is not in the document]
chrisinajar: Oh i see, i didn't notice the code to put the two abilities on cooldown.

Do you think implementing RefreshCharges on the ability will make refresher work with it? I know it'll make it work with duels and our custom refreshers at least, we call that method...

game/scripts/vscripts/abilities/sohei/sohei_guard.lua
```diff
@@ -0,0 +1,370 @@
...
+		local modifier_charges = caster:FindModifierByName( "modifier_sohei_dash_charges" )
+		local talent = caster:FindAbilityByName( "special_bonus_sohei_guard_allycast" )
+
+		-- unfinished talent stuff
+		if talent and talent:GetLevel() > 0 then
+			behavior = bit.bor( DOTA_ABILITY_BEHAVIOR_UNIT_TARGET, DOTA_ABILITY_BEHAVIOR_IMMEDIATE, DOTA_ABILITY_BEHAVIOR_DONT_CANCEL_CHANNEL )
+		end
+
+		if modifier_charges and modifier_charges:GetStackCount() >= 2 then
+			behavior = bit.bor( behavior, DOTA_ABILITY_BEHAVIOR_IGNORE_PSEUDO_QUEUE )
```
chrisinajar: OOHHHH i was wondering how you got the 2 charges thing to work.... Nice.

game/scripts/vscripts/abilities/sohei/sohei_guard.lua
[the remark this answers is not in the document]
SquawkyArctangent: Looks like it doesn't work with vanilla orb but works with everything else, though the code I set up didn't accommodate for other sources of charge gaining so I'm redoing that part of the code a bit too.

PR #2027, opened 2018-02-23
Duel refactor stage 1: modularization, some cleanup, teleport fixes

Trildar:
Split up duels.lua into more modules and refactor some of the code. Notable changes off the top of my head include:
* Storing Duel zones into an array and using loops to avoid repetition
* Store item cooldowns with the handles as keys and the cooldowns as values and loop over them to restore (cause that function chain thing was seriously hard to understand)
* Rewrite summon finding code, in SafeTeleportAll, in functional style (tried to use GetAdditionalOwnedUnits as well, but it's not usable in this case due to the reasons stated in the comment)

I was going to do a major overhaul of the Duel system at the same time, but figured I should probably stagger the changes a little. Goodness knows I ran into 99 googol `nil` errors just making these changes. Oh how I long for static analysis.


Trildar:
Ah, yes. Also changed the teleport retries in `SafeTeleport` to not add 100 more max distance every retry. I just know that's going to end up disagreeing with something... as much as I wished it didn't.

chrisinajar:
Will merge in ~20 hours. Test well and make sure final duels still work.

PR #2028, opened 2018-02-26
Sohei lint fixes

Trildar:
You can shut up now, Travis.

PR #2029, opened 2018-02-26
Sohei abilities warnings fix attempt

zelding:
- removed whitespace (tab)
- changed target resolving to circumvent variable shadowing


zelding:
fixed already in #2028

PR #2031, opened 2018-02-28
Sohei model animation and particles fixes

carlosrpg:


game/scripts/vscripts/components/duels/duels.lua
```diff
@@ -722,6 +722,7 @@ function Duels:SavePlayerState (hero)
 end
 
 function Duels:RestorePlayerState (hero, state)
+  hero:Interrupt(  )
```
chrisinajar: Should we Stop, instead, to prevent queued commands from running?

PR #2038, opened 2018-03-02
Puck invulnerability

GymLeaderGiovanni:
Phase shift gave potentential invulnerability. Reduced duration to prevent this.


GymLeaderGiovanni:
do i need to fix all of the travis errors it looks like they are all because of the new update except for puck phase shift which is what I changed so it doesn't match vanilla but also doesn't give the potential for permanent phase shift.

l34Um1:
Just want to make clear that this should be merged before the tournament. We don't want Phase Shift abuse.

PR #2040, opened 2018-03-02
Fix Valve's 7.10 update. And fix of Puck's Wanning Rift damage and Disruptor's Thunder Strike damage

Anonymous331297:


imaGecko:
Not sure why it shows travis as pending, it did run

game/scripts/npc/abilities/lone_druid_spirit_bear.txt
```diff
@@ -37,7 +37,7 @@
       "02"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "bear_regen_tooltip"                              "4 5 6 7 11 15"
+        "bear_regen_tooltip"                              "5 6 7 8 11 15" //OAA dota is wrong
```
chrisinajar: What?

Haganeko:
@chrisinajar for some of the abilities (kunkka) you changed the level 5/6 increases in proportion with the change in the increases on 1/2/3/4 (base increase went from 20 to 25 so you made epic increase from 60 to 75) while on others (starfall) you kept it the same. Why?

chrisinajar:
Sometimes it warranted it and sometimes it didn't. I don't usually change late game scaling when the early game was nerfed without changing the last level much, since by our equations we should *buff* the ability which was nerfed, and that seems weird to me. A lot of these changes I'm called "early game nerf, same late game."

Haganeko:
Ugh, this is going to disrupt the whole meta

PR #2042, opened 2018-03-02
Fix Massive lag found during stream

carlosrpg:


Haganeko:
I assume this reduces the number of calls?

Trildar:
I'm fairly certain this has never been a problem before. How do you know this was causing lag?

carlosrpg:
https://clips.twitch.tv/CheerfulNeighborlyButterFailFish

Trildar:
I'm not sure those errors are actually related to the lag. Given the other things in the console after it, it seems like it happened a while ago. Though it could be that it just stopped printing them since that tends to happen with those types of errors.

There's also the matter that this error would only come up after upgrading GGG or destroying it. And if it's happening continuously, it's likely a sign of the modifier not getting removed somehow, which would be a much bigger issue.

Melongod:
There has been way more lag spikes lately.

chrisinajar:
This wont fix the client side exception that didn't cause the lag.

PR #2043, opened 2018-03-02
Fix Map Exploits

warpdragon:
adjusted a bounding box to prevent possible duel exploits.


warpdragon:
@chrisinajar this should go in b4 tourney

PR #2046, opened 2018-03-05
bottle statue

warpdragon:
bottle statue


warpdragon:
Bottle statue complete with particles.
Panel complete
Opens with right click instead of left click currently.

chrisinajar:
I don't think standins should be on the status. They don't get any of the other benefits from winning, and then even if they are included then they should be listed last.

Also did we double check with the players how they'd like their names? I assume tournament tags are fine, but worth checking.

warpdragon:
fixed.

PR #2047, opened 2018-03-05
Rework networth ranking code for hero kills

Trildar:
This method should remove what appeared to be a race condition with networth changing in the middle of the kill gold function.


Haganeko:
Where do I file the "does not delete GPM boots" complaint?

Trildar:
Shh... That needs to be hidden amongst 20 other commits that change 50 files.

Haganeko:
Oh. You mean when I redo gold again again again. Sure! No one will expect it from me.

PR #2051, opened 2018-03-08
sohei tooltip fixes

OTAJumperjay:


chrisinajar:
+"DOTA_Tooltip_Ability_sohei_flurry_of_blows_description"                        "CHANNELED - Sohei becomes invulnerable, randomly attacking one target at a time in a selected area with the intention of pushing them back into the epicenter. Each hit triggers Momentum. Ends after either the duration or the maximum amount of attacks have run out."

still could be worded a lot better, but this is still a strict improvement in most ways

PR #2061, opened 2018-03-11
Make Zuus nimbus not affected by CDR WIP

OTAJumperjay:
Share bug bounty 50:50 with carlos_rpg <3 he helped a lot

### DIFF
--- a/game/scripts/npc/abilities/zuus_cloud.txt
+++ b/game/scripts/npc/abilities/zuus_cloud.txt
@@ -8,6 +8,9 @@
     "AbilityBehavior"                                     "DOTA_ABILITY_BEHAVIOR_POINT | DOTA_ABILITY_BEHAVIOR_AOE | DOTA_ABILITY_BEHAVIOR_HIDDEN | DOTA_ABILITY_BEHAVIOR_SHOW_IN_GUIDES"
     "AbilityUnitDamageType"                               "DAMAGE_TYPE_MAGICAL"
     "SpellImmunityType"                                   "SPELL_IMMUNITY_ENEMIES_NO"
+    "BaseClass"                                           "ability_lua"
+    "ScriptFile"                                          "abilities/oaa_zuus_cloud.lua"
+    "AbilityTextureName"                                  "zuus_cloud"
     "MaxLevel"                                            "1"
     "FightRecapLevel"                                     "1"
     "IsGrantedByScepter"                                  "1"
@@ -52,4 +55,3 @@
       }
     }
   }
-}
--- a/game/scripts/vscripts/abilities/oaa_zuus_cloud.lua
+++ b/game/scripts/vscripts/abilities/oaa_zuus_cloud.lua
@@ -60,7 +60,7 @@ function modifier_zuus_cloud_oaa:OnIntervalThink()
       self.Radius,
       DOTA_UNIT_TARGET_TEAM_ENEMY,
       DOTA_UNIT_TARGET_HERO,
-      bit.bor( DOTA_UNIT_TARGET_FLAG_NOT_MAGIC_IMMUNE_ALLIES ),
+      DOTA_UNIT_TARGET_FLAG_NONE,
       FIND_CLOSEST,
       false
     )
@@ -72,7 +72,7 @@ function modifier_zuus_cloud_oaa:OnIntervalThink()
         self.Radius,
         DOTA_UNIT_TARGET_TEAM_ENEMY,
         DOTA_UNIT_TARGET_CREEP,
-        bit.bor( DOTA_UNIT_TARGET_FLAG_NOT_MAGIC_IMMUNE_ALLIES ),
+        DOTA_UNIT_TARGET_FLAG_NONE,
         FIND_CLOSEST,
         false
       )
--- a/game/scripts/vscripts/abilities/oaa_zuus_cloud.lua
+++ b/game/scripts/vscripts/abilities/oaa_zuus_cloud.lua
@@ -107,7 +107,7 @@ function modifier_zuus_cloud_oaa:CastLightningBolt(target)
     -- Renders the particle on the sigil
     local particle = ParticleManager:CreateParticle("particles/units/heroes/hero_zeus/zeus_cloud_strike.vpcf", PATTACH_POINT_FOLLOW, self:GetParent())
 
-    ParticleManager:SetParticleControlEnt(self.cloud_particle, 0, parent, PATTACH_POINT_FOLLOW, "attach_hitloc", parent:GetAbsOrigin(), true)
+    ParticleManager:SetParticleControlEnt(particle, 0, parent, PATTACH_POINT_FOLLOW, "attach_hitloc", parent:GetAbsOrigin(), true)
     ParticleManager:SetParticleControl(particle, 1, Vector(target:GetAbsOrigin().x,target:GetAbsOrigin().y,target:GetAbsOrigin().z + target:GetBoundingMaxs().z ))
     EmitSoundOn("Hero_Zuus.LightningBolt.Cloud", target)
   end
--- a/game/scripts/vscripts/abilities/oaa_zuus_cloud.lua
+++ b/game/scripts/vscripts/abilities/oaa_zuus_cloud.lua
@@ -6,7 +6,7 @@ function zuus_cloud:OnSpellStart()
   local hCloud = CreateUnitByName( "npc_dota_zeus_cloud", self:GetCursorPosition(), true, caster, caster, caster:GetTeamNumber() )
   hCloud:SetOwner( self:GetCaster() )
   hCloud:SetControllableByPlayer( self:GetCaster():GetPlayerOwnerID(), false )
-  hCloud:AddNewModifier( caster, self, "modifier_zuus_cloud_oaa", { Interval = self:GetSpecialValueFor( "cloud_bolt_interval" ), Radius = self:GetSpecialValueFor( "cloud_radius" ) } )
+  hCloud:AddNewModifier( caster, self, "modifier_zuus_cloud_oaa", nil)
   hCloud:AddNewModifier( caster, self, "modifier_kill", { duration = self:GetSpecialValueFor( "cloud_duration" ) } )
   FindClearSpaceForUnit( hCloud, self:GetCursorPosition(), true )
 end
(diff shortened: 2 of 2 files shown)


game/scripts/npc/abilities/zuus_cloud.txt
```diff
@@ -8,6 +8,9 @@
     "AbilityBehavior"                                     "DOTA_ABILITY_BEHAVIOR_POINT | DOTA_ABILITY_BEHAVIOR_AOE | DOTA_ABILITY_BEHAVIOR_HIDDEN | DOTA_ABILITY_BEHAVIOR_SHOW_IN_GUIDES"
     "AbilityUnitDamageType"                               "DAMAGE_TYPE_MAGICAL"
     "SpellImmunityType"                                   "SPELL_IMMUNITY_ENEMIES_NO"
+    "BaseClass"                                           "ability_lua"
+    "ScriptFile"                                          "abilities/oaa_zuus_cloud.lua"
```
Trildar: Create a completely new ability called `zuus_cloud_oaa` and change Zeus' ability KV. Overwriting default abilities is bad practice as it can sometimes still have hardcoded behaviour. Change the tooltip to have a note about not being affected by CDR while you're at it, since you'll need to also copy the tooltip.

The naming convention for the Lua file is usually to not have the hero name, but maybe this should be an exception since "oaa_cloud" is not really a great name.

game/scripts/vscripts/abilities/oaa_zuus_cloud.lua
```diff
@@ -0,0 +1,114 @@
...
+LinkLuaModifier( "modifier_zuus_cloud_oaa", "abilities/oaa_zuus_cloud.lua", LUA_MODIFIER_MOTION_NONE )
+
+function zuus_cloud:OnSpellStart()
+  local caster = self:GetCaster()
+  local hCloud = CreateUnitByName( "npc_dota_zeus_cloud", self:GetCursorPosition(), true, caster, caster, caster:GetTeamNumber() )
+  hCloud:SetOwner( self:GetCaster() )
+  hCloud:SetControllableByPlayer( self:GetCaster():GetPlayerOwnerID(), false )
+  hCloud:AddNewModifier( caster, self, "modifier_zuus_cloud_oaa", { Interval = self:GetSpecialValueFor( "cloud_bolt_interval" ), Radius = self:GetSpecialValueFor( "cloud_radius" ) } )
+  hCloud:AddNewModifier( caster, self, "modifier_kill", { duration = self:GetSpecialValueFor( "cloud_duration" ) } )
+  FindClearSpaceForUnit( hCloud, self:GetCursorPosition(), true )
```
Trildar: I don't think this is really necessary since the Nimbus should be flying and have no collision. On that note though, you should check if it has proper behaviour when placed over trees and that you can walk under it.

game/scripts/vscripts/abilities/oaa_zuus_cloud.lua
```diff
@@ -0,0 +1,114 @@
...
+end
+
+function modifier_zuus_cloud_oaa:RemoveOnDeath()
+  return true
+end
+
+function modifier_zuus_cloud_oaa:OnCreated( kv )
+  if IsServer() then
+    self.Interval = kv.Interval
+    self.Radius = kv.Radius
```
Trildar: There isn't really any reason to have these values passed in on creation rather than just being read from the ability here.

game/scripts/vscripts/abilities/oaa_zuus_cloud.lua
```diff
@@ -0,0 +1,114 @@
...
+    self:StartIntervalThink( 0.1 )
+    EmitSoundOn("Hero_Zuus.Cloud.Cast", self:GetParent())
+  end
+end
+
+function modifier_zuus_cloud_oaa:OnDestroy()
+  if IsServer() then
+    ParticleManager:DestroyParticle( self.cloud_particle, false)
+    ParticleManager:ReleaseParticleIndex(self.cloud_particle)
+    self.cloud_particle= nil
```
Trildar: Missing space.

game/scripts/vscripts/abilities/oaa_zuus_cloud.lua
```diff
@@ -0,0 +1,114 @@
...
+
+    ParticleManager:SetParticleControlEnt(self.cloud_particle, 2, self:GetParent(), PATTACH_POINT_FOLLOW, "attach_hitloc", self:GetParent():GetAbsOrigin(), true)
+    self:StartIntervalThink( 0.1 )
+    EmitSoundOn("Hero_Zuus.Cloud.Cast", self:GetParent())
+  end
+end
+
+function modifier_zuus_cloud_oaa:OnDestroy()
+  if IsServer() then
+    ParticleManager:DestroyParticle( self.cloud_particle, false)
```
Trildar: I would ask that either everything uses spaces between the brackets or doesn't, but that's probably too nitpicky. I would like it to at least be symmetrical in a single function call though.

game/scripts/vscripts/abilities/oaa_zuus_cloud.lua
```diff
@@ -0,0 +1,114 @@
...
+      self:CastLightningBolt(targets[1])
+      self.LastStrike = GameRules:GetDOTATime(false, false)
+    end
+  end
+end
+
+function modifier_zuus_cloud_oaa:CastLightningBolt(target)
+  local caster = self:GetCaster()
+  local parent = self:GetParent()
+  local ability = caster:FindAbilityByName( 'zuus_lightning_bolt' )
```
Trildar: `ability` in a context like this typically refers to the modifier's parent ability, so it might be a good idea to name it something else.

game/scripts/vscripts/abilities/oaa_zuus_cloud.lua
```diff
@@ -0,0 +1,114 @@
...
+  local radius = ability:GetSpecialValueFor("spread_aoe")
```
Trildar: This looks like it's unused.

game/scripts/vscripts/abilities/oaa_zuus_cloud.lua
```diff
@@ -0,0 +1,114 @@
...
+  local sight_radius =  0
+  if GameRules:IsDaytime() then
+    sight_radius = ability:GetSpecialValueFor("sight_radius_day")
+  else
+    sight_radius = ability:GetSpecialValueFor("sight_radius_night")
+  end
+  local sight_duration = ability:GetSpecialValueFor("sight_duration")
+
+  if ability:GetLevel() > 0 then
+    AddFOWViewer(caster:GetTeam(), target:GetAbsOrigin(), sight_radius, sight_duration, false)
```
Trildar: Need to add true sight.

game/scripts/vscripts/abilities/oaa_zuus_cloud.lua
```diff
@@ -0,0 +1,114 @@
...
+
+    -- Checks if the target has been set yet
+    target:AddNewModifier(caster, ability, "modifier_stunned", {Duration = 0.1})
```
Trildar: Missing code for ministun Talent.

game/scripts/vscripts/abilities/oaa_zuus_cloud.lua
```diff
@@ -0,0 +1,114 @@
...
+    ApplyDamage({victim = target, attacker = caster, damage = ability:GetAbilityDamage(), damage_type = ability:GetAbilityDamageType()})
```
Trildar: Vanilla Nimbus actually sources the damage to the Nimbus, so the attacker should be the parent rather than the caster.

game/scripts/vscripts/abilities/oaa_zuus_cloud.lua
carlosrpg: FindClearSpaceForUnit does not change unit colision, it is just to place it.
it has no effect.
It does work on trees and you can walk pass it

OTAJumperjay: It has proper behavior. I can place it over trees and the trees don't get destroyed and i can walk underneath the nimbus

game/scripts/npc/abilities/zuus_cloud.txt
carlosrpg: this is already an overrid of a dota abiltiy, but I got your point.

game/scripts/vscripts/abilities/oaa_zuus_cloud.lua
Trildar: I know it doesn't affect collision, I just meant there isn't really any point to doing this when it shouldn't need "clear space" since it has no collision. And now I realised that actually this should already be taken of by `CreateUnitByName` in the first place.

game/scripts/npc/abilities/zuus_cloud.txt
carlosrpg: I tried to create a new zuus_cloud_oaa but it is not showing when scepter I have scepter.
maybe you could try something but it did not worked for me.
probably some valve hardcoded bullshit

game/scripts/vscripts/abilities/oaa_zuus_cloud.lua
carlosrpg: True sight is working

OTAJumperjay:
actually still need to add the tooltip whoops

carlosrpg:
Can someone else test this as well?
I tested and it is not applying Cooldown Reduction but Jumperlay says it is applying CDR on his branch

carlosrpg:
So, no one is willing to test it? @chrisinajar can you just merge it and we verify if it works on the game itself?

Trildar:
Well, the ability was also supposed to be renamed to avoid vanilla behaviour interference. Shouldn't be hard. The code for unhiding the ability when Agh's is picked up can be copied from Sohei's Palm of Life.

game/scripts/npc/abilities/zuus_cloud.txt
```diff
@@ -52,4 +55,3 @@
       }
     }
   }
-}
```
Trildar: This should not be removed. The file has unbalanced brackets because of this. Fixing the issue seems to make it not affected by CDR. Though it also seems to make the Nimbus die in one hit from a hero. Probably needs code in the modifier to do the hits-to-kill thing.

carlosrpg:
I tried renaming, it makes the aghs not give the ability anymore, could not fix it

PR #2063, opened 2018-03-11
Reposition the discord link

OTAJumperjay:


chrisinajar:
We should really make this dismissable...

PR #2075, opened 2018-03-17
Add debug messages for JavaScript loading order

Trildar:


chrisinajar:
Fixed bug, it was never loading (no load order errors).

PR #2077, opened 2018-03-18
Load Screen Workaround 

carlosrpg:
Trying to minimize errors by loading the FindDotaHudElement function in all Js files that uses it


carlosrpg:
I would like to add, this is a shitty workaround! I just want to having people playing again. It has being a week and I hate seeing people calling the game shit because of that

PR #2079, opened 2018-03-19
Minor Sohei fixes

SquawkyArctangent:
Sohei relevant stuff:
- Momentum can now be broken. It will neither build charges nor be able
to unleash a charged attack. Flurry of Blows ignores this, however.
- Momentum no longer loses its charges when upgrading it.
- Internally increased the distance of Dash by 1, which seems to fix a
precision error leading to it having a distance of 297 instead.
- Flurry of Blows attacks no longer consume Momentum charges.
Rubick relevant stuff:
- Dash no longer gives Rubick a charge beyond the maximum if stolen
while he has max charges.
- Stealing Flurry of Blows will now also steal Momentum, which is now
hidden while stolen and will behave as if it was broken while hidden.
Effectively, Flurry of Blows is now given the Requiem of Souls
treatment.
- Palm of Life now acts independently from Dash and Momentum while
stolen.
- Also added a IsHiddenWhenStolen override to AbilityBaseClass which
calls the ability basest class's IsHiddenWhenStolen without the added
argument. Because for some reason it's called with some sort of table in
C++ but the Lua version hasn't been adapted to that change.

### DIFF
--- a/game/scripts/vscripts/abilities/sohei/sohei_flurry_of_blows.lua
+++ b/game/scripts/vscripts/abilities/sohei/sohei_flurry_of_blows.lua
@@ -33,6 +33,12 @@ end
 
 --------------------------------------------------------------------------------
 
+function sohei_flurry_of_blows:GetAssociatedSecondaryAbilities()
+  return "sohei_momentum"
+end
+
+--------------------------------------------------------------------------------
+
 function sohei_flurry_of_blows:GetChannelTime()
   --[[
   if self:GetCaster():HasScepter() then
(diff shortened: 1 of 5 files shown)


game/scripts/vscripts/abilities/sohei/sohei_flurry_of_blows.lua
```diff
@@ -33,6 +33,12 @@ end
 
 --------------------------------------------------------------------------------
 
+function sohei_flurry_of_blows:GetAssociatedSecondaryAbilities()
```
chrisinajar: Wow.

chrisinajar:
utACK

PR #2129, opened 2018-03-25
Sohei rework bug bounty

Yahnich:
and a free talent system because i honestly hate not having my talent system


chrisinajar:
PS: .luacheckrc

PR #2151, opened 2018-03-31
Capture Points

M-Faraday:
Capture Points Finished


M-Faraday:
The capture points were made for the oaa.test map

l34Um1:
I tested this locally and the capture points don't show up for me and aren't pinged out on the minimap. Frankly none of it works. How would you recommend I test it to get the capture points to show up?

Haganeko:
Perhaps stuff needs to be compiled?

M-Faraday:
It was made for the test map. I also added a command to start capture points manually -capture I believe

PR #2152, opened 2018-03-31
Zones hotfix

warpdragon:
fixed misalignment in both rad/dire offsides zone. fixed hole in dire large arena zone. requires map rebuild.


warpdragon:
updated in another pr

PR #2166, opened 2018-04-06
Make charger spawn less towers closer to him

### LINKED ISSUE #2162: Stop Charger towers from spawning on stairs
Sometimes Chargers towers spawn on the stairs leading into his arena, blocking the path in. This is especially bad in the two center boss arenas.  I'm unsure what the best way to stop this from happening is, maybe just reducing the possible spawn AOE is good enough.

imaGecko:
closes #2162 
Towers in radius from 8 to 4
Spawn radius from 700 - 800 to 500 - 600
Towers can still spawn far away if charger attempts to drop multiple towers on the same location

Also increased the radius to reduce towers being outside of the search radius
This can happen when charger tried to spawn a tower on top of a tower at a cliff pushing the spawned tower far away from the intended spawn location

856f1c341dc1921c6fe9c001c738b76d


carlosrpg:
can we reduce to 6 instead?
only 4 is almost impossible if the team does not kill him before his last stage, and this is now a tier 2 boss, we expect the team to be weaker when facing him

imaGecko:
he does respawn the towers up to 4 after every charge

carlosrpg:
can we bump to 6 instead?
same argument as before,

carlosrpg:
I disagree. it is way more dificult to agro him to a tower when there is less tower when he is in the last stage.
I guess this was discussed on discord right?
I definitely don't agree with this change but if this was discussed I guess I have to accept

chrisinajar:
this makes charger easier :(

chrisinajar:
The thing that makes charger hard is getting him to hit the towers. The reaction of dodging him has always been based on the channel time, not the charge speed, which is crazy fast at higher phases where it matters. When towers are farther away there is less margin for error to get him to hit them and that's harder, when they're close he's bound to hit one if he goes basically anywhere in that direction.

imaGecko:
The issue is not difficulty, the issue is functionality. More towers cause more problems with the arena becoming cluttered and towers being spawned outside of the arena, it already being hard to move around with accidentally clicking on charger espeically with the now slightly smaller (mostly more closed off) arena. Reducing the amount of towers regardless of difficulty is an improvement of the feel. If it is too easy mechanics wise then the charge itself is still open to adjustment.

Haganeko:
Could make the towers smaller, too. As in reduce their hit boxes.

Issue #2161, opened 2018-04-06
Level Cap at 50


l34Um1:
- [ ] Add a level cap at 50
- [ ] Remove the diminishing stats past level 25

PR #2157, opened 2018-04-03
travis

imaGecko:
travis

PR #2167, opened 2018-04-08
Introduce Level Cap

Newton314:
#2161 
Maximum Level is now 50
Removes diminishing stats past lvl 25


imaGecko:
#2157 fixes the errors
errors can occur when base dota changes something and the game isn't updated yet

game/scripts/vscripts/settings.lua
```diff
@@ -133,7 +133,7 @@ END_GAME_ON_KILLS = false               -- Should the game end after a certain n
 KILLS_TO_END_GAME_FOR_TEAM = 50         -- How many kills for a team should signify an end of game?
 
 USE_CUSTOM_HERO_LEVELS = true           -- Should we allow heroes to have custom levels?
-MAX_LEVEL = 5000                        -- What level should we let heroes get to?
+MAX_LEVEL = 50                        -- What level should we let heroes get to?
```
Trildar: Just to check. Have you tested this? I'm not sure if this value is used or if only the XP per level table matters. Also, do align the comments if you don't mind.

And you should probably modify the XP per level table even if this does work as it would have thousands of unnecessary values otherwise.

Newton314: Yes i tried it and i couldn't lvl up past level 50. 
I will take a look at the table.

Newton314: luckely you implemented the "xp_per_level_table" in such a way, that it only calculates the xp up to the value of "MAX_LEVEL".
So by changing the max_level, the size of the table changes automaticaly.

Trildar: Oh? Dur. Maybe I should actually read code that I suggest changes to.

l34Um1:
This works fine. I think it can be merged.

PR #2170, opened 2018-04-13
Remove attribute Status Resistance

carlosrpg:


carlosrpg:
@chrisinajar , if you don't have the time just merge and submit that to make the game playable again

Trildar:
Muncorn on the community Discord said that it's working now. Dunno if maybe Valve did something so that the functions don't just error and break everything.

warpdragon:
was in the middle of a game playing io. about 2/3 the way thru the game, after having already lost one player due to QQ, the game decided to hard-crash everyone from the lobby. upon reconnect there was the "WAIT" pick screen. I think a fix is still in order

Trildar:
That seems unrelated though. This function only runs on setup to set the gain per point of a stat.

imaGecko:
I still has to be removed, as does the intelligence magic resistance
And the strength magic resistance has to be lowered to what the intelligence magic resistance preiously was

imaGecko:
tbh I don't quite know what the best approach for magic resistance per strength would be. Strength heroes getting 0.1 is certainly going to be too high

Issue #2068, opened 2018-03-14
Surrender Option | 15€ Bounty


Haganeko:
Allow a team to vote to surrender once they are fifty kills behind. Four votes in favor are necessary to give up a game. If less than five players are in the team, then reduce the necessary amount of votes accordingly:
```
Players | Votes
   5    |   4
   4    |   3
   3    |   2
   2    |   2
   1    |   1
```

After the threshhold has been reached players can type "-surrender" into chat. If they do an interface pops up that lets them vote to surrender or not to surrender. If a player abstains they count as a "no".
This vote interface stays for up to a minute or until a conclusive vote has been reached. It will then announce it in chat. If a surrender has been voted for, the game will end after a 10s timer and a message "The Radiant/Dire team surrendered" displays for both teams on the top of the screen.

A surrender vote can only be initiated after at least a minuted passed after the last one. A single player can only intitate one surrender vote every 5 minutes.

l34Um1:
You need to be more precise about how this vote would happen. What's the interface for it?

RodneyMcKay:
I'm working on that for IMBA, once it's done and fully working i'll do it in OAA if no-one did it in the meantime.

carlosrpg:
I Claim this

carlosrpg:
I unclaim due to lack of time

M-i-k-e-l:
I claim this

PR #2178, opened 2018-04-22
Add surrender option

M-i-k-e-l:
There are two parts to this:
1. Rename LinkCommand to LinkDevCommand (LinkCommand will be used for in-game user commands)
2. Add the -surrender option ( #2068)


game/scripts/vscripts/components/surrender/surrender.lua
```diff
@@ -0,0 +1,147 @@
...
+function SurrenderManager:CheckSurrenderConditions(keys)
+  local teamId = PlayerResource:GetTeam(keys.playerid)
+  local now = GameRules:GetGameTime()
+  if SurrenderManager:ScoreAllowsSurrender(teamId) and
+      SurrenderManager:TimeAllowsSurrender(keys.playerid, now) then
+    lastTimeSurrenderWasCalled = GameRules:GetGameTime()
+    lastTimeSurrenderWasCalledByPlayer[keys.playerid] = now
+    teamIdToSurrender = teamId
+    local timeout = 10
+    local text = "Would you like to surrender?"
```
Trildar: Send a localisation key like "#surrender_prompt" and do `$.Localize(data.text)` in the JavaScript.

game/scripts/vscripts/components/surrender/surrender.lua
```diff
@@ -0,0 +1,147 @@
...
+    numberOfVotesCast = 0
+    numberOfVotesExpected = 0
+    DebugPrint("requiredNumberOfYesVotes = " .. requiredNumberOfYesVotes)
+    if requiredNumberOfYesVotes <= yesVotesCast then
+      DebugPrint("End game")
+      local teamText = nil
+      if teamIdToSurrender == DOTA_TEAM_GOODGUYS then
+        teamText = "Radiant"
+      elseif teamIdToSurrender == DOTA_TEAM_BADGUYS then
+        teamText = "Dire"
```
Trildar: I feel like the team names shouldn't be hardcoded. Unsure what the best way to handle this is. One way would be to just have completely separate keys. Not sure if the system localises the DialogVariable stuff as well. Could try I suppose. Vanilla Dota already has keys for "Radiant" and "Dire", "DOTA_GoodGuysShort" and "DOTA_BadGuysShort", respectively.

l34Um1:
This needs better text but other than that it seems good to go. Maybe ready to merge? I'm unsure, someone else should review this too.

game/scripts/vscripts/components/surrender/surrender.lua
M-i-k-e-l: I cannot find these constants - I'm thinking they are unavailable in the mod(?)
Should I add these constants? Where do they go?

Trildar: They're in vanilla Dota, they can be used without any need to put them in our localisation files. https://raw.githubusercontent.com/SteamDatabase/GameTracking-Dota2/master/game/dota/resource/dota_english.txt

M-i-k-e-l: I tried using them and it did not work, do I need to declare something (usage\import etc.)?

M-i-k-e-l: Anyone?

Trildar: No. If you put them in the variable part and it didn't work then it just means those don't get run through localisation, which is unfortunate. The other option is just to have keys for surrender text for both teams. Similar to the victory messages in vanilla Dota.

M-i-k-e-l: Is it possible to merge as is and I'll change the text later? - I tried to change it, then run the translation but it had some errors. Maybe I need to rebase, but I'm new to the git environment and not sure I'll have time to research how to do that here - I'm used to having the option in the UI and can't seem to find it in the GitHub Desktop.

carlosrpg:
Is this ready? @Trildar did you tested it? can you guys explain again how should this work?

Trildar:
I didn't test. Only did code review. Would need multiple people to test I think. IIRC, and based on the ticket. It's supposed to allow starting a surrender vote by typing -surrender in chat when you're 50 or more points behind. It brings up a vote panel. Team surrenders if vote has 2/3 majority of still connected players.

Trildar:
I also didn't really check everything in detail. Like I'm now realising there's stuff which should be moved to global constants in settings.lua.

PR #2179, opened 2018-04-25
Fixes for the new Bosses

TideSofDarK:


game/scripts/vscripts/units/swiper.lua
```diff
@@ -89,19 +89,20 @@ function SwiperBossThink()
...
+		local frontSwipeEnemies = FindUnitsInRadius(
 			thisEntity:GetTeamNumber(),
+			thisEntity:GetOrigin(), nil,
+			swipeRange,
 			DOTA_UNIT_TARGET_TEAM_ENEMY,
-			DOTA_UNIT_TARGET_ALL,
-			DOTA_UNIT_TARGET_FLAG_NONE,
-			FIND_CLOSEST)
+			DOTA_UNIT_TARGET_HERO,
+			DOTA_UNIT_TARGET_FLAG_MAGIC_IMMUNE_ENEMIES + DOTA_UNIT_TARGET_FLAG_FOW_VISIBLE,
```
carlosrpg: Why do you find only the visible units? If I'm invisible and in the area I should take damage

TideSofDarK: this particular FindUnits call is used only for deciding if we should cast it. actuall impact searches for all units

PR #2184, opened 2018-04-30
Azazel's buildings

ZzZombo:
Notes:

- Still no suitable model for the Watch Tower found. The current one is only a placeholder, and is NOT usable in real game!
- There is currently no safeguard against placing the Wall such so absolutely no segment is possible to place there, rendering the item usage wasted.
- Roshan's Bash can bash buildings, you probably want to look into this.
- I could do vector targeting for the Wall, but I would like a bounty raise then, due to the fact there are many gotchas in implementing it yourself, since Valve didn't expose, AFAIK, Swashbuckler's targeting to custom games yet.
- The Wall should, IMO, have shared with the team ability to destroy its individual segments, since an improperly placed one can really mess the game up w/ no way of getting rid of it from the team that placed it, blocking a passage indefinitely.


l34Um1:
- the items are missing icons and tooltips
- the items are being consumed on use
- the KVs currently include "bonus" health and damage values and then the level 1 is hardcoded. The level 1 should also be in KV.
- the towers should only start giving vision once they have been fully constructed
- for now please place them items in the regular shop so players can buy them

ZzZombo:
You sure you tested both commits? Because it sounds exactly like if you for some reason only pulled only the first; for instance, I did put tooltips ([example](https://github.com/OpenAngelArena/oaa/pull/2184/files#diff-c51a41906e4573e8cbf54077b63d898c)). As for the missing icons, there were no requirements to make them, and you also told me to "just put items in", so I went with submitting w/o icons; what exactly do you want me to do with them? In what shops should they be?

l34Um1:
Oh yeah, they are *not* being consumed, my bad

l34Um1:
Few things then I think we are done:

- put the items in the first shop page, not the second
- the items currently don't require cores. they should require cores to buy.
- the watch tower is not selectable for some reason
- reduce the model size of both towers. they are too big atm.

ZzZombo:
I'm done, take a look now, @l34Um1, please.

l34Um1:
Apart from still missing tooltips this all seems to be in order and ready to merge. I'll need your paypal information so I can pay you once it has been merged.

ZzZombo:
Sorry, I don't think we are on the same wavelength here. No tooltips are missing, I know for sure. Even the tests pass successfully for me and Travis. Are you sure you have run the tooltip script locally?

l34Um1:
When towers finish building they have their HP set to 1000. So as you build a tower it will have the KV set HP, then, when it finishes, it's set to 1000.

game/scripts/vscripts/items/azazel_tower_defense.lua
```diff
@@ -0,0 +1,158 @@
...
+
+function modifier_defense_tower_construction:OnCreated()
+  if IsServer() then
+    local target = self:GetParent()
+    local ab = self:GetAbility()
+    local maxhealth = target:GetMaxHealth() + ab:GetSpecialValueFor("bonus_health")
+    local location = target:GetOrigin()
+    local time = ab:GetSpecialValueFor("construction_time")
+    target:Attribute_SetIntValue("construction_time", time)
+    target:Attribute_SetIntValue("bonus_damage", ab:GetSpecialValueFor("bonus_damage"))
```
Trildar: I think just saving these as properties on `self` makes more sense, but it's fine as is.

game/scripts/vscripts/items/azazel_tower_defense.lua
```diff
@@ -0,0 +1,158 @@
...
+function modifier_defense_tower_construction:OnIntervalThink()
+  if IsServer() then
+    local target = self:GetParent()
+    local time = target:Attribute_GetIntValue("construction_time", 10)
+    local count = self:GetStackCount()
+    local location = target:GetOrigin()
+    target:SetOrigin(target:GetOrigin() + Vector(0, 0, SINK_HEIGHT / (time / THINK_INTERVAL)))
+    self:SetStackCount(count - 1)
+    if count < 1 then
+      self:StartIntervalThink(-1)
```
Trildar: I think the modifier should just self-destruct (`self:Destroy()`). And the other stuff that persists after construction like damage put on another modifier.

game/scripts/vscripts/items/azazel_tower_defense.lua
```diff
@@ -0,0 +1,158 @@
...
+    end
+  end
+end
+function modifier_defense_tower_construction:GetAttributes()
+  return MODIFIER_ATTRIBUTE_IGNORE_INVULNERABLE
```
Trildar: Is this actually necessary?

game/scripts/npc/items/item_azazel_wall_4.txt
```diff
@@ -0,0 +1,67 @@
...
+        {
+            "01"
+            {
+                "var_type"                  "FIELD_INTEGER"
+                "construction_time"         "10"
+            }
+            "02"
+            {
+                "var_type"                  "FIELD_INTEGER"
+                "bonus_health"              "0 4500 10500 22500"
```
Trildar: You're already using `SetMaxHealth`. I think the KV might as well be just the total health.

game/scripts/npc/items/item_azazel_tower_watch_4.txt
```diff
@@ -0,0 +1,77 @@
...
+            }
+            "02"
+            {
+                "var_type"                  "FIELD_INTEGER"
+                "bonus_health"              "0 1500 3500 7500"
+            }
+            "03"
+            {
+                "var_type"                  "FIELD_INTEGER"
+                "bonus_vision_range"        "0 500 1000 1500"
```
Trildar: Can be changed to just be the total vision range by using the set vision functions, or the fixed vision modifier properties.

game/scripts/vscripts/items/azazel_tower_watch.lua
```diff
@@ -0,0 +1,142 @@
...
+  if IsServer() then
+    local target = self:GetParent()
+    local count = self:GetStackCount()
+    if count > 0 then
+      local time = target:Attribute_GetIntValue("construction_time", 10)
+      local location = target:GetOrigin()
+      target:SetOrigin(target:GetOrigin() + Vector(0, 0, SINK_HEIGHT / (time / THINK_INTERVAL)))
+      self:SetStackCount(count - 1)
+    else
+      AddFOWViewer(target:GetTeam(), target:GetOrigin(), target:GetCurrentVisionRange(), THINK_INTERVAL, false)
```
Trildar: This is cringey as anything, but I'm actually not sure there's a better way to do it. I would at least like it to have a higher refresh interval, though I'm unsure what the cleanest way to do that would be. Maybe just another modifier, and have this one self-destruct like the other construction modifiers ought to do.

game/scripts/vscripts/items/azazel_tower_defense.lua
ZzZombo: I don't think a load of modifiers is any better. But if you insist, so be it.

game/scripts/vscripts/items/azazel_tower_defense.lua
ZzZombo: Yes. What if Song of the Siren affects them?

game/scripts/vscripts/items/azazel_tower_watch.lua
ZzZombo: I can make a higher interval for this.

game/scripts/npc/items/item_azazel_wall_4.txt
ZzZombo: Yeah, I'll do this.

game/scripts/vscripts/items/azazel_tower_defense.lua
ZzZombo: Renaming it then seems the way to go then, IMHO.

Trildar: Well, there's also the matter that this method requires a bunch of places to check whether construction is done.

game/scripts/vscripts/items/azazel_tower_defense.lua
ZzZombo: Are we still discussing the ultimate fate of this piece of code or just chit-chatting about modding? To be sure.

ZzZombo:
Paid, thanks.

PR #2186, opened 2018-05-05
Disable BAREBONES_DEBUG_SPEW

Trildar:
This should never have been enabled in the first place. It spams the console so hard.


carlosrpg:
Thank GOD! that was annoying

PR #2196, opened 2018-05-12
Aeon disk tooltip fix

GymLeaderGiovanni:


Trildar:
The key is just "...status_resistance", not "bonus_status_resistance. There's also another error about Mango that I think is because your branch is outdated.

l34Um1:
I had that same error, you just have to change mango Regen value to 0.6

GymLeaderGiovanni:
should be all good now

PR #2202, opened 2018-05-19
Azazel Temp Shop + fix to summon second summon bug

carlosrpg:


carlosrpg:
@l34Um1 can you check if this is can work until we have the new Shop?
image

I got one column for Buildings / Summons / Elixirs

l34Um1:
Nice seems good

PR #2205, opened 2018-05-21
Swiper Boss

Trufllehunt:
I believe this is everything ive done so far. 
Swiper Model and textures
Animations for model
will add icons later


warpdragon:
someone throw this on a test client

carlosrpg:
The smd are left over files or you use both fbx and smd for animations?
I'll do the code to integrate with the swiper and create a test client for us to play with him

Trufllehunt:
only the .fbx files are animations. The smd was accidental upload

carlosrpg:
You need to add the materials to this PR

PR #2216, opened 2018-05-31
patch 7.16

imaGecko:


game/scripts/npc/abilities/doom_bringer_infernal_blade.txt
```diff
@@ -29,7 +29,7 @@
       "01"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "burn_damage"                                     "25 30 35 40 60 90"
+        "burn_damage"                                     "25"
```
carlosrpg: I think we should scale this in oaa

carlosrpg:
Blades of Attack, mango and sentry.
The removal of most of the things intentional?

game/scripts/npc/abilities/doom_bringer_infernal_blade.txt
imaGecko: not necessary since the % hp damage scales up quite a bit and they also removed scaling of the flat damage

imaGecko:
yeah it was intended because those items don't have any changed stats, they just exist for shop or adding charges

carlosrpg:
OK, blind merge then

imaGecko:
inb4 game breaks

PR #2223, opened 2018-06-07
Fix missing concat operators in hero_kills.lua

Trildar:
Once again, this would have been a syntax error in linters if only the no brackets call syntax didn't exist. *mumble* *grumble*


chrisinajar:
I hate lua.

PR #2248, opened 2018-06-18
Save/load state

chrisinajar:
Copy and pasted lots of code from @carlosrpg, refactored simplified and rebased on the latest code.

Still missing a bunch of stuff, and of course needs to be integrated with the bottlepass server...


chrisinajar:
* [ ] Heroes
  * [x] xp/level
  * [x] items (including backpack / stash) including charges
  * [x] ability level and cooldowns
  * [x] current location
  * [x] gold
  * [x] Silencer stolen intelligence 
  * [x] LC duel damage
  * [ ] Pudge flesh heap
  * [ ] Lone Druid bear items
* [x] passive xp/gold changed to be based on game clock
* [x] Boss level
* [x] creep camp levels
* [x] keep creep spawns synced with clock
* [x] game time (hud clock)
* [x] time of day (day/night cycle)
* [x] team scores
* [ ] ward positions / time remaining (poop and non-poop)


What else does it need for MVP? I don't think that living creeps (stack up) nor azazel nor summons are worth tracking for V1 unless an incredibly simple solution is developed...

chrisinajar:
Needs to change all the playerID based values to instead be steamid based, and then the hero selections need to be stored in a dedicated tree of the state. Right now the order at which the lobby was joined is enforced since playerID is just player slot #...

That + wards = i vote its good to go and we finish up the server integration (I have most of the server already written)... then it just needs `-save` and `-load` which work outside of cheats mode so that we can test it.

PR #2249, opened 2018-06-21
Fix tier 4 Uriel ability list to match tier 5

Trildar:


chrisinajar:
lol we dumb, nice find trildar 👍

Issue #2236, opened 2018-06-12
Fix Tranquil Boots


l34Um1:
Currently they gather no charges upon movement. This is a simple bug fix.

PR #2251, opened 2018-06-22
Fix Greater Tranquil Boots not doing anything

Trildar:
Seems like Valve changed the behaviour of SetDuration so that setting it to 0 immediately destroys the modifier even when DestroyOnExpire is false #ThanksValve. Either that, or this has been broken since #1748, but I'm certain I tested it, so that's quite unlikely.

### DIFF
--- a/game/scripts/vscripts/items/farming/greater_tranquil_boots.lua
+++ b/game/scripts/vscripts/items/farming/greater_tranquil_boots.lua
@@ -272,7 +272,10 @@ if IsServer() then
 
 			-- less so is actually making this do anything
 			-- because valve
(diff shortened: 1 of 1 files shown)


Trildar:
Fixes #2236.

Haganeko:
Does line 275 also require changing?

carlosrpg:
Can you explain what does Duration does for Tranquils? Did not knew any Cooldown on the spell until now

Trildar:
@Haganeko That one should never be 0 since the cooldown is started right before it. Though maybe it's a good idea to put in a redundant check anyway.

PR #2257, opened 2018-06-27
status resistance working for all debuffs

Yahnich:
note that modifier filter works before the modifier is applied, so findmodifierbyname will not work unless a buff is refreshed; only way to find if a modifier is a debuff currently is comparing caster and parent teams


chrisinajar:
utACK

PR #2260, opened 2018-07-03
Swiper Model for testing (Courtesy of Trufllehunt)

warpdragon:
ability icons, materials, animations, and models to be used for Swiper visual update. I have NOT implemented the model in the unit's KV file because I'm not able to test at the moment.

Anyone who wants to test simply has to change the "Model" entry in:
`game/scripts/npc/units/swiper/npc_dota_boss_swiper.txt`
from:
`"Model"                                               "models/heroes/sven/sven.vmdl"  // Model.`
to:
`"Model"                                               "models/bosses/swiper/swiperreal.vmdl"  // Model.`


warpdragon:
What do u mean? Does it work?

carlosrpg:
yeap, with animations and such
It needs better texturing tho

warpdragon:
i don't think we need this PR anymore

PR #2263, opened 2018-07-06
Duel start invulnerability

DimasDSF:
Should work unless I completely misunderstood how to get players in the active duel
(I was not able to test it since I cannot start a duel alone)

The modifier cannot be purged by Nullifier unlike the default `modifier_invulnerable`

- Probably needs some ParticleFX since it would confuse people who dont read changelogs and find it weird when enemies walk out of their perfectly placed chronosphere

### DIFF
--- a/game/scripts/vscripts/components/duels/duels.lua
+++ b/game/scripts/vscripts/components/duels/duels.lua
@@ -2,6 +2,7 @@ local HeroState = require("components/duels/savestate")
 local SafeTeleportAll = require("components/duels/teleport").SafeTeleportAll
 
 LinkLuaModifier("modifier_out_of_duel", "modifiers/modifier_out_of_duel.lua", LUA_MODIFIER_MOTION_NONE)
+LinkLuaModifier("modifier_duel_invulnerability", "modifiers/modifier_duel_invulnerability", LUA_MODIFIER_MOTION_NONE)
 
 DUEL_IS_STARTING = 21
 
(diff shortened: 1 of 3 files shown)


game/scripts/vscripts/modifiers/modifier_duel_invulnerability.lua
```diff
@@ -0,0 +1,20 @@
...
+
+function modifier_duel_invulnerability:IsHidden() return true end
+function modifier_duel_invulnerability:IsPurgable()	return false end
+function modifier_duel_invulnerability:DestroyOnExpire() return true end
+
+
+function modifier_duel_invulnerability:OnCreated(kv) 
+	if IsServer() then
+		self.duration = kv.duration
+		self:SetDuration(self.duration, true)
```
Trildar: This is completely unnecessary. `duration` is a special key that will spawn the modifier with the set duration automatically.

game/scripts/vscripts/modifiers/modifier_duel_invulnerability.lua
```diff
@@ -0,0 +1,20 @@
+modifier_duel_invulnerability = class({})
+
+function modifier_duel_invulnerability:CheckState()
+	return {
+	[MODIFIER_STATE_INVULNERABLE] = true,
+	[MODIFIER_STATE_MAGIC_IMMUNE] = true,
```
Trildar: I think magic immunity is redundant when they're already invulnerable, but I suppose it doesn't matter anyway.

game/scripts/vscripts/components/duels/duels.lua
```diff
@@ -444,6 +445,23 @@ function Duels:ActuallyStartDuel(options)
     end
   end
 
+  --duel start invulnerability
+  --Invulnerability Duration/ Should probably move this to options
+  local invulnerabilityduration = 2.0
```
Trildar: This is bad naming, please use either camelCase or snake_case. Or at least something other than all lowercase with no separators.

Trildar: I would actually suggest going the opposite direction and making it a global set in `settings.lua`.

PR #2269, opened 2018-07-10
make drums not crash the game

imaGecko:
"-1" charges don't work with drums
drums don't have an effectwhen activated with 0 charges either, so a high number of charges will be essentially unlimited


chrisinajar:
Does `ItemDisplayCharges` need to be activated? It'd be great to hide this from players.

imaGecko:
I tried to have it on or off, it didn't actually hide the charges

DarthRecca:
Can you try it out with my drums implementation? Because it didn't crash my game, also it would complete the testing tack.

imaGecko:
It was mostly me being in doubt whether new drums would get into the game pre tournament, this would remove a (sometimes on purpose?) lag/crash from the game without changing functionality

chrisinajar:
Darth, your solution is riskier than this solution _for now_. We have a tournament in a few days so stability is extremely important, but after that we're gonna go nots on patches so it'll definitely get in then 👍

PR #2273, opened 2018-07-15
Add ability icons in pick screen

chrisinajar:


chrisinajar:
take that, trildar :P

PR #2274, opened 2018-07-15
Devastator instant attack and tooltips

imaGecko:
All enemies hit will now also be hit with an instant attack
Made the damage physical
Now applies minus armour before dealing damage
Added tooltips
Fixed devastator shop placement and added it to the missing shops
Moved kvfiles into the custom folder


imaGecko:
Tested it in game and it seems to work

PR #2283, opened 2018-07-17
make furion spawn levelled treants

imaGecko:
Natures call will spawn a treant based on the level of the ability
Nature's Wrath using Aghs will spawn a treant or a large treant based on the level of Nature's call


imaGecko:
it's down to 1 error, not sure what to put there as it is a copy of the lua example ability

PR #2284, opened 2018-07-18
better icon, still too small

warpdragon:


warpdragon:
Basically the existing icon, the pink dot, is ugly and doesn't really communicate anything about the objective. The new icon imitates the appearance of hte control point (a hollow cylinder wall) aka an empty 3D circle with depth. It also is roughly the same size as the capture point, giving it even more value.

We gucci? Do we like the color? @l34Um1 
image

warpdragon:
screencap

detail

warpdragon:
Punch it Chewie!

PR #2301, opened 2018-07-20
Enable hidden duel arena and small duel code refactory

carlosrpg:


carlosrpg:
imo that new arena just need a few trees

warpdragon:
1 sec

chrisinajar:
I like that you made it so the larger duel group always goes into a large duel pit, also means final duel will always be in a large duel pit. I think that's an improvement.

PR #2303, opened 2018-07-21
Electrician hero selection model

carlosrpg:


carlosrpg:
@chrisinajar don't forget that you need to manually compile the vmap

PR #2314, opened 2018-07-24
Fix get dota stats exception from missing cave

chrisinajar:


chrisinajar:
gonna do a bunch of these in 1 pr

PR #2321, opened 2018-07-25
flesh heap minus talent

imaGecko:


chrisinajar:
Can someone add in the talent functionality and then also make it work with save/load state?

DarkoniusXNG:
Not needed anymore.

PR #2327, opened 2018-07-26
Minimap Icons

Satsaa:
This would change the minimap creep camp icons to be similar to Dota 2's. Hues and shapes of old icons are mostly preserved. Icons are slightly darker. Ancient and hard camps are a tiny bit smaller.
preview


warpdragon:
SeemsSohei

PR #2331, opened 2018-07-28
Swiper files

Trufllehunt:
All textures and model needs
All but rush animation as im having difficulty importing it


Trufllehunt:
There are also a few issues with the animations playing properly. Part if the purpose for this PR is due to carlos request so he can help

PR #2334, opened 2018-07-29
Add ranked picking

chrisinajar:


chrisinajar:
Tested with a few small lobbies, seems to work :+1:

PR #2335, opened 2018-07-29
D x S balance patch

chrisinajar:


l34Um1:
My comments are regarding the individual commits. I made them on the commits, it didn't really work.

game/scripts/npc/items/custom/item_dagger_of_moriah.txt
```diff
@@ -78,7 +78,7 @@
       "02"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "sangromancy_self_damage"                         "100 100"
+        "sangromancy_self_damage"                         "75 100"
```
imaGecko: second level shouldn't really be worse to use than the first level. I would put both at 75

imaGecko:
I would hold off making exceptions for refresher and bkb for the time being. Implementing a purge and/or waiting time after the duel tp would help with the triple ultimate shenanigans, if havingthe item ready every duel still proves to be too strong afterwards then the exceptions can still be introduced at that point

chrisinajar:
Reverted all the things baumi disagreed with.

chrisinajar:
@l34Um1 give it another quick look through to make sure it's good, i should have reverted everything you said no to :+1:

l34Um1:
seems good

Issue #2165, opened 2018-04-06
Add a Duel Timer, Kill Score Limit and capture point timing indicator | 20€ Bounty


l34Um1:
![unbenannt-1](https://user-images.githubusercontent.com/24982519/35267044-50285c80-0025-11e8-8179-d96915598fc1.jpg)

Picture taken from Black Star. Similar to this. The menu below the clock shows up when you hold alt. Add the text "Score limit" and "Next Duel" in front of the 2 numbers, as well as a third value "next capture point" and the corresponding timer.

chrisinajar:
I like not having duel timer, I disagree it should be added. I'd be fine with it appearing during the last 60 seconds, but not before then.

Haganeko:
Why so? I thought this had been debated and all objections overruled.

carlosrpg:
Claim

carlosrpg:
@l34Um1 I would like to receive that on a different paypal account, let me know when you plan to pay this

DarkoniusXNG:
This was implemented.

PR #2337, opened 2018-07-29
Oaa dashboard

carlosrpg:
Fix Issue #2165


carlosrpg:
image

chrisinajar:
Does it only show up when you're holding alt?

carlosrpg:
It was in the ticket as it was done on other game I think

chrisinajar:
that was not an answer to my question at all, hahahahah

I assume that means "yes" since the ticket says `Picture taken from Black Star. Similar to this. The menu below the clock shows up when you hold alt. `

carlosrpg:
oh, yep, the answer  is Yes

PR #2340, opened 2018-08-01
Clinkz's Strafe's cooldown can be reduced by new lvl 25 talent.

AbitTheGray:
Added new lvl 25 talent to reduce cooldown of Strafe (Q).
For Issue #2297

PR #2342, opened 2018-08-01
Sohei's ultimate range Aghanim fix

AbitTheGray:
When solving problems for Clinkz's Strafe ( #2340 ) I found that I may have problem with dropping aghanim when if I did it Sohei's way.

Thanks to @Yahnich for `caster:FindTalentValue()` which solved it.


game/scripts/npc/abilities/sohei_flurry_of_blows_oaa.txt
```diff
@@ -68,12 +69,6 @@
 				"var_type"					"FIELD_FLOAT"
 				"attack_interval"			"0.6"
 			}
-      // the bonus must be set on sohei_talents_oaa.txt to update its tooltips
-      "05"
-			{
-				"var_type"					      "FIELD_FLOAT"
-				"talent_bonus_radius"			"200"
```
chrisinajar: As per the comments, this is needed for tooltips. The key should end in `_tooltip` so that's more apparent. The rest seems legit :+1:

AbitTheGray: `"talent_bonus_radius"` is NOT used in description of the ability and the value `"200"` is stored in `sohei_talents_oaa.txt` which can be retrieved using `caster:FindTalentValue()`

Issue #2160, opened 2018-04-06
New Mirana level 25 Talent | 5€ Bounty


l34Um1:
"Global Range Sacred Arrow".

As the name implies the arrow now has global range. This does not increase the damage or stun duration of the arrow. It simply makes the arrow not have a maximum travel distance and gives it global cast range.

This talent replaces "-80 Moonlight Shadow Cooldown".

AbitTheGray:
Gonna try it.

AbitTheGray:
Ready but still require in-game test and setup git for pull request.
And no, I did not forget to implement arrow multi-shot or cooldown reduction.

AbitTheGray:
Status:
- Multishot works, can also add more then 2 arrows
- Arrows do not trigger on hit
- Arrow stun not implemented (but calculated), deals only damage (if only on hit trigger work)

AbitTheGray:
Finally got it working.
Gonna PR in about a day (still need to prevent multiple arrows (lvl 25 talent) to hit same unit) but it is working!

One small but will appear because ExtraData of projectile cannot transfer Vector (which did not cause the problem but I needed to fix it too).
With refresher (or WTF) you can get maximum arrow damage/stun by recasting it further away the target before the first arrow hits. This is because each cast resets start (cast) location which is referenced by arrows to calculate stun duration.

DarkoniusXNG:
This can be closed.

PR #2344, opened 2018-08-03
[FIXED_BRANCH] Clinkz's Strafe's cooldown lvl 25 talent

AbitTheGray:
Added new lvl 25 talent to reduce cooldown of Strafe (Q).
For Issue #2297

This time right branch not like #2340 .

PR #2346, opened 2018-08-07
Mirana piercing arrow talent

AbitTheGray:
For #2160 

Still require testing (and minor code to write) but looks working in general.

Do not know why `npc_heroes_custom.txt` is having "problem" (huge amount of changes on GIT) with 1 new line.

Left all arrow-related talents in their original form. Even allow more-then-two added multishot arrows to be set in the other lvl 25 talent.

TODO
- [X] Kill creeps (not ancient creeps or hero or creep-hero (considered-hero)) on hit `target:ForceKill(reincarnation=false)`
- [X] Verify hitting but not affecting spell-immune units
- [X] Keep vision on target point for 3s instead of (currently) duration of the stun
- [X] Verify that multishot cannot hit same target
- [X] Verify arrow cannot hit invulnerable units
- [X] Change to first apply damage, then debuff
- [X] Start arrow `51` units in front of caster (change current value to be loaded from `AbilitySpecial`)
- [X] Verify travel speed of `857`
- **UNABLE** Fix ability to be displayed on picking screen (not a problem because it is same ability but...), not a bug but feature!
  - Can be solved by modifying picking screen

Personal note: Replace this new talent with talent that allows the arrow to pass through targets triggering on all of them. Seems more fitting for level 25.


chrisinajar:
it's just because you saved the file with an editorconfig active, the npc custom thingy

AbitTheGray:
Same as the Clinkz from #2344 this ability does not appear on hero picking screen instead the old (original) one.
Not tested translation, cannot get it working.

AbitTheGray:
Added sounds on both cast and impact.
[Fly sound](https://dota2.gamepedia.com/Mirana/Sounds#Sacred_Arrow) is not used in Dota.

**Not tested (yet).
Will test ASAP.**
Should work anyway.

chrisinajar:
does it have double cast to throw backwards?

AbitTheGray:
**Not tested** - will test tomorrow after work.

Number of pierces is defined in `mirana_talents.txt` file in `game/scripts/npc/abilities/`.
The amount of pierces can be (but currently is not) referenced by tooltips using `arrow_pierce_count` (not tested).

Also replaced the Global Arrow with this near Pierce Arrow (with no translation - I am going into bed 23:54 here).

AbitTheGray:
**WIP** of 2nd talent - Arrow Piercing Units.
Gonna find a way to modify Arrow's ExtraData or recast the arrow on every hit with decreased hit count by 1.

AbitTheGray:
I though I had it fixed but just tested it and it goes through all units.

Will take a look at it tomorrow (sadly I only got time today to verify Kunkka's Tidebringer).

AbitTheGray:
@chrisinajar Now I am happy with the current state.

The targeting cursor snaps to enemy unit and aggro creeps but I do not consider it a bug but how unit-target abilities work (which is required for self-cast, your way did not allow casting with cursor over the unit).

chrisinajar:
You need to reject those other targets in a unit cast filter, it prevents it from even highlightning them. I got that working in my upgradable ward stack item.

PR #2357, opened 2018-08-10
Balance suggestions

chrisinajar:


chrisinajar:
Rationale:
### Magic Wand
Magic wand is an item you either buy already maxed for the stats, or buy at level 1 for early roaming aggression. Removing just the level 1 version of the item is a substantial nerf to supports, and i love aggressive supports, so I added it back. Cheap support buildup items are very important since they can't afford the other stronger early game items

### Halberg
This is more or less a nerf to all right click carries -- allow the richer supports to start disabling the enemy right clicks earlier. I agree that Halberg was a strong item, so I also nerfed the lower levels and added scaling to the duration. The max level version has a slightly higher duration than before, making upgrading this item extremely worth it.

### Battlefury
bfury is a farming item and deso is only a farming item on a very very limited selection of the cast. I made bfury also build out of Maelstrom (nor Mjolnir) to give it a more farming-centric build up. Lightnings are often bought in place of bfury since it gives a sort of cleave effect, bouncing the magic damage across the camp. Additionally, the cost is more in line with the previous cost of bfury, and you have the option to build mjolnir instead if you'd like.

### Halberg -> Nullifier
Nullifier building out of echo sabre is "not terrible" at best, and on any ranged hero it is ubercringe. This gives ranged heroes, and many melee heroes as well, a much better build up to Nullifier. Nullifier is purchased as a disable item with damage, so I thought that Halberg would be a good build up in to it. Another good option would be Diffusal, but I like Halberg personally.

### Kaya Buff
Slight numbers buff to kaya to try and make it more worth buying. This item is basically only bought in specific cheese strats where you go *all* spell amp or as a build up item into Moriahs. I'd like to make spellcasters a bit stronger against carries, this should help and also make the item get picked more.

### Radiance build order
Item recipe order was wrong and it makes the shop act weird, this isn't a balance change.

### Mask of Madness vs Satanic
Satanic is a stronger later-game item than MoM is. MoM is also a farming item on many heroes. It is because of these reasons that I believe MoM should come first and build into satanic, as opposed to the other way around.

### GGG only build from Arcanas
Currently you can upgrade other boots straight to GGG without going through arcanes. I think it's worth forcing the extra gold expenditure in order to get the purge effect from GGG boots, you shouldn't be able to skip the mek purchase to get these. Also it removes 1 boot buildup and that's nice because the UI is confusing with so many options.

### Lucience build out of drums
It feels wrong because we've had Lucience for so long, but drums is a movement speed aura item and radiance has nothing in common with Lucience other than the last 3 letters. The only reason anyone wont want this is because they're used to the current buildup.

### Fix GGG boot recipe
You can't upgrade GGG boots into other boots reliably because they all accidentally required the GGG recipe as well. This fixes that and makes the upgrades smoother.

### Tranqs and Travels buff
These two boots are the least popular, but I think their numbers are reasonable. I made both easier to use by making them less restrictive -- Travels have 100 more radius on the gold assist at all levels, making it easier to be in range of farm. Tranqs have rupture's distance limit instead of mana leak's, allowing heroes with short blinks to crank up stacks much quicker. There are already a few heroes who can abuse tranqs well, this just adds more. Even heroes who can abuse tranqs often don't buy them now.

### No more travels 2
Currently everyone buys BOT2 and puts it in their backpack so they can TP to an allied hero at any time. This change makes the teleport-to-hero functionality gated by a farming core, and also makes the completed greater travel boots 500 gold cheaper which makes it much more appealing to poor supports.

l34Um1:
1. Remove all Magic Wand upgrades. Keep it as an early game item.
2. Halberd level 1, Echo Sabre level 2, Nullifier level 4.
3. Undo.
4. See 2.
5. Good.
6. Fine.
7. I think we MoM instead. 
8. Fine.
9. Drums and Lucience go into Helm of the Dom and Vladimirs. Sange and Yasha goes into Skadi.
10. Fine.
11. No travel nerf. Tranqus give more money instead.
12. Thank god.

PR #2383, opened 2018-08-21
Rebalance Armlet

warpdragon:
rebalanced Armlet. Overall this is a minor buff to regen and unholy strength, as well as all benefits from the addition of strength. Most importantly, the calculated number of seconds it takes to degen all Unholy Bonus HP goes 9/10/11/12/13 instead of what it was before 9/14/14/13/13.

My calculations:
graph


warpdragon:
Oh and i also fixed a typo

PR #2384, opened 2018-08-22
magma mage boss particle and code

jboxGIT:
travis


jboxGIT:
this adds magma mage boss and his volcano ability but does not change anything to make him spawn anywhere. so there's quite a bit for design/balance people to do. all that's left for me to do is polishing the particles and adding temp sounds.

PR #2172, opened 2018-04-15
Item: Sonic

genious97:

Issue #1549, opened 2017-12-06
**TRANSFORMATION** Item: Sonic


warpdragon:
# Sonic
## Explanation
Toggle. When toggled on your hero gains flying movement, increased movement speed and bonus vision. Your hero also gains negative status resistance, meaning anything that would usually be reduced by status resistance gets increased instead.
## Recipe
- ancient_janggo_2, upgrade_core_2, recipe
## Behavior
- Targetting: Self/Instant
- New Active Ability: Increased movement speed, flying movement, vision bonus, negative status resistance (status vulnerability)
![sonic](https://user-images.githubusercontent.com/13878439/33667930-46f8f82e-da6c-11e7-95d8-01e3ebf4cc0e.png)
- Notes: Breaks trees in small AoE at the end to prevent pathing errors.
- cooldown, movement speed bonus, negative status resistance, vision bonus

## KV
- speed_bonus, vision_bonus, negative_status_resist, janggo stats

Haganeko:
When was this design debated?

warpdragon:
it wasn't

Haganeko:
Then just putting it up for implementation is bad form.

warpdragon:
So it feeding 11 kills as Spectre in 15 minutes.

Haganeko:
Irrelevant.

Haganeko:
Irrelevant.

deadshoes:
hello completely new to this, looked through the kv tutorials today and made an attempt at creating this item. I dont really know how to use the dota client to test this though.
[item_sonic_1.txt](https://github.com/OpenAngelArena/oaa/files/1857236/item_sonic_1.txt)

Haganeko:
You need to look at the install instructions for how to test things. Also, the KV by itself will do nothing. You'll need a lua code as well.

deadshoes:
my bad i thought i could avoid lua by making it a data driven item

Haganeko:
You can, but I thought the toggle with a cooldown would be impossible to make datadriven.

Yahnich:
don't use datadriven

genious97:
Is anyone currently working on this issue, because if not I would be interested?

zelding:
in progress at #2172

mikibun:
@zelding Is this still being worked on?

DarkoniusXNG:
This was implemented.

PR #2392, opened 2018-09-05
Finished Sonic item

VilkkuV:
Finished #1549. Continued from the previous pull request #2172. The item works with the custom status resistance code. Only problem now is that you can use the flying movement to move out of the map.


game/scripts/vscripts/components/filters/statusresistance.lua
```diff
@@ -30,7 +30,7 @@ function StatusResistance:StatusResistanceFilter(filterTable)
       if modifier.GetModifierStatusResistanceStacking and modifier:GetModifierStatusResistanceStacking(params) then
         stackResist = (stackResist or 0) + modifier:GetModifierStatusResistanceStacking(params)
       end
-      if modifier.GetModifierStatusResistance and modifier:GetModifierStatusResistance(params) and modifier:GetModifierStatusResistance(params) > resistance then
+      if modifier.GetModifierStatusResistance and modifier:GetModifierStatusResistance(params) and math.abs(modifier:GetModifierStatusResistance(params)) > math.abs(resistance) then
```
chrisinajar: i think this math.abs isn't right here, really non-stacking status resistance should never be negative, it only makes sense with stacking...

game/scripts/vscripts/modifiers/modifier_sonic_fly.lua
```diff
@@ -0,0 +1,28 @@
+modifier_sonic_fly = class(ModifierBaseClass)
+
+function modifier_sonic_fly:DeclareFunctions()
+  return {
+      MODIFIER_PROPERTY_MOVESPEED_BONUS_CONSTANT,
+      MODIFIER_PROPERTY_BONUS_VISION_PERCENTAGE,
+      MODIFIER_PROPERTY_STATUS_RESISTANCE
```
chrisinajar: should probably be stacking right?

VilkkuV:
Made the requested changes.

PR #2394, opened 2018-09-10
Disjoint (most) projectiles when duel starts

DarkoniusXNG:
Lich ulti, Alchemist stun, Oracle Q, Witch Doctor's stun still cannot be disjointed.


chrisinajar:
hahaha, i like the way you did it :+1:

Issue #2233, opened 2018-06-12
Towers shouldn't always be visible on the minimap, only when in sight


l34Um1:
Currently, after a tower has been placed, you can always see its location on the minimap, even if you are on the enemy team.

This should be changed so that towers are treated the same on the minimap as heroes. Only visible to the enemy if they have vision on them.

DarkoniusXNG:
BaseClass of **npc_azazel_tower_defense** and **npc_azazel_tower_watch** is causing this issue. Changing the BaseClass would solve the problem but it would raise more issues (for example if you change the BaseClass to npc_dota_creature, towers will be like creeps - not be immune to spells and cleave). 

Applying a modifier with MODIFIER_STATE_NOT_ON_MINIMAP_FOR_ENEMIES to towers (and walls) could solve the issue only if Valve didn't hardcode the visibility of buildings in fog.

zelding:
so we would have to re-implement buildings ?

DarkoniusXNG:
Yes, or add bunch of exceptions in Damage Filter.

PR #2395, opened 2018-09-10
Hide Buildings in FoW and on minimap

DarkoniusXNG:
* MODIFIER_STATE_NOT_ON_MINIMAP_FOR_ENEMIES doesn't work some (Valve) reason. 
* Buildings cannot be hidden in a fog of war without using some wacky workaround. 
Here are my thoughts on that: 
1) OnIntervalThink check if the enemy heroes have vision on the building. If they don't have vision, hide the tower with AddNoDraw() and apply MODIFIER_STATE_NO_HEALTH_BAR. This doesn't seem good because you and your allies will not see your tower too.
2) Make buildings pop out from the ground when the enemy comes near them. Otherwise they are underground and hidden (not visible to everyone at all). SetAbsOrigin(Vector(origin.x, origin.y, origin.z - height_of_the_building))
3) Change BaseClass of buildings to npc_dota_creature. Buildings would be easier to destroy with all spells. This solution is good only if buildings need to be nerfed.


DarkoniusXNG:
https://github.com/OpenAngelArena/oaa/issues/2233

chrisinajar:
`Change BaseClass of buildings to npc_dota_creature. Buildings would be easier to destroy with all spells. This solution is good only if buildings need to be nerfed.`
this option.... FWIW charger towers are creatures for this same reason :+1:

chrisinajar:
I am extremely unsure how I want to handle buildings, after the tournament we realize they're **a lot** stronger than we originally thought. I think we probably still want a dota creature based solution though, having them statically appear on the minimap at all things is pretty broken and wacky, on the other hand not being able to identify them quickly is even more broken... i dunno...

DarkoniusXNG:
We could change only Watch Towers for now. Their BaseClass I mean.

chrisinajar:
thinking more, definitely creature baseclass. I actually like that they'll start taking magic damage (not sure what their resistance should be, probably high)

DarkoniusXNG:
Are you sure about this? Lich Sacrifice, Enigma Demonic Conversion and Lifestealer Infest will insta-kill buildings.

DarkoniusXNG:
Oh wait the list goes on:  Clinkz Death Pact, HOTD, Chen Holy Persuasion, Doom Devour, Hand of Midas, Enchantress Enchant.

DarkoniusXNG:
I can add "ConsideredHero" "1" but that would give buildings styled health bar. [http://puu.sh/ho2pt/c687566db4.jpg](url)

DarkoniusXNG:
* Changed BaseClass to npc_dota_creature. (This caused lot of bugs.)
* Added magic resistance to defense towers and walls.
* Watch towers don't have magic resistance but are hidden on minimap (even from allies).
* Added ConsideredHero flag so buildings cannot be killed instantly (or dominated) with some abilities.
* Fixed ability in modifiers being nil sometimes because of RemoveItem(..) - replaced code that controls charges with line: ability:SpendCharge()
* Created FindCustomBuildingsInRadius(position, radius) so it can be used for finding this custom buildings.
* Fixed buildings being able to be placed on top of other buildings.
* Removed building:SetOrigin(..) because its not needed.
* Commented parts of the code that will work with real buildings.
* Fixed buildings not rising from the ground.
* Fixed walls not spawning properly -> changed hull radius in .txt and with SetHullRadius later,
* Tested if Force Staff can be used on buildings -> It can be used but it does nothing, goes on cd.

TO DO:

- [ ] Fix units being stuck on the walls if the caster uses the wall on them - without using ResolveNPCPositions because buildings are units now as well.
- [ ] Fix lag caused on Watch Tower first cast.
- [ ] More tests.

DarkoniusXNG:
* Fixed collision bugs.
* Fixed lag on Watch Tower spawn - added precache.
* Buildings cannot be knocked back with abilities.
* Buildings will attack neutral creeps always now. No special aggro rules.
* Gave Walls 75% magic resistance.

@chrisinajar Tell me if I need to change something.

DarkoniusXNG:
You can merge this after I match HP, vision etc. to those in Baumi's Pull request.

PR #2402, opened 2018-09-13
Map Visual Update and Balancing

warpdragon:
This PR fixes the path overlays, removes 2 runes, and replaces some terrain with props. Let's get this in ASAP


warpdragon:
This is done except the patch seems to have broken the game... I can't even test the map because i get "WAIT" glitch before game and everything stops loading...

We need to fix what the patch broke before we can merge this, i think

warpdragon:
Abit Gray and Darkonius were so helpful. Dark eventually found missing entities referenced by Chris's new anti-camping feature. Thanks guys! This PR fixes the path overlays, removes 2 runes, and replaces some terrain with props. Let's get this in ASAP

PR #2403, opened 2018-09-17
Minimaps

Satsaa:
Updates the main minimap to match Wrapdagons pr. (Merge this after it)
10v10 minimap updated too but it was not included in Wraps pr so it matches the v4.0.1 map.


warpdragon:
man i fuckin love these

PR #2412, opened 2018-09-19
Fix minimaps and update 10v10 map

Satsaa:
Forgot to fix some issues in last pr..
Also makes to 10v10 vmap the same as the others. Should be okay as they have been identical in previous patches?


content/materials/overviews/10v10.vmat
```diff
@@ -2,5 +2,5 @@
 {
 	"Shader"	"ui.vfx"
 	"F_TRANSLUCENT"	"0"
-	"Texture"	"materials/overviews/10v10.tga"
+	"Texture"	"materials/overviews/oaa.tga"
```
chrisinajar: 10v10 has a different minimap since the map is shaped differently

Satsaa: This pull makes 10v10.vmap the same as others.
Looked in to "Add borders around the map to prevent fountain camping #2399 " and seemed they (vmaps) were identical in size and previews on asset browser, so i thought its fine.

PR #2416, opened 2018-09-29
Update some components to be self-registering

chrisinajar:
also delete some dead code..


DarkoniusXNG:
delete modmaker too, maybe?

Issue #2341, opened 2018-08-01
Sohei's W on allies talent breaks the ability


AbitTheGray:
Sohei's W (shield) can be also cast on allies when you get the talent for it. If you choose the other one but get this one with aghanim, the ability breaks and not allow you to cast it until the aghanim is removed from inventory.

DarkoniusXNG:
This can be closed.

PR #2420, opened 2018-09-30
Fixing Sohei bugs and adding QoL changes

DarkoniusXNG:
Making Sohei great again :D

Objectives:
- [x] Fix Sohei's W percentage heal (add heal from all damage instances, not only from last damage instance)
- [x] Fix https://github.com/OpenAngelArena/oaa/issues/2341
- [x] Fix "cannot stop Sohei's ult with silence" bug
- [x] (Optional) Make Sohei's E (momentum) toggleable.
- [x] Fix other Sohei bugs if possible.

* Cause of the second bug: Yahnich's function HasTalent doesn't work with talents that are added with Aghanim.


DarkoniusXNG:
I finished this. Beside fixing bugs I did the following:
1) Readded tooltips for modifiers (buffs).
2) Added sound to Wholeness of Body.
3) Made Momentum toggleable. Automatically Toggled ON when learned and while using Flurry of Blows.
4) Flurry of Blows now gives unobstructed vision in the effect radius. This will enable Flurry of Blows to actually work if you cast the spell on the high ground while you are on the low ground.

PR #2432, opened 2018-10-23
Experience Rubberband and Bounty Runes

DarkoniusXNG:
Goals of this PR:
- [x] Actually activate dota vanilla experience gain and rubberband effect when you kill a hero. 

Explanation: This is not active right now because API function SetUseCustomHeroXPValues is set to true. Barebones has a wrong description in settings.lua. SetUseCustomHeroXPValues allows heroes in the map to give a specific amount of XP when killed. 

- [x] Fix Bounty runes and ~~normal runes~~ (~~spawn times~~, gold they give, experience they give etc.)

P.S. If xp gained by killing a hero is too high or too low we should use custom experience system. For that we would need `hero:SetDeathXP(xp_value)` and/or `hero:AddExperience(float amount, int nReason, bool bApplyBotDifficultyScaling, bool bIncrementTotal)`. Also we need to make sure to not make the same bug like LOD with level 100 xP


DarkoniusXNG:
There are 2 bugs with runes currently: 
1) bug that hides runes that are not picked up on time.
2) If runes are picked few seconds before next spawn, new runes will not unhide on spawn location.

DarkoniusXNG:
I finished this. I didn't fix rune spawn timings. But I made it so runes don't spawn before the first duel and that's something. 
* Rune spawn interval is changed too -> they spawn every 2 minutes instead of every minute. But not at 2:00, 4:00, ... but at random times (depends on hero ban/pick screen duration).
* I gave up on the idea to hide runes in the corner or underground and then reveal them when its their time to be spawned. There were bugs I mentioned above and I was out of ideas.
* Added a template code for custom runes,
* Arcane runes still don't spawn unfortunately. RuneSpawnFilter doesn't work for me for some reason so I can't fix this issue.. I asked people on moddota, They don't know what's the issue. Arcane runes will spawn if we use dota default rune system, but then bounties will spawn every 5 minutes.
* Bounties now give **more gold** and **xp**. Formula is defined in Baumi's document.

PR #2436, opened 2018-10-25
creep bounties

l34Um1:


Haganeko:
Repeating the comment from #2435:
You changed both gold and XP values. Also, this is changing the base instead of the multiplier, so it would be better to change the values at creep_types.lua instead so the multiplier at minute 0 stays equal to 1.

chrisinajar:
this is easier for testing though... if we like it we'll probably revert and then set the base stats correctly, it's just a lot of data entry for global % based scaling

PR #2443, opened 2018-11-04
Fix drums/lucience/sonic/S&Y upgrade trees

chrisinajar:
currently you can't upgrade back into the items you upgrade from in many cases, such as lucience out of S&Y or converting sonic back into anything other than sonic...


DarkoniusXNG:
I noticed that for Martyr's Mail, you can't upgrade Assault into Martyr's Mail.

PR #2445, opened 2018-11-19
7.20 patch parity (WIP)

DarkoniusXNG:


DarkoniusXNG:
I will probably finish tomorrow.

DarkoniusXNG:
I finished items. 
* All items with hp and mp regen are updated (rebalanced). 
* Farming boots are updated (percentage based ms bonus). Greater Phase Boots and Greater Power Treads both give dmg and attack speed (balance it later if needed). Greater Phase Boots still have an active but the phase buff only lasts for 0.5 second. New 7.20 passive phase effect works normally.
* Fixed all? errors with items and fixed item costs.
I didn't fix tooltips, but neither has Valve.

DarkoniusXNG:
126 more files left to do ...
After testing Dazzle: Bad Juju is OP. I nerfed perma-grave though.
After testing Drow: Marksmanship doesn't one-shot bosses.

DarkoniusXNG:
100 files left ...
* Made Beastmaster Call of the Wild Boar and Hawk abilities completely custom. Hawks are not invisible but controllable. Talent that reduces cd of Hawk ability currently doesn't work.
* Didn't add new Faceless Void Time Lock, because you can perma-bash people with that shit with just one hit.
* Huskar new ability is awesome.
* Lich new abilities are awesome.

PR #2446, opened 2018-11-22
Add hidden armor modification for equivalency with old armor formula

Trildar:
Shoutout to DG for helping me be less of a complete noob on the armor modifier part.


DarkoniusXNG:
Nice

PR #2457, opened 2018-12-13
Attempt to reduce Lag

DarkoniusXNG:


chrisinajar:
Does this make illusions super tanky?

DarkoniusXNG:
Illusions don't benefit from armor.

chrisinajar:
oh okay, makes sense. it does make creeps just a little tankier though right?

DarkoniusXNG:
Yeah, but not bosses. And creeps can never reach 225 armor afaik.

PR #2461, opened 2018-12-22
OD fix

DarkoniusXNG:
Needs testing on dedicated server.


game/resource/English/ability/units/tooltip_oaa_arcane_orb.txt
```diff
@@ -2,11 +2,11 @@
 // Outworld Devourer Arcane Orb (OAA rework)
 //=============================================================================
 "DOTA_Tooltip_ability_obsidian_destroyer_arcane_orb_oaa"                          "#{DOTA_Tooltip_ability_obsidian_destroyer_arcane_orb}"
-"DOTA_Tooltip_ability_obsidian_destroyer_arcane_orb_oaa_Description"              "<font color='#FFA500'>Adds extra magical damage</font> to Outworld Devourer's attacks, based on his remaining mana pool. Steals %int_steal% intelligence for %int_steal_duration% seconds per hit when attacking an enemy hero. Arcane Orb also does bonus damage to summoned units and illusions. Arcane Orb damage splashes in a small area."
+"DOTA_Tooltip_ability_obsidian_destroyer_arcane_orb_oaa_Description"              "<font color='#FFA500'>Adds extra magical damage</font> to Outworld Devourer's attacks, based on his remaining mana pool. Steals %int_steal% intelligence per hit when attacking an enemy hero for the duration. Arcane Orb damage splashes in a small area."
```
Trildar: Why did you take out the duration?

game/scripts/vscripts/abilities/oaa_arcane_orb.lua
```diff
@@ -31,175 +43,209 @@ function modifier_oaa_arcane_orb:RemoveOnDeath()
...
-  parent:ChangeAttackProjectile()
+  if target:IsNull() then
+    return
+  end
+
+  -- This happens only when Arcane Orb is cast manually.
+  self.manual_cast = true
+
+  if ability:IsOwnersManaEnough() and ability:IsCooldownReady() and (not parent:IsSilenced()) and (not target:IsMagicImmune()) then
+    parent:SetRangedProjectileName("particles/units/heroes/hero_obsidian_destroyer/obsidian_destroyer_arcane_orb.vpcf")
```
Trildar: I think OnAttack was supposed to change the projectile back to normal. Cause this event fires just after the projectile already flies.

game/scripts/vscripts/abilities/oaa_arcane_orb.lua
```diff
@@ -31,175 +43,209 @@ function modifier_oaa_arcane_orb:RemoveOnDeath()
...
+  end
 
-  if keys.attacker ~= parent or not self.procRecords[attackRecord] or not keys.process_procs then
+  if target:IsNull() then
     return
   end
 
-  local bonusDamage = parent:GetMana() * ability:GetSpecialValueFor("mana_pool_damage_pct") / 100
-  self.procRecords[attackRecord] = nil
+  if ability:IsOwnersManaEnough() and ability:IsCooldownReady() and (not parent:IsSilenced()) and (not target:IsMagicImmune()) then
```
Trildar: There's a reason for all the proc record stuff. This stuff shouldn't be getting checked here. What happens if the caster runs out of Mana while the attack is flying? Or if autocast is turned on while a normal attack is flying? Only magic immunity is properly relevant.

game/scripts/vscripts/abilities/oaa_arcane_orb.lua
```diff
@@ -31,175 +43,209 @@ function modifier_oaa_arcane_orb:RemoveOnDeath()
...
+  if event.attacker ~= parent then
     return
   end
 
-  local ability = self:GetAbility()
-  local target = keys.target
-  -- Wrap in function to defer evaluation
-  local function autocast()
-    return (
-      target.GetUnitName and -- Check for existence of GetUnitName method to determine if target is a unit
```
Trildar: This check is here for a reason. All this stuff runs when you're attacking items too, which shouldn't use Arcane Orb attacks.

game/scripts/vscripts/abilities/oaa_arcane_orb.lua
Trildar: That's really late though. I think OnAttackFinished might work?

game/scripts/vscripts/abilities/oaa_arcane_orb.lua
DarkoniusXNG: Mana is removed **when the projectile lands**. Similar behavior to one before patch 7.14.
If autocast is turned on while normal attack is flying, arcane orb will be activated, bonus damage will be dealt and mana will be spent.
If caster runs out of Mana while the attack is flying, arcane orb will not do bonus damage, because IsOwnersManaEnough() will be false.

game/scripts/vscripts/abilities/oaa_arcane_orb.lua
DarkoniusXNG: I forgot about attacking items. I will add it.

game/scripts/vscripts/abilities/oaa_arcane_orb.lua
Trildar: I know about the mana being removed when it lands. I thought it was dumb behaviour when I found out, and so didn't bother matching it.
Does vanilla seriously just do all cast checks when it lands though? As in proccing when you toggle autocast while in flight? That's super jank.

game/scripts/vscripts/abilities/oaa_arcane_orb.lua
DarkoniusXNG: Why make it complicated? Its not that late.

game/scripts/vscripts/abilities/oaa_arcane_orb.lua
DarkoniusXNG: After patch 7.14. they made it so mana is removed after launching, instead of after landing. And that was a nerf to OD. Right now I did not bother with vanilla matching, I wanted to make it simple and working. If current behavior (with my implementation) is irritating you that match, I will try to match vanilla. But imo ''what is not broken should not be fixed''.

game/scripts/vscripts/abilities/oaa_arcane_orb.lua
Trildar: It can be really late though (think TPs), and I think in the case of misses or disjointed attacks, it'll just never even fire.

DarkoniusXNG: Yeah you are right. OnAttackFinished will work.

DarkoniusXNG:
Did I forget something else?

game/scripts/vscripts/abilities/oaa_arcane_orb.lua
```diff
@@ -31,175 +43,227 @@ function modifier_oaa_arcane_orb:RemoveOnDeath()
...
-  if parent:GetCurrentActiveAbility() ~= ability and not autocast() then
+  if target:IsNull() then
     return
   end
 
-  -- Add modifier to change attack sound
-  parent:AddNewModifier(parent, ability, "modifier_oaa_arcane_orb_sound", {})
-  -- Set projectile
-  parent:ChangeAttackProjectile()
+  -- Check for existence of GetUnitName method to determine if target is a unit
```
Trildar: Maybe clarify this with "(and not, e.g., an item)

DarkoniusXNG: I added more comments to clarify. I wonder what other **attackable** entities beside items don't have a GetUnitName method.

PR #2463, opened 2018-12-30
Add pre-game time to sync rune timers

chrisinajar:


game/scripts/vscripts/settings.lua
```diff
@@ -259,7 +259,7 @@ STARTING_GOLD = 825                     -- How much starting gold should we give
 DISABLE_DAY_NIGHT_CYCLE = false         -- Should we disable the day night cycle from naturally occurring? (Manual adjustment still possible)
 DISABLE_KILLING_SPREE_ANNOUNCER = false -- Should we disable the killing spree announcer?
 DISABLE_STICKY_ITEM = false             -- Should we disable the sticky item button in the quick buy area?
-SKIP_TEAM_SETUP = false                  -- Should we skip the team setup entirely?
+SKIP_TEAM_SETUP = false and IsInToolsMode()       -- Should we skip the team setup entirely?
```
DarkoniusXNG: It should be: `false or IsInToolsMode()` right? or am I dumb?

Trildar: Yeah, `false and ...` will just always be false. Though there isn't really any point to `false or` either. Might as well be just `IsInToolsMode()`
I guess maybe this was added to safeguard against accidentally publishing a version with it set to true, but still default to not skipping even in tools mode.

PR #2476, opened 2019-01-10
Alchemist spider egg/ball fix

DarkoniusXNG:
This will hopefully fix the server crashes with spider eggs.


chrisinajar:
it works! :D

PR #2482, opened 2019-01-23
Fix Captains Mode not being able to select hero at the end

chrisinajar:


chrisinajar:
thanks blu3 and max for reminding me regularly xD

PR #2485, opened 2019-01-25
Balance part 2

DarkoniusXNG:
1) KOTL hp talent buff.
2) Puck turn rate improvement and Dream Coil cd at OAA levels buffed
3) Timbersaw talent from +150 mana to +150 hp.
4) Warlock Chaotic Offering fixes and buffs.
5) Weaver bonus dmg talent nerf because its a lvl 10 talent.
6) Riki talent rework. Removed Cloak and Dagger Doesn't Reveal talent. Nerfed bonus cast range on blink strike.
7) Minor cleanup of some KV files.


DarkoniusXNG:
* Reverted Weaver nerf.
* Buffed Warlock ult.
* Buffed Crystal Maiden's ult armor bonus (made it scale).

If there is something you don't agree with, let me know.

PR #2487, opened 2019-01-26
Fix Boots of Travel and TP Scroll sharing the cooldown

DarkoniusXNG:
This PR should fix this issue.


DarkoniusXNG:
Teleport Scroll and Boots of Travel don't share cd anymore after 7.20. There is a 10 seconds penalty though in boots of travel code: https://github.com/OpenAngelArena/oaa/blob/master/game/scripts/npc/items/item_travel_boots.txt#L84

PR #2494, opened 2019-02-11
Silencer Glaives of Wisdom fix attempt

DarkoniusXNG:
Needs testing on dedicated servers.

* If mana is not spent, then its because of line 139: `ability:CastFilterResultTarget(target)` is probably nil. 
* If damage doesn't work in OnAttackLanded then probably `event.record` is nil.


game/scripts/vscripts/abilities/oaa_glaives_of_wisdom.lua
```diff
@@ -155,25 +203,30 @@ function modifier_oaa_glaives_of_wisdom:OnAttackLanded(keys)
...
-      damage = bonusDamage,
-      damage_type = ability:GetAbilityDamageType(),
-      ability = ability
-    }
+    local damageTable = {}
+    damageTable.victim = target
+    damageTable.attacker = parent
+    damageTable.damage = bonusDamage
+    damageTable.damage_type = ability:GetAbilityDamageType()
+    damageTable.ability = ability
```
Trildar: I really just have to ask. Why would you write the damageTable this way instead of what it was before?

DarkoniusXNG: Because I change the damage type when target has spell immunity.

SquawkyArctangent: That's not exactly necessary, though. Something like:

```
local table = {
    x = y,
}

if a >= b then
    table.x = z
end
```

works with zero issue.

DarkoniusXNG: I didn't know that.

PR #2495, opened 2019-02-12
Fix some Capture Points abuse

DarkoniusXNG:
* Prevent hero creep units (like Spirit bear) to count as a hero for capture point.
I also want to check if the hero has wraith form buff, to remove him from the table.


DarkoniusXNG:
Travil still broken. PepeHands.

chrisinajar:
im going to disable the wraith king thing until after the tournament because it's a significant nerf, also not sure if it makes sense

DarkoniusXNG:
Wraith King aghs ult capturing while dead is bullshit imo.

PR #2497, opened 2019-02-15
Balance

chrisinajar:


DarkoniusXNG:
Travis working again?

chrisinajar:
apparently? i didn't change anything

lovesmuffins:
with the travel buffs does this mean my changes will still be put in or should I scale it back now.

lovesmuffins:
some of these changes also interfere with shannon's proposed item changes. You pushed an update to arcanes rather than just pushing shannon's arcane buff (and their practically the same buff to gold)

PR #2499, opened 2019-02-21
Bloodstone fix to not stack

DarkoniusXNG:
These changes (hopefully) nerf cancer Storm Spirit ult and prevents Storm Spirit to zip from one side of the map to the other with infinite mana even if he has multiple bloodstones.

These nerfs are justified because OAA map is smaller than dota map, and because in OAA there is an item that gives more mana cost reduction (Yasha and Kaya 4/5).
Max mana cost reduction in normal dota is 16% (without CM aura and without Arcane Rune).
In OAA its 42%. 
Math behind the nerf: 42/16=2.625; 0.7 x 2.625 = 1.8375;


game/scripts/npc/items/item_bloodstone.txt
```diff
@@ -85,7 +85,7 @@
       "05"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "mana_regen_multiplier"                           "150 200 350 500 750"
+        "mana_regen_multiplier"                           "200 250 300 350 400"
```
lovesmuffins: I think while storm nerf's are justified nerfing bloodstone could hurt some other heroes. If we're going to be conservative before tournament I probably wouldn't nerf late game bloodstone by over 300%. I think 200 275 350 425 500 would be a fairer nerf. I also don't think storm's mana is entirely the issue and would probably increase the unit interval for his talent as well from like 300 to 500 or 400.

lovesmuffins: or 150 250 350 450 550

POCKETMONEY99: I dont think any other heroes actually have mana issues after a bloodstone apart from storm

lovesmuffins: does chatterjee still have mana issues or did they fix her up a bit?

DarkoniusXNG: The nerf on bloodstone is harsh because mana regen amp stacks. And I don't want BTS to see Storm Spirit with multiple bloodstones on their stream. Other heroes don't suffer from the nerf (not even Chatterjee).

POCKETMONEY99: I honestly think Storm Spirit should just be removed from the game, along with Tinker. These heroes are pretty much 100% balanced around mana pool and I personally don't think there is much way for these heroes to be balanced without having unique limiters on their heroes.

POCKETMONEY99: One hero shouldn't have to be the reason why bloodstone has to be nerfed. But Because this probably won't happen I have some other suggestions:

- Change Storm spirits level 25 talent 550 auto remnant to another talent (maybe +40 all stats or something.) 

- Make Bloodstones Not stack with eachother (Valve may do this but stacking bloodstones is impractical in normal dota). Storm spirit can get +300 Mana regen in normal dota just by stacking bloodstones, and that is enough for it to be stupid. And In OAA farming up 4/5 bloodstones is easy.

DarkoniusXNG: Ok, I forgot to say this: This PR is just a temporary solution.

POCKETMONEY99: ok

DarkoniusXNG:
* Reverted bloodstone nerf.
* Made mana regen amp not stack (didnt test).
* Toned down Ball Lightning nerf.

lovesmuffins:
While I love that it no longer stacks, I find that maybe we should just have that be the only change and revert the lightning ball nerf as well to see where storm is at. Then change him or leave him accordingly. Without the stacking of it perhaps storm will be back to the garbage and then nerfing his ult furthermore would make him more garbage. He was hardly used when bloodstone's passive was broken (despite you being able to stack the mana regen back then). Just food for thought

DarkoniusXNG:
Like I said in the first post: nerf to Ball lightning is justified because OAA map is smaller than dota map, and because in OAA you can have more than 16% mana cost reduction.

PR #2500, opened 2019-02-25
Timbersaw Whirling Death rebalance

DarkoniusXNG:
This is a nerf to Timbersaw because he terrorizes pubs if played correctly. Merge after minor/major.

In 7.20 Whirling Death damage type was changed to be always pure. And bonus damage per tree cut was added. We didn't change the base damage so its a justified nerf.


lovesmuffins:
you could just make extra tree dmg 0?

POCKETMONEY99:
> 
> 
> you could just make extra tree dmg 0?



> 
> 
> you could just make extra tree dmg 0?

Whirling deaths damage itself is just too high for a pure damage nuke (Combinded with his other spells). I think these changes are fine but I would keep a close eye on timber because 7.20 is the first time hes been good in OAA to my knowledge

DarkoniusXNG:
He was fine before 7.20 (good pick but not OP). Buff to Q made him what he is today.

PR #2508, opened 2019-03-11
Queen of Pain Buffs

POCKETMONEY99:
Buffs to right click QoP and Sonic Wave.


DarkoniusXNG:
My suggestion is either buff the talent or BAT, not both. BAT was already improved few patches ago...

DarkoniusXNG:
This is fine.

PR #2512, opened 2019-03-12
Small buffs

TheUnholyWrath:
Arcanes- movespeed equal to guardian greaves
Ac- slight aura buff
Dragon Lance- better stat scaling
Silver Edge- stat scaling

### DIFF
--- a/game/scripts/npc/items/item_silver_edge.txt
+++ b/game/scripts/npc/items/item_silver_edge.txt
@@ -66,7 +66,7 @@
       "03"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "bonus_all_stats"                                 "15 18 21 25 28"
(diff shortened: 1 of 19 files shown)


game/scripts/npc/items/item_silver_edge.txt
```diff
@@ -66,7 +66,7 @@
       "03"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "bonus_all_stats"                                 "15 18 21 25 28"
+        "bonus_all_stats"                                 "15 19 24 30 38"
```
POCKETMONEY99: I feel like the attack speed and stats should scale like 
Attack speed "30 35 40 45 50"
Then the stats scale like "15 20 25 30 35" (I have no idea if they ever did but this item isn't a terrible stat item in dota and it shouldn't be in OAA either)

PR #2515, opened 2019-03-13
Ogre Magi Multicast Fix

POCKETMONEY99:
Fixed an Issue causing Ogre Magi to multicast less with the level 4 and 5 ultimate.


lovesmuffins:
how does lowering the tooltip affect the chance at all?

POCKETMONEY99:
> 
> 
> how does lowering the tooltip affect the chance at all?

oops, fixed it agian

DarkoniusXNG:
Now its ok.

PR #2519, opened 2019-03-13
Doom Buffs

POCKETMONEY99:
Reduced the cooldown of scorched earth to match dota levels. Then I buffed scorched earth slightly at OAA levels. Also reduced the cooldown of Doom.


DarkoniusXNG:
We tested this in the test client. Its fine.

PR #2527, opened 2019-03-15
Legion Commander Buffs

POCKETMONEY99:
Buffed duel duration and damage recieved, as legion has a very difficult time getting duel vs midly coordinated teams. This is because peoples items proggess faster then legion commander gets duel damage. 

Buffed overwhelming odds slightly to help with early camp clear etc. 

Also buffed  her agility gain, allowing this hero to secure more duel kills.


DarkoniusXNG:
I don't like AS talent buff and AGI gain buff.

DarkoniusXNG:
Can you not buff Overwhelming Odds? Its not a bad ability.

PR #2528, opened 2019-03-15
Mirana Buffs

POCKETMONEY99:
Reduced the charge restore time on leap. Decreased the cooldown and mana cost on starstorm to help her clear camps faster. Also increased her base agility by 4 (from 18 to 22)


DarkoniusXNG:
Mirana is one of the heroes that have highest AGI gain. Here is the list: Terrorblade (+4.8 per level), Morphling (+4.3), Medusa (+3.9), Vengeful (+3.8), Mirana (+3.7). So I don't like your buff to AGI gain. Buffing starting STR or starting AGI on the other hand I would approve.

PR #2530, opened 2019-03-16
Dark Willow Buffs

POCKETMONEY99:
Reduced the Cooldown on shadow realm. 
Increased the damage that bedlam does. 
Changed her weak armour talent into a spell amplification talent. 
Increased her agility gain.


lovesmuffins:
Why does willow needs buffs exactly?

PR #2534, opened 2019-03-18
Talents refactor

DarkoniusXNG:


DarkoniusXNG:
Only one balance change: Lone Druid Spirit Bear BAT talent from -0.5 to -0.3

More changes to non-meta heroes next time.

DarkoniusXNG:
Refactored talent files in a following way:
* Instead of one .txt file per talent, now its one .txt file per hero and all talents of that hero are in that .txt file.
* Moved all talents (Sohei and Chatterjee talents) into talents subfolder. 
* Added bunch of talents for some non-meta heroes so they can be changed more easily in the future.

Downside: When Valve changes talent values, we will need to change them manually.
Problem: Travis doesn't react to talent changes.

PR #2535, opened 2019-03-19
Add better XP gain on Hero Kill

DarkoniusXNG:
XP system is now similar to dota system with reduced values.

Change from constant _62/number of killers_ to **(20 + XPStreakBonus + 0.07*DyingHeroExperience)/number of killers**.

**XPStreakBonus** varies from 60 to 420. 60 for killing spree, 420 for beyond godlike.

Also exposed these constants in settings.lua so they can be changed more easily in the future.


lovesmuffins:
Whilst I understand the complaints from "top" teams to not add it to the major, I feel like if we can test this out and see if it makes the game pace or comeback mechanic better to improve the spectator feel of oaa. We didn't do any huge changes from minor to major yet so I feel like if there's any 1 major change that should be put in it would be this. Just my cup of joe though.

DarkoniusXNG:
Creeps XP bounties needs reducing too. That's why I think this fixed XP gain on hero kill would lead to too much experience on the map.

lovesmuffins:
> Creeps XP bounties needs reducing too. That's why I think this fixed XP gain on hero kill would result in too much experience on the map.

You are right in that regard perhaps we should tinker with it in oaa test client

DarkoniusXNG:
Haganeko is working on it.

DarkoniusXNG:
This is good enough for merging.

Issue #2540, opened 2019-04-03
Boots movement speed stacks


ScrubbyNubby101:
![dsd](https://user-images.githubusercontent.com/34697728/55518056-380c5380-5638-11e9-97a7-d71c06b4c8cc.png)

DarkoniusXNG:
This can be closed now.

PR #2541, opened 2019-04-04
Boots MS should not stack

DarkoniusXNG:
https://github.com/OpenAngelArena/oaa/issues/2540

Didn't test if this fixes it. Will test it tomorrow.


DarkoniusXNG:
After testing: 
* Greater Phase Boots (passive ms bonus) still stack with any other upgraded boots but doesn't stack with non-upgraded (vanilla boots).
* Other upgraded boots stack with non-upgraded (vanilla) boots but don't stack with each other.

Issue #2411, opened 2018-09-18
Implement new creep camps


l34Um1:
# New Creep Camps

We want to introduce more interesting creep camps. Below you see a map. Each camp will be replaced as is marked on the map and described below.

![creep camps](https://user-images.githubusercontent.com/24982519/45715975-da318600-bb95-11e8-92af-7fe56beaba02.jpg)

The new creep camps each have at least one unit each with two abilities. These creeps will use the abilities described randomly while they are fighting an enemy.

Abilities that will have to be newly created are marked in **bold**.

## 3 Corner camps
- Camp #1 
- Black ancient dragon
- Ability 1: The fireball ability of the Ancient Black Dragon as described [here](https://dota2.gamepedia.com/Ancient_Black_Dragon).
- **Ability 2: Armor Aura. Provides all allies in a 900 radius with additional armor. Unlike the Black Ancient Dragon armor aura it doesn't stack.** 
- Spawn quantity: 1 
- Max quantity: 1

### 2 base camps are the same
- Camp #2 
- uses ghost model
- **Ability 1: Passive. On attack reduces all healing the target receives by 40% for 5s.**
- **Ability 2: Active. Slows the attack speed of enemies in a small AOE for a few seconds.**
- Spawn quantity: 3
- Max quantity: 6

### top near the bosses is different

- Camp #3 
- Always spawns as 1 big centaur and 2 small ones
- Ability 1: Active: Centaur stun as described [here](https://dota2.gamepedia.com/Centaur_Camp).
- Ability 2: Passive: Provides allies in a 900 radius with bonus attack speed.
- Only the big centaur has abilities.
- Spawn quantity: 3 
- Max quantity: 6

### 1 bottom near the bosses boss

- Camp #4
- Blueberrys
- Ability 1: Hellbear clap as described [here](https://dota2.gamepedia.com/Hellbear_Camp).
- Ability 2: Provides allies in a 900 radius with magic resistance.
- Spawn quantity: 3
- max quantity: 3

### 2 near the base exits

- Camp #5
- give most XP, but that xp is split up into both mud golems and smaller golems
- mud golems but without the insane magic resistance
- Ability 1: Active. Hurl Boulder as described [here](https://dota2.gamepedia.com/Golem_Camp).
- Ability 2: Passive. Shard Split as described [here](https://dota2.gamepedia.com/Golem_Camp).
- Spawn quantity: 4
- Max quantity: 8

### 1 near the center highground

- Camp #6
[...]

PR #2548, opened 2019-04-08
New Neutral Creeps

DarkoniusXNG:


DarkoniusXNG:
https://github.com/OpenAngelArena/oaa/issues/2411
I didnt do:
1) Special rules where each creep spawn and how many are allowed to spawn in the camp. Current system is fine if you ask me.
2) AI for creeps with abilities. AI with thinkers could cause lag and affect performance. Boss AI already lags a bit. I was trying to find a way to do AI without thinkers somehow but I gave up.
3) new Ancients (Ogres that smash), I don't like the idea that there are only 2 Ancients so I gave up on this. But I like the idea of making Ancients harder to kill.
4) Animation and particle for Mud Golem Shard Split.
5) Sound for Kobold Foreman War Cry.

PR #2552, opened 2019-04-10
Post april major

l34Um1:
The map currently breaks the game and I don't know why. I assume one of the zones is wrong. Need help with it.


l34Um1:
So travis is telling me that all sorts of things are wrong but none of them are files I actually touched so I dunno what to do about it.

SquawkyArctangent:
it's because we haven't caught up to vanilla dota yet and those values haven't beem marked as changed for oaa, dark has a pull request that fixes that

l34Um1:
Ok that's good. Map still breaks the game though lol

lovesmuffins:
is this a new map or old map?

PR #2554, opened 2019-04-11
Removed break distance from Static Grip.

SquawkyArctangent:
Was a pretty niche mechanic considering how neither hero could move on their own accord, plus now cast range bonuses won't feel as bad on her.


lovesmuffins:
With this change i might suggest to make the dispel type basic so that supports can more easily free their allies (since they can't use movement abilities or items to free them)

PR #2558, opened 2019-04-12
item buffs and nerfs

TheUnholyWrath:
far sight cd buff and aoe scaling

butterfly worse level 2, more consistent scaling

### DIFF
--- a/game/scripts/npc/items/custom/item_far_sight.txt
+++ b/game/scripts/npc/items/custom/item_far_sight.txt
@@ -48,11 +48,11 @@
 
     // Stats
     //-------------------------------------------------------------------------------------------------------------
-    "AbilityCooldown"                                     "90.0 75.0 60.0 45.0"
+    "AbilityCooldown"                                     "60.0 50.0 40.0 30.0"
     "AbilitySharedCooldown"                               "farsight"
     "AbilityCastRange"                                    "0"
     "AbilityCastPoint"                                    "0.0"
-    "AoERadius"                                           "750"
+    "AoERadius"                                           "750 800 850 900"
 
     // Spicy Parameters
     //-------------------------------------------------------------------------------------------------------------
@@ -107,7 +107,7 @@
       "06"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "reveal_radius"                                   "750"
+        "reveal_radius"                                   "750 800 850 900"
       }
       "07"
       {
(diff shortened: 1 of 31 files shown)


lovesmuffins:
you've kinda nerfed butterfly into the ground...
I agree it needs better scaling but it shouldn't be nerfed as well. It's an item that isn't really picked up due to MKB countering it's unique component. Most people either go into Yasha Sange/Kaya or stay diffusal. The one reason people may of picked it up was for the evasion vs bosses but you're also reducing that. I don't know how I feel about these changes overall.

TheUnholyWrath:
> you've kinda nerfed butterfly into the ground...
> I agree it needs better scaling but it shouldn't be nerfed as well. It's an item that isn't really picked up due to MKB countering it's unique component. Most people either go into Yasha Sange/Kaya or stay diffusal. The one reason people may of picked it up was for the evasion vs bosses but you're also reducing that. I don't know how I feel about these changes overall.

seems like my title is pretty misleading, initially there were butterfly buffs. you're totally right, when i'm back i'll change it to something more reasonable

TheUnholyWrath:
far sight nothing changed
butterfly might be more reasonable now, probably still needs fiddling
mkb probably doesnt need a nerf
idea for a glimmer cape change
yasha variants should have actual movement speed, this is less than actual

pretty happy to just remove anything too unreasonable/insignificant/completely off the mark. mostly just feelers for feedback

lovesmuffins:
I'd actually really like it if MKB's chance would be the numbers getting the nerf. To make it so evasion isn't completely useless when your opponents first get the item.

Something like 55/60/65/70/75?

TheUnholyWrath:
> I'd actually really like it if MKB's chance would be the numbers getting the nerf. To make it so evasion isn't completely useless when your opponents first get the item.
> 
> Something like 55/60/65/70/75?

the damage on the procs would have to be fiddled with if you want to do that, almost had the exact opposite idea with though, want to make mkb a bit more based on the proc, i was thinking about scaling it when changing but something much more mild, not sure yet maybe i'm the way i'm thinking is the wrong direction to move mkb to

lovesmuffins:
As of right now, MKB is just a better version of Mjolnir, without the attack speed. I liked how you scaled the Mjolnir upwards and I think MKB should also start less to reflect that change. I don't think damage would have to be fiddled with that much as of right now it's powerful enough and your damage nerf could just be reverted.

l34Um1:
Can you please take out the movement speed changes? Then I'll merge this.

TheUnholyWrath:
yasha gives flat movement speed these changes are fine

PR #2559, opened 2019-04-13
Pipe aura buff

TheUnholyWrath:
might be too much, on pipe carrier nothing should change, team gets 8% more at max


POCKETMONEY99:
I would make the scaling "10 12 14 16 18". I don't want to see OAA enter a tank meta because every aura item is insane.

DarkoniusXNG:
Even a small change will remind people that this item exists and that Skywrath Mage is not an overpowered hero. Glimmer Cape is worse than Pipe tbh.

PR #2560, opened 2019-04-13
Rehaul boss regeneration and implement boss degeneration

SquawkyArctangent:
Replaces the random Heart each boss had with a new ability that does the same thing, but pauses during duels ( though this can be disabled ) and also causes bosses to _lose_ 1% of their max health each second after being randomly damaged.

Also generated tooltips so Baumi doesn't have to mess around with node.js stuff.


lovesmuffins:
Inb4 zeus can solo any boss with nimbus

DarkoniusXNG:
@SquawkyArctangent Use editorconfig. It will make your life easier. And travis will yell less.
On topic: I think the new ability (boss_regen) should be hidden. And regen buff and regen debuff should be visible when they are active. Some bosses have a lot of active abilities so it will be less cluttered.

PR #2564, opened 2019-04-14
Io level 25 talent nerf

DarkoniusXNG:
With this Io (Wisp) will be unable to attack while disarmed, stunned, hexed or out of world (Disruption, Astral Imprisonment).


POCKETMONEY99:
I can't see this changing much. (It will just make io players buy bkb more which they definitely have a slot for) The best way to deal with the io combo is to disable the tethered target. 
The issue with this talent is that your combined damage output is just to high. Especially with any split shot mechanics (Drow, Gyro, Clinkz and Medusa)
I would remove the talent or add an internal cooldown to it.

lovesmuffins:
You could also make IO disarmed permanently and ONLY attacking when the tethered target attacks.

Larenall:
Just remove this cancer talent pls, with thing like bubble orb its ez to not give a F about any disables,disarms and etc.

DarkoniusXNG:
1) I proposed replacing this talent with Tether Stuns. It was rejected. And for a reason, if this talent is removed, Io would just be a support whole game and boring imo.
2) I proposed damage reduction based on distance (like reverse Enchantress ult - the closer are you to the attacked target the more dmg you do). Mixed feedback on this idea: some people said its interesting, but other people said its complicated and unintuitive. Also it adds extra rules, which is a bad game design.
3) Another idea is 50% attack dmg reduction while Io is tethered to an ally. This dmg reduction would apply only when you take lvl 25 talent. 
4) Bubble orb will always be a thing, maybe the item is the problem. Remove the item? Bubble orb is also broken on Nyx.

POCKETMONEY99:
People who enjoy io probably aren't playing it to go carry io. While a 50% Damage reduction would help I feel like this talent is either going to be completely overpowered or too bad to the point where you wouldn't go it anymore.

lovesmuffins:
I like my idea :C

l34Um1:
I just tested this and it doesn't seem to be working.

DarkoniusXNG:
Weird. I will test it again on my end.

PR #2572, opened 2019-04-23
Add Wandering boss

chrisinajar:


lovesmuffins:
the bosses already wander 🤣 

I hope with a wandering boss you also take the chance to actually fix non-wandering bosses aggro. Otherwise this change will only piss people off even more especially if this boss's aggro system is anything like what we have currently.

chrisinajar:
aggro system on this one is a lot better and a lot better tested :+1:

PR #2576, opened 2019-04-25
Luna Aghs Eciplse Rebalance + Talent Changes

POCKETMONEY99:
Increased the Beams on Eclipse at later levels. Additionally I lowered the interval on aghs to its original value. However the maximum beams per target with aghs is now limited per target. I also decreased the mana cost of Eclipse to buff right click carry luna.

Nerfed the 25 ministun talent to 0.1 seconds to prevent perma stunning allowing you to react. 

Heavily reduced the cast range on Aghs Eclipse to prevent you from ulting capture points and at the start of duels from the other side of the map. 

Buffed the level 20 all stats talent. 
lineararised the scaling on Lunar Blessing Bonus night vision at level 6. 

**Addtional Notes** I was considering changing the 25 ministun  talent to spell lifesteal because imo I think this talent is too strong as you can guanteed a aghs + refresher in OAA.


POCKETMONEY99:
I want to reduce the stun duration on the level 25 talent to 0.15 seconds

PR #2577, opened 2019-04-25
Slardar and Faceless Void Bash. Mini Rework

POCKETMONEY99:
Gave Faceless Void and Slardar Bash a 'mini rework'. I have changed the numbers on both of these spells heavily to make them more suitable to the heroes without having to rework either ability. 

For Faceless Void I went in the direction of changing timelock so it is a good enough DPS skill for him to carry with. However I had to sacrifice alot of the stun duration to allow this passive to scale with attack speed. I then increased the proc chance and damage output on time lock to compensate for this. 

For Slardar I tried to make his bash closer replicate how it is currently in OAA. I increased the Cooldown and the stun duration. I also increased the proc chance to 99%. This means that when the spell is off cooldown you are guanteed a bash. Meaning you can play slardar more simillar to how you do in dota. I also increased the damage of this spell. 

For Both heroes the only talent values I changed was the bash damage, increasing them both. 

**Tl;dr:** Void can now deal enough damage to be played as carry, however he can only ministun. 
Slardar bash is guanteed off cooldown and he can stun enemies for longer, however the cooldown is longer then before.


Larenall:
VERY interesting, absolutely love those changes but id give slardar`s bash a bit different numbers to prevent permabashing. Maybe 3 sec cd intstead of 2.

game/scripts/npc/abilities/faceless_void_time_lock.txt
```diff
@@ -24,7 +24,7 @@
       "01"
       {
         "var_type"                                        "FIELD_FLOAT"
-        "duration"                                        "1.0" //OAA, for picking screen tooltip
+        "duration"                                        "0.2" //OAA, for picking screen tooltip
```
lovesmuffins: I'd double this

game/scripts/npc/abilities/faceless_void_time_lock.txt
```diff
@@ -34,12 +34,12 @@
       "03"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "chance_pct"                                      "10 15 20 25 25 25" //OAA, for picking screen tooltip
+        "chance_pct"                                      "15 20 25 30 35 40" //OAA, for picking screen tooltip
```
lovesmuffins: I advocate for a starting percentage of 25%
25/25/25/25/30/35

game/scripts/npc/abilities/faceless_void_time_lock.txt
```diff
@@ -34,12 +34,12 @@
...
       }
       "04"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "bonus_damage"                                    "75 100 125 150 325 500"  //OAA, for picking screen tooltip
+        "bonus_damage"                                    "100 125 150 175 350 550"  //OAA, for picking screen tooltip
```
lovesmuffins: if you do double the duration of stun I would just revert these changes (not additional nerfs needed)

game/scripts/npc/abilities/faceless_void_time_lock_oaa.txt
```diff
@@ -41,18 +41,18 @@
       "03"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "chance_pct"                                      "10 15 20 25 25 25"
+        "chance_pct"                                      "15 20 25 30 35 40"
```
lovesmuffins: everything i said on the non_oaa version should apply here. Cooldown should be more than .2 (if u make it .4 stun duration i'd make the cooldown 1 second since you have to consider void can go octarine to reduce the cooldown of this spell.)

game/scripts/npc/abilities/slardar_bash_oaa.txt
```diff
@@ -34,18 +34,18 @@
       "01"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "chance"                                          "25"
+        "chance"                                          "95"
```
lovesmuffins: Make this like 65%.

game/scripts/npc/abilities/faceless_void_time_lock_oaa.txt
POCKETMONEY99: The whole point of the void bash changes is to make him actually get good damage from time lock so he can carry. Any cooldown around 1 second will ruin this.

game/scripts/npc/abilities/faceless_void_time_lock.txt
POCKETMONEY99: Ive heavily reduced bash duration to allow for voids bash cooldown to be extremely low so it can scale with attack speed. I want to leave this number in this ball park.

lovesmuffins: .2 isnt a value that matters. SR reduces this number even more so it's hardly a nerf at all. Also having the cooldown be the same allows void to exploit CDR

l34Um1:
These aren't good changes.

Regarding FV: I understand what you're trying to do but this wouldn't result in what you want it to result in. This is now just a better Juggernaut crit. Juggernaut crit is 136% avg dmg increase, this is 140% avg dmg increase. The 0.2s cd doesn't matter, the fastest attack rate that FV can achieve is 0.24s. This ability might as well not have a cooldown.
image
I also don't know why we're even changing this? Isn't FV fine?

Slardar: I don't hate the change but it's not clean. 99% chance? If 100% isn't possible rework the ability so it is.

DarkoniusXNG:
> ...and it can trigger attack modifiers...

In OAA Void's Bash is the old Bash. There are no secondary hits.

>I also don't know why we're even changing this? Isn't FV fine?

He is fine as a support/initiator. Pocket wanted to make Void more viable as a carry.

DarkoniusXNG:
And I already told Pocket I don't like his Void changes.

@POCKETMONEY99 For Slardar Bash, just put 100% bash chance (it is possible) and 5 or 6 seconds cd at all levels (to prevent permastun).

l34Um1:
> > ...and it can trigger attack modifiers...
> 
> In OAA Void's Bash is the old Bash. There are no secondary hits.

Oh yeah you're right, I assumed this was how Pocket had planned to implement the new bash but it's still using the old one. Then there is absolutely no point to this.

POCKETMONEY99:
The point is that both of these bash spells suck because they have a ~1.5 second cooldown. Both of these spells are meant to be good with attack speed but they aren't due to this astronomical cooldown.

POCKETMONEY99:
The slardar one is a bit more experimental but the void one would make timelock be a good DPS passive atleast, instead of being a budget skull basher.

PR #2580, opened 2019-04-26
Giant Form buff

TheUnholyWrath:
4 seconds level 4
2 seconds level 5
probably

attack rate down to 0.7

### DIFF
--- a/game/scripts/npc/items/custom/item_giant_form.txt
+++ b/game/scripts/npc/items/custom/item_giant_form.txt
@@ -92,7 +92,7 @@
       "06"
       {
         "var_type"                                        "FIELD_FLOAT"
-        "giant_attack_rate"                               "0.8"
+        "giant_attack_rate"                               "0.7 0.65"
       }
       "07"
       {
--- a/game/scripts/npc/items/custom/item_giant_form.txt
+++ b/game/scripts/npc/items/custom/item_giant_form.txt
@@ -92,7 +92,7 @@
       "06"
       {
         "var_type"                                        "FIELD_FLOAT"
-        "giant_attack_rate"                               "0.7 0.65"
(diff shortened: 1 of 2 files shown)


game/scripts/npc/items/custom/item_giant_form.txt
```diff
@@ -48,7 +48,7 @@
     "AbilityTextureName"                                  "custom/giant_form_1"
     "AbilityCastPoint"                                    "0.0"
     "AbilityManaCost"                                     "50"
-    "AbilityCooldown"                                     "6.0"
+    "AbilityCooldown"                                     "4.0"
```
POCKETMONEY99: Don't you have to write it like "4.0 2.0" Im not 100% sure though. Ask a real dev.

POCKETMONEY99:
This item needs a rework or removing imo. I currently don't see a middle ground where this item is balanaced in its current state.

TheUnholyWrath:
i think you're right i'll fix that, its scabos idea, thats what he wanted changed about it. if its still bad then no big deal life goes on, if its too good can always reel it back in a bit

lovesmuffins:
#LETMELEEMODIFIERSWORKONGIANTFORM
#IWANTMYECHOSABREGIANTFORMBUILDTOBEGOOD

SquawkyArctangent:
echo saber literally can't work with giant form on a fundamental level, because echo saber just gives you a lot of attack speed and giant form sets your attack speed to base

lovesmuffins:
> echo saber literally can't work with giant form on a fundamental level, because echo saber just gives you a lot of attack speed and giant form sets your attack speed to base
 

ye which sucks lol.
It should set your BAT to like 1.5 reduce your attack speed by 75% or something like that. (with a minimum attack speed of 150

PR #2582, opened 2019-04-26
Naga Siren, Tusk and Undying Buffs.

POCKETMONEY99:
Buffed Naga Sirens riptide as this heroes main issue is her damage output falls off extremely hard late game.

Set the mana cost on Tusk Tag team to 70 at all levels. This just makes the ability cleaner. 

Reduced the cooldown on Flesh Golem. This the main reason why undying popularity fell off heavily.


lovesmuffins:
tusk doesn't need any buffs at all

POCKETMONEY99:
For tusk its just making the spell look better. If a spell got set to 70 mana at all levels it should happen in OAA. I did the same for clockwerk and I will do the same for any hero I see with weird mana costs like 70/70/70/70/100/200 etc.

PR #2585, opened 2019-04-26
Bunch of boot changes.

SquawkyArctangent:
- Power Spark + Treads no longer have splash, but instead add bonus pure damage to attacks vs. neutrals. As always, Atan numbers are temporary.
- Tranquil Spark + Boots can no longer eat ancient creeps.
- Phase Spark + Boots now splashes attack damage, instead of using instant attacks. Still only applies to up to x units.
- Sparks can be built into any boot using a tier _1_ core. This is probably how all upgrades should work now ( since it's a newbie trap not to split cores, as that loses money for literally no benefit ), but if nothing else, you really shouldn't be able to waste a higher core on your first tier of boots.


POCKETMONEY99:
I love the all sparks being able to build into any boots. This will help out so many heroes.

Only critisim is that the only edge tranquil has over arcanes is being able to kill ancient creeps. (I'm assuming corner dragons will remain as ancient creeps.). Tranquil boots will probably be dead after this change because Greaves will just be better to have in your inventory.

game/scripts/vscripts/items/farming/greater_power_treads.lua
```diff
@@ -105,28 +105,30 @@ function modifier_item_greater_power_treads:OnCreated( event )
...
+    self.damageTable = { 
+      { self.creepDamageMelee, self.creepDamageMeleeIllusion, },
+      { self.creepDamageRanged, self.creepDamageRangedIllusion, },
+    }
+  end
 end
 
 --------------------------------------------------------------------------------
 
 function modifier_item_greater_power_treads:OnRefresh( event )
```
DarkoniusXNG: Afaik you can write  `modifier_item_greater_power_treads.OnRefresh =  modifier_item_greater_power_treads.OnCreated` instead of this. Example here: https://github.com/OpenAngelArena/oaa/blob/master/game/scripts/vscripts/abilities/oaa_glaives_of_wisdom.lua#L56

SquawkyArctangent: I'm aware. I just prefer this method since it makes it simpler to add OnCreated or OnRefresh-specific behavior afterwards should either become necessary.

l34Um1:
Tranquil change didn't work for me, I can still kill ancients.
Treads need updated tooltips.

Otherwise this all works and is good to go.

l34Um1:
All good to go now. Requires tooltips to be built.

lovesmuffins:
tbh I'd rather the tread passive be switched with phase, the cleave is unique to treads and allow more agi-type heroes (I'm thinking of AM in this case) would use.

SquawkyArctangent:
consider, however: bonus damage scales best with attack speed, and splash scales best with attack damage

uberquell on phase would be just kinda doubling up on the bonus damage part, while treads have honestly always felt like they've been low-impact early on because they just don't give enough damage to fuel their damage-reliant splash

PR #2588, opened 2019-04-27
Purge most buffs and debuffs at the start of the duel

DarkoniusXNG:
The list of removed buffs and debuffs include:

"black_king_bar spell immunity",
"skadi_slow"
"heavens_halberd disarm",
"silver_edge_debuff",
"ice_blast debuff",
"bloodseeker_rupture",
"doom_bringer_doom",
"invoker_deafening_blast_disarm",
"sand_king_caustic_finale_orb explode delay debuff",
"shadow_demon demonic purge_slow",
"venomancer_poison_nova debuff",
"viper_viper_strike_slow debuff",
"Witch Doctor maledict debuff"

If I need to update the list let me know. Transformation buffs (like Terrorblade's Metamorphosis and Dragon Knight Dragon Form should not be included to avoid issues.)


SquawkyArctangent:
Actually, random thought: if vanilla transformation buffs react poorly to forced removal, would setting their duration to like 0.03 work? That should let them expire normally.

Granted, I don't know what sort of eldritch magic would make the vanilla transformation buffs not like forced removal, but Dragon Knight _is_ one of the oldest heroes in the game, so ... I'm not surprised.

DarkoniusXNG:
Setting duration to low value could work.
I forgot to say: Removing transformations also seems wrong to me. Imagine a scenario, you play Sven and you activate your ult just before the duel, and then it gets removed in the duel. You are now in a duel without ult. This wouldn't happen at the competitive level but in casual games it ruins the game feeling and punishes you just because you didn't pay attention when duel starts.

DarkoniusXNG:
I updated the list. I will test tomorrow what can be removed and what can't.

l34Um1:
Working just fine for me, so whenever Darkonius says to merge I think it's ready to go.

DarkoniusXNG:
Works fine because I temporarily disabled problematic stuff.

PR #2595, opened 2019-05-01
Storm Spirit Rebalance

DarkoniusXNG:
The goal is to make rightclick/overload Storm Spirit more viable than carpet bomb (auto remnant) Storm Spirit.
* Buffed lvl 6 Electric Vortex -> Aghanims more viable.
* Buffed Overload dmg.
* Reverted the nerf for Static Remnant dmg at OAA levels.
* Nerfed Auto Remnant talent distance from 500 to 750.
* ~Made Ball Lightning scale better: early levels less mana cost but worse cast point. Higher levels: more dmg and better cast point, but bigger mana cost. I am trying to make people actually upgrade this and nerf lategame zip-zap at the same time.~ Nerfed Ball Lightning percentage mana cost, buffed base mana cost.
* Nerfed Kaya and Kaya upgrades manacost/mana loss reduction. ~Only Storm benefits from that anyway.~
* Reshuffled talents:
Lvl 10: Bonus mana regen vs bonus attack speed
Lvl 15: Bonus HP vs bonus remnant dmg
Lvl 20: Overload pierces immunity vs bonus electric vortex time
Lvl 25: 750 Auto Remnant vs +500 Overload dmg


game/scripts/npc/abilities/storm_spirit_ball_lightning.txt
```diff
@@ -36,17 +36,17 @@
...
       {
         "var_type"                                        "FIELD_INTEGER"
-        "ball_lightning_initial_mana_base"                "30"
+        "ball_lightning_initial_mana_base"                "10 20 30 40 50" //OAA
       }
       "03"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "ball_lightning_move_speed"                       "1250 1875 2500 3000 3500"
+        "ball_lightning_move_speed"                       "1250 1875 2500 3100 3500"
```
POCKETMONEY99: 3100. :)

game/scripts/npc/abilities/storm_spirit_static_remnant.txt
```diff
@@ -52,7 +52,7 @@
       "04"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "static_remnant_damage"                           "120 175 230 285 340 395"
+        "static_remnant_damage"                           "120 175 230 285 400 600"
```
chrisinajar: why though? this goes against the point of the patch...

game/scripts/npc/abilities/storm_spirit_ball_lightning.txt
```diff
@@ -19,15 +19,15 @@
...
 
     // Cost
     //-------------------------------------------------------------------------------------------------------------
-    "AbilityManaCost"                                     "30"
+    "AbilityManaCost"                                     "10 20 30 40 50" //OAA
 
     // Damage
     //-------------------------------------------------------------------------------------------------------------
-    "AbilityDamage"                                       "8 12 16 20 24"
+    "AbilityDamage"                                       "8 12 16 32 48"
```
chrisinajar: if the point is to make the storm engage more, why make the move where he's invulnerable do 2x damage?

game/scripts/npc/abilities/storm_spirit_static_remnant.txt
DarkoniusXNG: Because the scaling is bad? Also I asked Naeil about this, he told me to buff every level of remnant, lmao.

game/scripts/npc/abilities/storm_spirit_ball_lightning.txt
DarkoniusXNG: Buff to dmg is to encourage people more to upgrade the ability, because someone told me that if mana cost increases people will not want to upgrade it. Can be easily reverted.

DarkoniusXNG:
> Nerfed Kaya and Kaya upgrades manacost/mana loss reduction. Only Storm benefits from that anyway. is completely untrue. im not saying kaya doesn't need a nerf, im just saying there's a ton of heroes who benefit from it.

I wanted to say Storm benefits from it the most. Kaya needs a nerf like you said (even the spell amp needs a nerf but thats offtopic).

> I think you wildly both zip and remnant,

I buffed early levels of ball lightning, because its worse than in vanilla dota and storm is useless in early game. Also cast point is worse at early levels.
Nerfing the ball lightning completely would make the hero trash.
Remnant dmg values are reverted how it was (still worse than before the major), because I overnerfed it and because auto remnant is worse now. Also look at the scaling how bad it is.

chrisinajar:
larenall and I had an interesting conversation about storm spirit balance. If the purpose here is to make people not pick that talent, why not just remove the talent? With that out of the way we can buff the shit out of remnant again and make it relevant since you can't create them while invulnerable. We were talking about different options for an opposing level 25 talent to the overcharge damage and thought that a strong CDR talent would be interesting.
This would completely kill the non-interacting SS builds and enable a few other new interesting ones to be created; abusing the CDR with activatable items or going for massive overcharge right click damage, both of whom still benefit from the entire SS kit...
thoughts?

Larenall:
> larenall and I had an interesting conversation about storm spirit balance. If the purpose here is to make people not pick that talent, why not just remove the talent? With that out of the way we can buff the shit out of remnant again and make it relevant since you can't create them while invulnerable. We were talking about different options for an opposing level 25 talent to the overcharge damage and thought that a strong CDR talent would be interesting.
> This would completely kill the non-interacting SS builds and enable a few other new interesting ones to be created; abusing the CDR with activatable items or going for massive overcharge right click damage, both of whom still benefit from the entire SS kit...
> thoughts?

In order to make ppl pick cdr it has to be like 30-50%. else its always gonna be a nice to have thing with lvl 5 aghs. Also we shouldnt get carried away with changes like this cuz they may remove the point of the talents -  to chose 1 of 2 different playstyles. Main word here is DIFFERENT. In this case cdr can be a very nice and strong addition of E based storm and a reason to get aghs(not only cdr but also all other talent which arent bad and mini BH ofc). Do we actually care about the "point of talents" or not? If not then we can juggle talent however we want and that may give us other way to balance heroes apart from just value changes.

Larenall:
> we shouldnt get carried away with changes like this cuz they may remove the point of the talents - to chose 1 of 2 different playstyles. Main word here is DIFFERENT. In this case cdr can be a very nice and strong addition of E based storm and a reason to get aghs(not only cdr but also all othe

Cuz then if talent is stupid/imbalanced we can just remove it and add another one that will make hero play differently. Ofc that will not work for all heroes cuz some have more impactful talents and others dont but we can change that as well 🤔 cant we?

DarkoniusXNG:
I never said that the purpose of this PR is to make people not pick the auto remnant talent. The goal is to nerf auto remnant play-style. Also even if we remove the talent, that wouldn't remove the real issue: Infinite mana. Storm will still have 800 mana regen to zip around: https://cdn.discordapp.com/attachments/479059280805494795/572493299252723763/unknown.png
That's why I nerfed Kaya. In my other PR I will rework Refresher stats.

lovesmuffins:
I think people are overly overrating this talent. It's picked because it's good and makes storm a viable hero, the other talent doesn't do that which is why people only go that one style. I think this is definitely something we should try in the test client. The problem right now is that storm just doesn't use up enough mana and he can just ult forever. I honestly think decreasing cast point overall will allow people to catch storm after he zips, I don't think the distance (500 > 750 atm) should be to nerfed as 750 as I fear it's to drastic to fast, I say make it 600 units. I also think the problem is that the remnant duration lasts to long which allows storm to zone out people a little bit to hard. @chrisinajar, I think you should consider making the duration of his remnant to 6 rather than 12. This will allow us to nerf that playstyle while not directly changing it. I think initial manacost should be less overall while mana per unit is increased. This is intended to make storm players be closer to the fight or duel so that they're more easily able to be stopped along side the cast point nerf, but this won't directly nerf or buff how much mana storm uses in a fight. Rather than him doing longer zips he'll be forced to do shorter but more zips. I personally think speed shouldn't scale as much at OAA levels either. IMHo it should go form 2500 to 2750 to 3000 not what it is currently as Speed is probably the underlining most broken part as more speed means harder to catch and less mana usage to get from point a to point b. These are just my thoughts and I'll probably put them into my own test client to see how storm would be played.

Larenall:
> I think people are overly overrating this talent. It's picked because it's good and makes storm a viable hero, the other talent doesn't do that which is why people only go that one style. I think this is definitely something we should try in the test client. The problem right now is that storm just doesn't use up enough mana and he can just ult forever. I honestly think decreasing cast point overall will allow people to catch storm after he zips, I don't think the distance (500 > 750 atm) should be to nerfed as 750 as I fear it's to drastic to fast, I say make it 600 units. I also think the problem is that the remnant duration lasts to long which allows storm to zone out people a little bit to hard. @chrisinajar, I think you should consider making the duration of his remnant to 6 rather than 12. This will allow us to nerf that playstyle while not directly changing it. I think initial manacost should be less overall while mana per unit is increased. This is intended to make storm players be closer to the fight or duel so that they're more easily able to be stopped along side the cast point nerf, but this won't directly nerf or buff how much mana storm uses in a fight. Rather than him doing longer zips he'll be forced to do shorter but more zips. I personally think speed shouldn't scale as much at OAA levels either. IMHo it should go form 2500 to 2750 to 3000 not what it is currently as Speed is probably the underlining most broken part as more speed means harder to catch and less mana usage to get from point a to point b. These are just my thoughts and I'll probably put them into my own test client to see how storm would be played.

Speed doesnt effect manacost in any way. Actualy the more speed the worse in terms of mana cuz it takes more time to travel from a to b and during that time u regen more mana

chrisinajar:
The point isn't that the build is good, it's that it's non-interactive and lame. As long as remnants do any damage and you can place them while invulnerable then people *will* go that build because they're cheesy wieners and don't want to risk interacting with their opponent. It's not fun for anyone. From where I'm sitting, that talent is just bad game design.

I think the point of this PR should be to stop people from ever using that playstyle. If the talent needs to remain, then the remnant damage cannot. I am of the opinion that remnant damage is much more interesting than the talent, and prefer that over it quite a bit.

I was thinking 30%, yeah.

DarkoniusXNG:
@chrisinajar 
You can always place remnants while invulnerable (You can use Static Remnant ability while zipping) but I know what you are trying to say :D.

> I don't think the distance (500 > 750 atm) should be to nerfed as 750 as I fear it's to drastic to fast, I say make it 600 units

If a storm player thinks 750 is too much. The nerf is on point then.

lovesmuffins:
> The point isn't that the build is good, it's that it's non-interactive and lame. As long as remnants do any damage and you can place them while invulnerable then people _will_ go that build because they're cheesy wieners and don't want to risk interacting with their opponent. It's not fun for anyone. From where I'm sitting, that talent is just bad game design.
> 
> I think the point of this PR should be to stop people from ever using that playstyle. If the talent needs to remain, then the remnant damage cannot. I am of the opinion that remnant damage is much more interesting than the talent, and prefer that over it quite a bit.
> 
> I was thinking 30%, yeah.

The playstyle is absolutely interactive to the environment which is why it's in the game in the first place. Storm still needs to go from point A to B in order to put down the remnants. Making it harder for storm to get to Point A to Point B (suggestions I put above) should be the concern. Removing the talent or making it never be picked up is in essence ruining the game design of storm as it's the only thing that's keeping him relevant in oaa.

lovesmuffins:
> @chrisinajar
> You can always place remnants while invulnerable (You can use Static Remnant ability while zipping) but I know what you are trying to say :D.
> 
> > I don't think the distance (500 > 750 atm) should be to nerfed as 750 as I fear it's to drastic to fast, I say make it 600 units
> 
> If a storm player thinks 750 is too much. The nerf is on point then.

This is what I'm thinking. One level of Leshrac q increases radius by 25. At max level that is a 75 difference. Tell me how big of a difference that is to help you land that spell. If leshrac's q was stun at the base radius for the entire game nobody would pick it or even attempt to use it. I want to emphasize that at no point should the point of the PR be to make this talent useless, if you want to do that just remove it and replace with a boring talent that offers no unique play styles cause it seems that's what is encouraged around here.

DarkoniusXNG:
I asked almost every Storm player in OAA community how to change this hero. And for a reason, because they are biased. Then I asked some non-Storm players. This PR is the result.

To end this pointless discussion: With this PR and when I fix Refresher stats in another PR, then we will see if Storm's real issue is fixed or not. If not we remove the talent or the whole hero. All in favor?

DarkoniusXNG:
I tested the ball lightning changes. Scaling cast point and scaling mana cost is very bad. With less mana cost at lvl 1, no need to upgrade the spell, With item build I posted in a picture you literally never run out of mana, you don't even lose mana. Scaling cast point is bad because you can cast ball lightning while in ball lightning with ignoring cast point time - that means that bad cast point doesn't mean anything for good Storm players. Auto Remnant is worse than before though.

That's why I nerfed mana costs at all levels. Sorry Storm players, it needs to be done.

Larenall:
Why initial manacost needs to be nerfed? Isnt it bad for new storm playstyle?

PR #2596, opened 2019-05-01
Tooltips for spider boss abilities

AustraliaIsMyCity:
Made tooltips for the spider boos as they were previously just a bunch of numbers with no descriptions.


lovesmuffins:
Can you do one for weaver as well?

DarkoniusXNG:
Every non-siltbreaker boss doesn't have tooltips.

PR #2601, opened 2019-05-03
Devastator code refactor and fixes

DarkoniusXNG:
* Fixed Devastator passive being disabled by Break.
* Fixed Devastator and Desolator armor reduction stacking.
* Fixed Devastator active not doing anything if active armor reduction is lower than passive armor reduction.
* Better armor reductions now have more priority and they don't stack with each other.


DarkoniusXNG:
* Removed cast point on Devastator (items should not have cast points).
* Fixed active bonus damage being amplified by spell amplification and being blocked by damage block.

PR #2603, opened 2019-05-04
First Step for Buffing creeps Issue

Jochnickel:
spawner would have to be placed in the map with
CreepType = 100, (CreepMax = 1 of course)
im not sure if scaling would work, since its oaa mechanic that i dont know


l34Um1:
From what I understand this still needs zones hooked up to the spawning of the bosses. I would appreciate it if @chrisinajar or @DarkoniusXNG could have a look.

DarkoniusXNG:
Some things are not done in this PR:

> 1) They can't be stacked and spawn one at a time. Only one of each can be on the map at any given time.

This was not even done properly for corner creeps. How to do? Store mini bosses in a table or handle and check if they are alive every 0.1 seconds or something.

> 2) When killed, they give a point.

How to do? `PointsManager:AddPoints(teamId, 1)` where `teamId` is `DOTA_TEAM_GOODGUYS` or `DOTA_TEAM_BADGUYS`. You can get that with `GetTeamNumber()`.
>3) The creeps respawns at the normal minute interval as other creeps but only after 3 minutes have passed.

I have no idea how to do this atm, more OnIntervalThink code I guess. But that will clutter the code more lmao.
> 4) The creeps visually carry a buff. The buffs are visually obvious on the heroes carrying them.

Just add some particles to the buffs.
> 5) Which creep has which buff is random and is re-randomed every spawn. No matter which creep has which buff, both buffs always spawn. 

Buffs are random completely. Not sure.

Also @Jochnickel can you merge the contents of all modifier .lua files into .lua file of the ability `miniboss_apply_buff.lua`
Then:
* the code will be more readable
* you will not need `require (...) `
* LinkLuaModifier will all have the same filepath 
* it will be easier to debug because everything is in one file.

And @Jochnickel did you test this? Because it looks like one OnIntervalThink is overriding the base OnIntervalThink. Dunno if that was your intention.

PR #2611, opened 2019-05-08
Fix the broken fix of Greater Phase Boots

Trildar:
I need some shame bells in here. Shame on @SquawkyArctangent for derping `spell` vs `ability` and shame on @chrisinajar for ignoring the Travis fail. :V


chrisinajar:
hadn't refreshed the PR list when i was fixing it, lul

PR #2617, opened 2019-05-09
PA Crit Rebalance

POCKETMONEY99:
after playing PA alot and calculating her average damage increase from Coup de Grace I have concluded while her average damage increase is fine the amount of damage a single crit can do is too high. When the amount of damage a single crit can do gets too high she becomes exceptionally bursty. Easily being able to dispose of heroes if she double crits. 

(When the critical damage is more then 500% the passive just gets extremly dumb. 500% Critical damage at level 50 should be fine)

I have rescaled the critical strike to deal roughly the same average damage as before. (Worse average damage then PA's crit in Dota at level 18). The cirtical strike damage is lower but the chance is higher. 

The 5% Coup De Grace talent is slighlty worse on this version.

### DIFF
--- a/game/scripts/npc/abilities/phantom_assassin_coup_de_grace.txt
+++ b/game/scripts/npc/abilities/phantom_assassin_coup_de_grace.txt
@@ -19,7 +19,7 @@
       "01"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "crit_chance"                                     "20"
+        "crit_chance"                                     "20" //OAA
         "LinkedSpecialBonus"    "special_bonus_unique_phantom_assassin_2"
       }
       "02"
--- a/game/scripts/npc/abilities/phantom_assassin_coup_de_grace.txt
+++ b/game/scripts/npc/abilities/phantom_assassin_coup_de_grace.txt
@@ -19,13 +19,13 @@
       "01"
       {
         "var_type"                                        "FIELD_INTEGER"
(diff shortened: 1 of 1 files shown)


DarkoniusXNG:
Changing this wouldn't change much. It would actually buff PA. I need to remind you why PA is such a good carry (from a support point of view). Her Q slows and deals dmg. Dmg scales throughout the whole game. And also dagger cannot miss. Her W has a long cast range and allows her easily to close the gap and kill people. In combination with her Q she becomes unkite-able. You cannot run away from her, you cannot avoid her. If she has BKB, you pray that your carry will kill her first. Protect yourself with Ethereal or Euls? You can't because of her Blur active. Inbuilt smoke effect that allows PA to jump on people unprepared and deal dmg to them. Because of the ult, she can one-shot people instead of just dealing dmg. But is ult really the problem? Even if you remove PA's ult, she is a godlike physical dps carry just because of first 3 abilities.
Other things PA is good at is farming (because of the cleave talent and Blur passive/active) and killing bosses (again because of Evasion).
So what is the solution? How to nerf PA?
Lets ignore her evasion and crit for a moment (because those things make PA unique).
**Nerf cooldowns and cast range**. But you say, that will ruin the hero? Then put the duration of Blur to 4 seconds at all levels, so you can at least see her coming, or you can at least jump on her and kill her without her BKB.
Another way of nerfing her, but wouldn't address the real issue is removing her cleave talent.

PR #2621, opened 2019-05-14
Azazel Shop Rework

DarkoniusXNG:
* Removed Azazel Defense Towers
* Removed Azazel Watch Towers
* Removed Azazel Walls
* Removed Azazel Tank creeps
* Removed Azazel Farmer creeps
* Removed Azazel Scout creeps (lvl 1-3)
* Added Azazel Fighter creep (ranged creep with high dmg, armor, and hp), ~you can have more than 1~, ~infinite~ 5 min duration. Requires tier 4 core.
* Reworked Azazel Scout lvl 4 creep. Added 100% magic resistance. Added True Sight and Flying Vision to the Scout (1200 radius). ~You can have more than 1~, infinite duration. Requires tier 4 core.
* Added Azazel ~Teleport Point~ Shrine (normal building like a wall, but less wide, gives 500 vision). Requires tier 3 core.

We will probably add some abilities to the Shrine in the future to be more interesting.

~Work in Progress~


POCKETMONEY99:
Can you make the teleport point hits to destroy? I think it would work much better then the current version of structures.

DarkoniusXNG:
With these changes, all elixirs are visible in the shop (no need to type to find them). And makes the Azazel shop look clean and nice.

@POCKETMONEY99 We will balance it later. Adding hits to destroy would make Shrine (teleport point) useless imo.

chrisinajar:
`Added Azazel Fighter creep (ranged creep with high dmg, armor, and hp), you can have more than 1, infinite duration. Requires tier 4 core.` :thinking:

Larenall:
> `Added Azazel Fighter creep (ranged creep with high dmg, armor, and hp), you can have more than 1, infinite duration. Requires tier 4 core.` 🤔

^

DarkoniusXNG:
Maybe give some suggestions? You can have only 1 or 5 min duration. Or both.

l34Um1:
I think both.

PR #2627, opened 2019-05-16
Vampiric fang Rescaling

POCKETMONEY99:
This item is far to risky to be realisitically purchased. Especially with satanic giving status resistance and blade of judecca exisiting Vampiric fang isn't a viable item. I increased the normal lifetsteal to match satanic.

I have also rescaled the item to drain less HP but also give less lifesteal. 

[When I critically think about what this this item does I think it may need additional features when turned on. It seems counter intuitive to have an item which causes you to take damage per second, but only give you a way to heal that damage (lifesteal). Addtionally this would help the item not be completely useless vs regeneration reduction). Its like if regeneration crystal sacrificed some of your HP first before healing you.]


DarkoniusXNG:
Lifesteal buff is fine. Damage_per_second_pct buff is not. Remember my Armlet complaints? 
Item also needs some tweaking on the code level: armlet + fang interaction. Don't forget that Fang is insane on Gyro and Io.

POCKETMONEY99:
> 
> 
> Lifesteal buff is fine. Damage_per_second_pct buff is not. Remember my Armlet complaints?
> Item also needs some tweaking on the code level: armlet + fang interaction. Don't forget that Fang is insane on Gyro and Io.

I would rather nerf the active lifesteal. The item is too extreme in its current state, as is completely useless vs regen reduction.

Larenall:
I agree on nerfing active lifesteal and HP loss. Also just remove IO from captains mode. Even valve gave up on balancing it and removed him from cm as well

PR #2639, opened 2019-05-22
Tidehunter Talents Buffs

POCKETMONEY99:
Tidehunter has extremely long periods of "down-time", where he isn't very useful (pre 25). 
I decided to reshufle his talents and buff them to give him better skirmishing powers. 

Swapped the level 10 gush damage and level 15 anchor smash damage reduction talents around. However I reduced the value on the anchor smash by 10%, which he gets back from level 5 anchor smash. 

Increased the following talents:
Level 15 Gush damage from 100 to 400
level 20 Gush armour reduction from 5 to 7
Level 20 Kraken Shell Damage from 30 to 40


DarkoniusXNG:
> Enigma's main issue he is crippled by black holes insane cooldown.

Not true. Enigma also has a cd reduction talent. Black Hole cooldown is long and it should stay long because Enigma can kill a whole team with right item build. We already discussed this but you are stubborn.

POCKETMONEY99:
> 
> 
> > Enigma's main issue he is crippled by black holes insane cooldown.
> 
> Not true. Enigma also has a cd reduction talent. Black Hole cooldown is long and it should stay long because Enigma can kill a whole team with right item build. We already discussed this but you are stubborn.

The Spell has over a 3 minute cooldown level 1. This is OAA, 3 fights/skirmishes could easily happen in that time.

game/scripts/npc/abilities/tidehunter_anchor_smash.txt
```diff
@@ -47,7 +47,7 @@
       "02"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "damage_reduction"                                "-30 -40 -50 -60 -60 -60"
+        "damage_reduction"                                "-30 -40 -50 -60 -70 -70"
```
chrisinajar: kinda weird that level 5 scales but level 6 doesn't :thinking:

POCKETMONEY99: I did this because it would be nearly 100% damage reduction

PR #2645, opened 2019-05-27
OD nerfs

POCKETMONEY99:
Decreases the bonus int gained at OAA levels. 
Reverted the mana cost buff because of patch 7.22 reworking equilibrium.
I also linearised the scaling on the cast range for astral imprisonment.


chrisinajar:
this is a great PR because it doesn't overnerf -- OD will still be very very strong, as all heroes should be.

PR #2647, opened 2019-05-27
Tier 5 Twins nerf

POCKETMONEY99:
This is a very long overdue nerf to Tier 5 twins. I set the base attack time and attack range to the same as tier 2 twins. Their damage output makes it impossible for most team compostions to kill them. Addtionally their weird attack range at level 5 makes it impossible to separate them.


Larenall:
Also set their attack range to the same number so i can split them

POCKETMONEY99:
> 
> 
> Also set their attack range to the same number as t2 has so you can split them

PR #2651, opened 2019-05-29
Kaya, Yasha and Sange (Removed Backward Compatibility)

POCKETMONEY99:
***I have removed backward compatibility for these items***
Readded KYS level 1 and 2 variants to the shop. After maim proc was removed from Sange there was no reason for these items not to be tier 1. 

Sange and Yasha is currently unpurchaseable, because you have to buy a level 2 Sange or Yasha, which is an aweful way to spend a tier 1 core.  Kaya only gets purchased because it is mandatory for most casters to scale into the midgame. 

The only issue I see with readding level 1 and 2 KYS is that the level 2 upgrades could potentially be too valueable for their price. This can easily be fixed with some number changes. 

For the arguement that this is just a boots replacement, buffing phase and power treads will be a good way to make this not the case. Additionally there are other boots replacements which are currently level 1 items, (Manta style, blink dagger and drums come to mind. Or just do what most carry plays do and just getany item and get a support to buy luciences.)


Larenall:
I agree with KYS lvl1-2 BUT i dont with kaya stats nerf.

> Spell amplfication is too strong with damage which scales off spell amplfication and another variable. (Skywrath Q, Spirit Breaker Bash, All current/max health based damage, eblade etc).

no, it only increases dmg by 42% at lvl5 kaya, it doesnt stack in any way(doesnt amplify the dmg 2 times if that what are u talking about.) For example Skywrath Q: Its not (220+(1.3*Spell amp*int))*Spell amp, its  (220+1.3*int)*1.42 just like with any other spell. Same for SB bash: (1.92*MS)*Spell amp where 1.92 is Bash dmg. I see nothng wrong here.
With dmg based off HP its a diffrent thing.

POCKETMONEY99:
Reverted the manacost and spell amp nerf

PR #2652, opened 2019-05-30
Fix for Dagon + Fatal Bonds on illusions

DarkoniusXNG:
When you use Dagon on illusion that is affected by Fatal Bonds, everything that is fatally bonded will take 13 371 337 pure damage. This 'bug' allows Warlock to instakill the enemy team if they have illusions.


game/scripts/vscripts/items/dagon.lua
```diff
@@ -42,6 +42,12 @@ function item_dagon:OnSpellStart()
     return
   end
 
+  -- If the target is an illusion, just kill it and don't do damage
```
chrisinajar: Should it also do damage so that fatal bonds does reflect the damage that it should reflect? merging either way because it's a strict improvement, but I kinda think it should still do the damage

PR #2656, opened 2019-06-04
the map we need but not the map we deserve

l34Um1:
I dunno how to write radiant lmao


Larenall:
Power spark/boots buff is reasonable. Tranquil spark is in a perfect spot rn. Tranquil spark is better for roamers and arcane spark for caster farmers. Idk what has to be done to tranqul boots to make tham viable.
Their main 2 problems are that you gain much less gold while bossin cuz you move less and that they dont build into anything like greaves(greaves apart from giving u gold give u stats and active effect while tranquils don't)

PR #2657, opened 2019-06-05
Make Sparks cheaper

POCKETMONEY99:
Reduced the reciepe cost of sparks from 600 to 25. Currently there is no variety in starting items because most of your gold is commited for your spark. (Also everyone starts off with a spark so the cost shouldn't be be so high.)


Larenall:
Why they just cant be free, you buy them no matter what. By making so they cost money you decrease starting gold. If you want starting gold to be lower then just lower it 4head and makes spark free. This is very wierd to me. you = whoever made so spark recepie costs money.

POCKETMONEY99:
> 
> 
> Why they just cant be free, you buy them no matter what. By making so they cost money you decrease starting gold. If you want starting gold to be lower then just lower it 4head and makes spark free. This is very wierd to me. you = whoever made so spark recepie costs money.

I set the cost to 25 gold now. I don't think reciepes can cost 0 gold. Reciepes which cost 0 gold means there is no reciepe. This is about as 'free' as sparks will get. (I could set the cost to 1 but I don't think that will change anything).

Larenall:
> > Why they just cant be free, you buy them no matter what. By making so they cost money you decrease starting gold. If you want starting gold to be lower then just lower it 4head and makes spark free. This is very wierd to me. you = whoever made so spark recepie costs money.
> 
> I set the cost to 25 gold now. I don't think reciepes can cost 0 gold. Reciepes which cost 0 gold means there is no reciepe. This is about as 'free' as sparks will get. (I could set the cost to 1 but I don't think that will change anything).

Fair

DarkoniusXNG:
Wouldn't be easier to increase starting gold instead of doing this?

POCKETMONEY99:
> 
> 
> Wouldn't be easier to increase starting gold instead of doing this?

It probably was but I have done it now lol.

Issue #1606, opened 2017-12-15
Covfefe nerf


l34Um1:
- Have it decrease in effect over time. So at the start of the duration it's at 100% effectiveness, at the and it's at 0%.

- Have it be blockable by BKB and purgeable.

chrisinajar:
first thing is already done, just needs the second one.

PR #2672, opened 2019-06-13
Skadi/Abyssal/Judecca item tree

DarkoniusXNG:
Skadi, Judecca and Abyssal now give the same amount of HP (so you don't lose HP when you upgrade).
1) Skadi changes:
- Buffed stats at lvl 4 and 5.
- Buffed slow at all levels. (Against ranged units maybe its too much but after 7.22 Skadi slow is kinda garbage)
- Buffed HP bonus at early levels.
- Toned down scaling of mana (This needs to be done for every item that gives mana, because there is too much mana in the game anyway but this is a discussion for another time).
2) Abyssal Blade changes:
- Dmg nerf. Total Physical DPS for non-str heroes is now the same as DPS of a Skadi. 
Proc dmg on Bash unchanged. STR bonus unchanged.
- HP bonus buff. Total bonus hp (bonus STR + bonus hp) is now roughly the same as Skadi (75/0/25 HP difference).
3) Judecca changes:
* Buffed stats to match Skadi new stats.
* Total bonus HP (stats + bonus hp) is now the same as Skadi. Similar thing for total mana.
* ~~Frostbite pierces BKB again. I found this though: https://github.com/OpenAngelArena/oaa/issues/1606 It appears Covfefe was very powerful in 2017 so I am not sure about this change. Skadi and Abyssal pierce BKB so it was kinda logical to me that Judecca pierces BKB too.~~
* Frostbite duration is now the same as skadi slow duration.

Quick reminder for Judecca: Frostbite is dispellable.


POCKETMONEY99:
I was thinking about Judecca yesterday lol. I always thought judecca pierced bkb. I like the duration being 3 seconds. I feel like the problem with judecca is it's an extremely strong effect but their is no real counterplay to it as it has no cooldown. (Im assuming thas why it doesnt pierce spell immunity) Personally I would make it an undispellable active with a moderate cooldown (like break on silver edge). 

Also I think the bonus damage on judecca should either actually matter, or be removed entirely, with an all stats buff or something. 10 20 damage on a late game item is a joke.

DarkoniusXNG:
Numbers for bonus dmg dont matter if you look at the stats it provides.

PR #2674, opened 2019-06-13
Give Yearning Spark passive GPM

chrisinajar:
This makes the yearning spark a bit less of a noob trap by ticking passive GPM from arcane spark when un-upgraded. It does not have the active or anything from the arcane spark, just gives the GPM.


DarkoniusXNG:
Tooltip :)

PR #2680, opened 2019-06-14
Reduce Bloodstone Cooldown

POCKETMONEY99:
I have reduced the astronomical cooldown on Bloodstone to something actually usable in OAA. It is still a once per fight item, and at early levels you may have to fight without bloodstone active being up but is much better then bloodstones 250 second cooldown.

### DIFF
--- a/game/scripts/npc/items/item_bloodstone.txt
+++ b/game/scripts/npc/items/item_bloodstone.txt
@@ -37,7 +37,7 @@
     // Stats
     //-------------------------------------------------------------------------------------------------------------
     "AbilityCastPoint"                                    "0.0"
-    "AbilityCooldown"                                     "250.0"
+    "AbilityCooldown"                                     "160.0 140.0 120.0 100.0 80.0" //OAA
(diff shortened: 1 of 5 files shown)


DarkoniusXNG:
I wanted to rescale the cd as well. You maybe went a little overboard but I guess its fine. For example I would leave 250s at level 1. Can you also make `mana_cost_percentage` 30 at all levels? People told me that its stupid that it scales.

Larenall:
item with 4 mins CD :\

POCKETMONEY99:
I have rescaled the bloodstone cooldown to go from 160 to 80 instead of 120 to 80. While I agree early game this item should have a long cooldown. 160 seconds level 1 is still an extremly long time in OAA. If this item ends uo being OP I think the active cooldown should still be farily low, but make the active mana cost to "20 25 30 35 40" So you can't convert as much mana to health erly game, Making you easier to kill. Bloodstone is basically a regen crystal on crack, but it costs more resources and has a higher cooldown. 

I also removed the scaling mana costs.

PR #2699, opened 2019-07-19
Sohei and Chatterjee fix

DarkoniusXNG:
This is a fix for Sohei and Chatterjee not having stats, abilities, models etc. 
I tried to fix their hero portraits. Panorama is not working how it should ...
I tried to fix their hero icons on the minimap -> still juggernaut and dazzle


DarkoniusXNG:
Travis is complaining but not because of 7.22e. 
In dota imba, they have custom heroes in both `npc_heroes_custom.txt` and `npc_units_custom.txt`. 

Example for Sohei in dota imba:
[sohei.txt](https://github.com/EarthSalamander42/dota_imba/blob/master/game/dota_addons/dota_imba_reborn/scripts/npc/custom_heroes/sohei/hero.txt).
[npc_heroes_custom.txt](https://github.com/EarthSalamander42/dota_imba/blob/master/game/dota_addons/dota_imba_reborn/scripts/npc/npc_heroes_custom.txt#L29)
[npc_units_custom](https://github.com/EarthSalamander42/dota_imba/blob/master/game/dota_addons/dota_imba_reborn/scripts/npc/npc_units_custom.txt#L4)

When I did that here, keyvalues library malfunctioned (it said its duplicating KVs which is true, but it works in imba because they don't have that library), so I put hero files only in `npc_heroes_custom.txt` to bypass that error but with DOTAUnits as a header. My assumption is that Travis is complaining because of that, because its not DOTAHeroes. I don't know if it will work if I put DOTAHeroes.

TLDR: Valve fucked something up about custom heroes and now we need to guess how things work.

PR #2739, opened 2019-10-01
7.22h and other Valve nonsense

DarkoniusXNG:
* http://www.dota2.com/patches/7.22h
* https://steamdb.info/patchnotes/4232930/
* Improved talents for Lion, Omniknight, Oracle and Skywrath Mage.


DarkoniusXNG:
I did my best, Travis is still confused with some Valve shit they did. My guess Travis is confused that some keyvalues are appearing multiple times (one time in abilityspecials and one time normally, like abilityduration, abilitychanneltime, abilitycastrange etc.) This was Valve's ''fix'' for cast range in tooltips not updating with cast range bonuses (Aether Lens) and duration tooltips not increasing with negative status resistance. Maybe Outlanders update will bring an item that increases durations of everything, who knows.

PR #2741, opened 2019-10-05
Legion Commander Buffs 

POCKETMONEY99:
Despite numerous buffs legion commander is an extremely weak hero, being extremely easy to counter with the correct items or just by building tanky. LC is also very easy to kill, the sustain her kit offers isnt enough in a game which scales as hard as OAA>

**TALENTS:**
Improved both level 10 talents. I increased the strength from +8 to +15 and changed the other talent from +20 damage to +15% CDR. CDR improves all of her skills and is nice if your focus is more on dispelling allies rather then being a damage carry. 

Increased the level 15 attack speed talent from +30 to +50. This is to help her get duel kills. This hero struggles in duels as people get 6 slotted. (EBLADE :)) 

**SPELLS**
Increased the amount of HP regen Press the attack gives from levels 2-6 as the level of sustain this spell gives isn't strong enough considering the amount of damage flying around in OAA.


DarkoniusXNG:
Oh, so you gave up on idea of reshuffling talents.

POCKETMONEY99:
Cooldown Reduction Talent has been reduced from 25% to 15%.

PR #2743, opened 2019-10-07
Far Sight buffs

DarkoniusXNG:
* Reduced cd to 30s.
* Increased radius to 1000.
* Buffed stats to 15/30/45/60.


DarkoniusXNG:
* Fixed not revealing spell immune and invulnerable units.
* Fixed not revealing wards.
* Fixed AoE targeting not showing.

PR #2763, opened 2019-10-22
Kunkka Tidebringer Buffs

POCKETMONEY99:
Now that Kunkka is a weaker hero we can increase the power of tidebringer. Currently this spell is completely worthless, even if you go damage items. 

I heavily increased the damage on tidebringer at levels 5 and 6. I also decreased the cooldown at levels 5 and 6. This should make a carry kunkka be able to output alot more damage in a teamfight. This will also help increase his boss taking speed.


POCKETMONEY99:
I have tested this with triple daedalus, BKB, Silver Edge and armlet VS 0 Armour Tiny with Axe and Troll Warlord Behind them. I then tested it with an AC and Vladmir on the Axe. (In an actual game your going to have greaves aura aswell giving all your team over 50 armour. This isnt including any other items they have). It doesnt quite oneshot but it does alot of damage.

DarkoniusXNG:
Here is an idea: instead of buffing bonus damage at OAA levels to these numbers that look terrible, you make cleave percent scale.

POCKETMONEY99:
> 
> 
> Here is an idea: instead of buffing bonus damage at OAA levels to these numbers that look terrible, you make cleave percent scale.

I nerfed the amount of bonus damage on tidebringer. I tested increasing the cleave values at it was too good. All the multiplers would causes anyone behind the primary target to be oneshot. (Especially if the primary target has low armour.). It also doest help kunkka boss kill etc. I think increasing the bonus damage is the best option.

DarkoniusXNG:
>  It also doest help kunkka boss kill etc. 

Why do you want to make Kunkka better at killing bosses? :D

PR #2764, opened 2019-10-22
Battlefury Buffs

POCKETMONEY99:
Buffed battlefury. I am trying to make level 3-5 battlefury a viable item if you need the cleave vs alot of summons or illusions. The main strenght of this item allowing you to flash farm is irrelevant past 10 minutes. I increased the damage on BF heavilty from levels 3-5 to make is damage more comparable to other damage items. I also increased the Quell Damage. There is a huge increase levels 3-5. This should hopefully make the item a strong pickup if you need to clear summons as a right click carry.

PS; I am assuming level 1 battlefury is still OP and level 2 battlefury is viable.

### DIFF
--- a/game/scripts/npc/items/item_bfury.txt
+++ b/game/scripts/npc/items/item_bfury.txt
@@ -53,7 +53,7 @@
       "01"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "bonus_damage"                                    "45 60 120 180 240"
+        "bonus_damage"                                    "45 65 95 135 185"
       }
       "02"
       {
--- a/game/scripts/npc/items/item_bfury.txt
+++ b/game/scripts/npc/items/item_bfury.txt
@@ -68,7 +68,7 @@
       "04"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "cleave_damage_percent"                           "40 50 60 70 80"	//OAA
+        "cleave_damage_percent"                           "60 70 80 90 100"	//OAA
       }
       "05"
       {
--- a/game/scripts/npc/items/item_bfury.txt
+++ b/game/scripts/npc/items/item_bfury.txt
@@ -53,7 +53,7 @@
       "01"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "bonus_damage"                                    "45 60 85 125 180"
+        "bonus_damage"                                    "45 60 120 180 240"
       }
       "02"
       {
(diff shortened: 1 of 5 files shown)


game/scripts/npc/items/item_bfury.txt
```diff
@@ -53,7 +53,7 @@
       "01"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "bonus_damage"                                    "45 60 85 125 180"
+        "bonus_damage"                                    "45 60 120 180 240"
```
DarkoniusXNG: Change this to: 45 65 95 135 185

POCKETMONEY99: Can I get an explanation of why this item has to give abysmal damage. Im assuming the item was nerfed before fools gold was a thing and when quell was percentage based. Even with my buffs to 240 damage this item still gives lower DPS then all the other damage items (Bar nullifier but the active is far superior). The main strength of this item, which is allowing you to flash farm is irrellevant after 10 minutes (why would you upgrade it past level 2). If you need it to clear summons, radiance + octarine + shivas on anyone on your team is far superior, and will make them unkillable.

DarkoniusXNG: Battlefury is supposed to be **situational**, thats why I said buff cleave percent instead. I proposed those numbers because I used the same scaling logic on other dmg items: x/x+20/x+50/x+90/x+140, so mostly for consistency reasons. If you think damage is still crap DPS-wise, increase the cleave percent even more. If that is not enough, buff regen stats as well.

> If you need it to clear summons, radiance + octarine + shivas on anyone on your team is far superior, and will make them unkillable. 

Lycan wolves, Brood spiders and Enigma Eidolons have more magic resistance than armor. So cleave would be more effective against them.

POCKETMONEY99: Battle fury will still be situational after these buffs, you wont be buying a battlefury every game. BF is an item like echo sabre or armlet. These items need a little extra stat wise because they arent designed to be good late game items.

chrisinajar:
Items needs rework, since if it's strong it becomes a noob trap for farming.

POCKETMONEY99:
> 
> 
> Items needs rework, since if it's strong it becomes a noob trap for farming.


I reduced the bonus damage to "45 65 95 135 185" so the item still appears worse then desolator damage wise (Even if it gave 45 60 120 180 240 damage it would still be worse damage then deso in most cases). Battlefury will always be a noob trap item until it is actually good. Players who think BF is good will continue buying battle fury even if these buffs don't go through. These buffs will help make it less bad. (it will still be extremely bad in most situations.)

I dont think this item needs a rework but if you want to actually improve the item improve the damage the item gives by alot and make the cleave radius scale. This will make it into a strong splash damage into illusions etc. (Or for PA dagger builds/ Ember etc). 

If you still want to rework the item. Ravenous Hydra and Titanic Hydra from LoL are strong items with cleave (They have an active which deals burst physhical damage around you).

chrisinajar:
Rename it and you've got a deal

PR #2765, opened 2019-10-22
Euls and Rod of Atos Buffs. 

POCKETMONEY99:
Made allied euls int and ms match normal euls. Increased Normal Euls movespeed from levels 2-4. 
Improved rod of atos stats and reduces the items astronomical mana cost. These changes are aimed at making this items a viable alternative to hex. Current Hex outclasses these items, making the level 4 and 5 upgrades worthless. 

(I Also wanted to make euls damage scale but I think the damage is hardcoded.)

### DIFF
--- a/game/scripts/npc/items/custom/item_allied_cyclone.txt
+++ b/game/scripts/npc/items/custom/item_allied_cyclone.txt
@@ -76,7 +76,7 @@
       "01"
       {
         "var_type"                                        "FIELD_INTEGER"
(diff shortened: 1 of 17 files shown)


DarkoniusXNG:
I wanted to rebalance item stats of the whole euls-hex tree but I guess this is fine (a little bit overboard with the stats on the atos if you ask me but I can get over it).

P.S. Yes, euls damage is hardcoded :)

PR #2766, opened 2019-10-22
Sohei buffs

POCKETMONEY99:
Decreased the mana cost on Soheis Q. Currently the hero doesnt have a large enough mana pool to even cast all of his dashes at early levels. The cooldown prevents this spell from having too much uptime. Decreased the mana cost. 

Increased the Radius on soheis ultimate. Currently flurry of blows is broderline useless until your have level 25. Additionally I think ive seen sohei knock heroes out of his ultimate with its currrent radius. Increased the radius so you can actaully land it.


DarkoniusXNG:
I set that mana cost to 75 so Sohei could actually have some weakness in the early game. If you buff mana cost he would once again be extremely annoying in the early game (more annoying than he already is).

POCKETMONEY99:
> 
> 
> I set that mana cost to 75 so Sohei could actually have some weakness in the early game. If you buff mana cost he would once again be extremely annoying in the early game (more annoying than he already is).

This mana cost reduction is a necessity because Sohei doesnt have enough mana to cast all 4 dashes in rapid sucession at early levels. This spell effectievly has a 48 second cooldown and a 300 second mana cost at level 1. The actually distance you cover is around the same as a QoP blink. 

-The high cooldown on his Q charges combined with less charges is enough to make the hero less annoying early game. The massive mana cost on dash just ruins the flow of the hero.

-This spell only offers average mobility. 0 damage and isnt an extremely strong puesdo ultimate spell like tombstone. The main spell which makes sohei slippery is his W, a far superior spell. 

-The spell just isnt worth 75 mana per cast. For 75 mana you just move 300 units. That is so bad. 

As for sohei being annoying I swear by the end of this year I will rework this disfunctional hero. (would mainly be changing the Q and E, keeping the W and R mostly the same.)

DarkoniusXNG:
Ok fair points. I had an idea for his Q to slow enemies he dashes through but we will not discuss this now :D

PR #2767, opened 2019-10-22
Phantom lancer Doppelganger Buffs

POCKETMONEY99:
Decreased the cooldown on PL doppelganger from levels 1-5. Is now 15 second cooldown from level 4 onwards.


DarkoniusXNG:
With this change there will less incentive to lvl up the ability after lvl 4 ( I know that cast range scales but it doesnt matter). I would be fine with: `20 19 18 17 16 15` but it will look like I am bargaining with numbers.

Other and maybe better alternative is to **make 15s cd at all levels and scale illusion dmg taken/dealt**, but it would look really ugly in the tooltip.

Also, I have a question: Does PL really needs buffing?

POCKETMONEY99:
> 
> 
> With this change there will less incentive to lvl up the ability after lvl 4 ( I know that cast range scales but it doesnt matter). I would be fine with: `20 19 18 17 16 15` but it will look like I am bargaining with numbers.
> 
> Other and maybe better alternative is to **make 15s cd at all levels and scale illusion dmg taken/dealt**, but it would look really ugly in the tooltip.
> 
> Also, I have a question: Does PL really needs buffing?

PL doesnt need buffing in particular, this is a changed aimed at making doppelganger as strong as it can be without lagging the game. I was going to make the illusion damage scale but I dont think this hero needs more damage on his skilles. As for this spell being worthless to level past level 4 there are alot of skills like that on carry heroes, so I don't see much issue here.

POCKETMONEY99:
> 
> 
> With this change there will less incentive to lvl up the ability after lvl 4 ( I know that cast range scales but it doesnt matter). I would be fine with: `20 19 18 17 16 15` but it will look like I am bargaining with numbers.
> 
> Other and maybe better alternative is to **make 15s cd at all levels and scale illusion dmg taken/dealt**, but it would look really ugly in the tooltip.
> 
> Also, I have a question: Does PL really needs buffing?

The spell being a poor skill point past level 4 is fine imo, most other carry heroes have the same issue (Alot of them you probably shouldn't even level because of the mana cost increase). This change is aimed at making doppleganger as low cooldown as poissible without lagging the game. 

I was considering makign the illusions scale but I dont think PL needs more damage.

PR #2769, opened 2019-10-22
Abaddon Myst Coil heal nerf

DarkoniusXNG:
Nerf to the most banned hero at the tournament and in pubs. Level 5 is still strong. People will say this nerf is not enough but its a start.


POCKETMONEY99:
How about instead of nerfing mists coil, we let abba's ultimate provide AOE invunerability. With the current state of his aghs borrowed time either needs a rework or his Q and R need to be nerfed into oblivion for him to be balanced. Nerfing mist coil to the point where its balanced with the ultimate just means that mist coil will be an awful standalone spell. 

My proposal is (Leaving mist coil unchanged):
Borrowed time 
-Reduce search radius reduced to radius to 600. 
-Borrowed time duration reduced to 4 seconds at all levels. 

What this means?
-Mist Coil can remain unchanged, making this spell useful even if you dont have aghs.  
-Borrowed Time Threshold doesnt have to be increased to the point where his aghs is useless before OAA levels mist coil and allies having 6k + HP. 
-Abba has to be extremely close to his teammates to make them invunerable. This means he is going to get less people with his ultimate in a teamfight. If alot of his teammates group up with him for borrowed time this makes them all suspeptable to AOE CC.
- 4 second of invuerability is enough to mitigate important damage from big ultimates, but it also means you can counterplay thiis spell.

DarkoniusXNG:
> My proposal is (Leaving mist coil unchanged):
Borrowed time
-Reduce ally search radius to 600.
-Borrowed time duration reduced to 4 seconds at all levels.

I dont like this, it nerfs Borrowed Time too much. This Myst Coil nerf doesnt destroy the hero or the ability. Its only a nerf to the heal part (615 heal in an AoE on 3sec cd doesnt sound too bad tbh). I will revert this nerf when I rework the aghs upgrade to be the same like the old one.

The code for Borrowed Time with old aghs upgrade should be similar to this: https://github.com/EarthSalamander42/dota_imba/blob/master/game/dota_addons/dota_imba_reborn/scripts/vscripts/components/abilities/heroes/hero_abaddon.lua#L1046

PR #2770, opened 2019-10-22
Anti-mage Buffs

DarkoniusXNG:
These buffs are enough. Hero is fun to play already imo. There were bunch of ideas for his mana break to make it percentage based (of enemy max or current mana). That would make Anti-Mage overpowered and good against every hero, even if we put it as an Aghs upgrade. So I am against that idea. 

There were also arguments that Diffusal Blade does more damage than Anti-Mage Mana Break - this is not true if Anti-Mage has illusions. People forget that Antimage's mana break on illusions works differently than Diffusal Blade's mana break. Illusions with Antimage's mana break burn the full amount of mana and deal the same damage as the original hero. Illusions with diffusal blade don't. There is another argument that illusions are crap in OAA but I will not go there.

Some people also cannot believe that Anti-Mage can be played in a non carry position.


game/scripts/npc/abilities/antimage_mana_break.txt
```diff
@@ -28,7 +28,7 @@
       "02"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "mana_per_hit"                                    "28 40 52 64 88 124"
+        "mana_per_hit"                                    "28 40 52 64 88 160"
```
POCKETMONEY99: Can the scaling on mana break be "96 128" at OAA levels and then we get antimages current base attack time of 1.4 instead of his nerfed BAT of 1.5 in OAA.

DarkoniusXNG: You said it yourself that Anti-Mage is fine. Don't get influenced by Pigeon.

PR #2772, opened 2019-10-23
Clinkz Burning Army nerf

DarkoniusXNG:
Making Burning Army better in lategame damage-wise and worse in early game -> smoother power progression.


chrisinajar:
clinkz is very strong right now after recent buffs, i don't think he needs more buffs

DarkoniusXNG:
It should be fine now.

PR #2776, opened 2019-11-03
Change Midas Spark scaling

DarkoniusXNG:
... from scaling over time to scale (increase) bonus gold on each creep killed with midas spark.
New formula: **base gold + creep kill counter x increment gold**
**base gold = 250**
**increment gold = 100**
With this change: if you kill a creep with Midas spark each time you have 100 charges, at minute 5 you will have at least 3000 gold (with GPM spark you will have 2500). 3000 gold is less than the current Midas spark: if you save 4 usages (400 charges) and you use midas spark at or after 5 min (when it upgrades), you will have at least 5500 gold. Call it an abuse, optimization or a feature but its **not** OK.

P.S. With this change Midas spark is infinitely scaling so heroes that start with Midas will consider to never change it.


chrisinajar:
this sounds pretty cool

PR #2781, opened 2019-11-06
Hand of Midas fixes and nerf

DarkoniusXNG:
* Fixed Hand of Midas on Spirit Bear and Arc Warden Tempest Double not giving XP to the main hero.
* Hand of Midas cd from 45 to 55.
* Hand of Midas now kills a unit without changing bounties to 0 (like Midas spark). Because of this change: killing Mud Golems with Hand of Midas will not make Mud Golems splits useless.
* Total XP gained for killing the unit with Hand of Midas is the same.
* Total gold gained for killing the unit with Hand of Midas is slightly better (most noticable in the early game when it matters). Thats one of the reasons why we nerf the cd.


chrisinajar:
Can you increase the price of midas 2 recipe to be equal to or greater than the price of midas 1 recipe? It seems weird for the recipe costs to not be sorted.

DarkoniusXNG:
Yeah sure.

PR #2790, opened 2019-11-16
MKB proc damage buff

POCKETMONEY99:
Buffed the proc damage that MKB gives. MKB is far worse the Mjollnir currently.


DarkoniusXNG:
Increase PROC damage not physical part.

PR #2791, opened 2019-11-16
Spirit Vessel Nerf

POCKETMONEY99:
After greaves not being mandatory on most heroes, spirit vessel recieved a large indirect buff. Now this item is oppressive allowing any hero to deal high damage. 

Reduced the base damage on spirit vessel.

### DIFF
--- a/game/scripts/npc/items/item_spirit_vessel.txt
+++ b/game/scripts/npc/items/item_spirit_vessel.txt
@@ -74,7 +74,7 @@
       "04"
       {
         "var_type"                    "FIELD_INTEGER"
-        "bonus_all_stats"             "2 10 20 30 40"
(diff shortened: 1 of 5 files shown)


POCKETMONEY99:
I need to test this to make sure spirit vessel current % damage isn't hardcoded.

DarkoniusXNG:
> and I increased the Current HP % Damage to scale up to 5%.  This should make the item better at tank busting without it demolishing squishy targets.

I did the math and this change will not do what you want, I mean it will, but its insignificant, it doesn't make it better at ''tank busting''. It will maybe become relevant when its higher than 6% (like 7.20 Rupture which was 8% iirc), but I would leave it at 4% at all levels anyway because you don't get it for the dmg, you get it for the healing reduction.

PR #2795, opened 2019-11-20
Windranger nerfs

DarkoniusXNG:
Chemotherapy for OAA.


POCKETMONEY99:
Can we please wait until the patch has dropped. If valve arent messing it shouldn't be more then a week. I am 90% sure WR is going to get a rework. Cuz the hero dominates low MMR games while being weak in high tier pubs. Also im fairly certain this hero doesn't exist in pro matches. 

If they dont rework WR, change the cooldown reduction talent instead of nerfing her. (talents + OAA levels + Scaling Octarine Core is what makes most heroes chain stun your ass.) Spell block, cast range or armour all potential replacements. 

Pre 7.22 WR was one of the most balanced heroes in OAA imo. So if she needs nerfing post patch returning her to a state similliar to that would be preferable.

chrisinajar:
Individual PR's are effortless to revert, it's actually just 1 command. Let us fix the pubs now, then fix them again later, and then probably again after that.

PR #2798, opened 2019-11-28
Easy and medium camps experience bounties nerf

DarkoniusXNG:
This may not change anything because the formula for xp bounty increases per minute with this formula: 2 x (9 x min^2+17 x min + 607)/(607 x 3) and there are too much creeps in camps.


DarkoniusXNG:
This is just a band-aid for the problem (carries getting 10 levels in less than 5 minutes by farming only medium and easy camps). The real fix should be limiting number of creeps per camp and making them spawn in a specific way like in this issue: <https://github.com/OpenAngelArena/oaa/issues/2411> Map is out of date there but the idea is the same. And I think that idea should be implemented after next tournament/league because its a fundamental change to OAA.

POCKETMONEY99:
I dont think there will be another tornament in the near future. (No way we get enough players atm without a big prizepool or baumis influence.) The creep camps in #2411 is definetly the way to go. (I wouldn't do certain creeps like the ogre ancient camp but I think most of them are good ideas.). 

I think another issue is that non farming supports cannot keep up with carries anymore on exp. Ever since the tranquil cahnge to not kill dragons non farming heroes have struggled more. 

I think adding an EXP multiplier to Midas Spark will help alot.

PR #2799, opened 2019-11-29
Earthshaker Nerf

POCKETMONEY99:
Aftershock Duration Doesnt go past 1.8 seconds. Now Earthshaker Cannot Permastun People. (Will also be combined with octarine core nerfs and Dotas recent Aghs Nerf).


game/scripts/npc/abilities/earthshaker_aftershock.txt
```diff
@@ -17,7 +17,7 @@
 
     // Time
     //-------------------------------------------------------------------------------------------------------------
-    "AbilityDuration"                                     "0.6 0.9 1.2 1.5 1.8 2.1"
+    "AbilityDuration"                                     "0.6 0.9 1.2 1.5 1.5 1.5" //OAA
```
Trildar: Not like I've been keeping track of OAA developments, but the `//OAA` comment shouldn't be necessary unless the vanilla values are modified right?

DarkoniusXNG: Yeah, //OAA is not needed here.

PR #2800, opened 2019-11-29
Octarine Core Nerf

POCKETMONEY99:
Cooldown reduction no longer scales. Scaling CDR on octarine core pushes alot of heroes into broken territories. Several heroes have already been nerfed due to scaling octarine core + their talents/skills. Other heroes become impossible to deal with due to scaling octarine core. Flat 25% CDR is already great and this doesn't need to scale for the item to be good.

If a hero needs octarine just so there cooldowns aren't terrible they may need their base CD adjusting (Not in all cases, some heroes should have cooldown related weaknesses). 

Rescaled Octarine core intelligence gain. Same level 4 and 5 but better levels 2 and 3.


game/scripts/npc/items/item_octarine_core.txt
```diff
@@ -47,7 +47,7 @@
       "01"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "bonus_intelligence"                              "25 38 57 85 115"
+        "bonus_intelligence"                              "25 40 65 100 145"
```
DarkoniusXNG: Revert this. Octarine Core already gives more INT than it should.

POCKETMONEY99: Normalised the scaling. Goes +15 +20 +25 +30 INT. Same levels 4 and 5 but better levels 2 and 3 by 2 and 3 INT respectively.

lovesmuffins:
Do keep in mind the effect this will have on satanic core.

DarkoniusXNG:
This doesn't affect Satanic Core in any way.

DarkoniusXNG:
@chrisinajar Please merge this when you get the chance. This PR solves so many balance problems.

PR #2801, opened 2019-11-29
Death prophet Adjustments

POCKETMONEY99:
The idea of these changes is to make death prohpet playable in the early game while not being an unstoppable monster once she reaches OAA levels. 

Set the cooldown of crypt swarm to 5 seconds at all levels. This will help increase her farming speed early game. (This hero abilities do not let her roam well). 

Rescaled E damage. Is now slightly better levels 2-4. The Same at level 5 and worse at level 6. The damage drop off is extremly notciable early game while this spell is insane in the late game. 

(These changes will also be combined with octarine core nerf.)


DarkoniusXNG:
I still don't like the cd decrease on Q. Other stuff are fine I guess.

PR #2803, opened 2019-11-29
Lina buffs (Updated)

POCKETMONEY99:
Increased damage on her Q and W (in patch 6.83 her W damage was exaclty that value). This will improve her early game roam and farming speed. 

Improve light strike array talent damage

Improved linas health talent.

Improved linas attack damage talent to make right clcik lina more viable.


game/scripts/npc/abilities/lina_dragon_slave.txt
```diff
@@ -27,7 +27,7 @@
 
     // Damage.
     //-------------------------------------------------------------------------------------------------------------
-    "AbilityDamage"                                       "85 160 235 310 660 1060"
+    "AbilityDamage"                                       "125 200 275 350 750 1150" //OAA
```
DarkoniusXNG: It seems you need a reminder why this is bad. You would say the spell is the same as DP Crypt Swarm so lets see: Lina has a spell amp talent (DP doesn't). Lina has a lvl 25 talent that reduces cd of this spell (DP doesn't, DPS with Dragon Slave with this change will be way way higher than DP Crypt Swarm). Lina has a stun and she can land this spell more reliably than DP. Dragon Slave has a longer cast range/travel distance, so it is better than DP Crypt Swarm. Another factors why this is bad: Patch 7.23 introduces 110% spell amp (veil +bloodstone + kaya and yasha not counting spell amp from INT and spell amp from bloodstone charges and not counting Ethereal Blade as well). So this spell definitely deals damage.

POCKETMONEY99: DP Q is worse then lina Q post level 25. However DP is a raid boss, whereas lina is supposed glass cannon. She doesnt have enough damage to warrant her being so squishy. 

Unless OAA is different, bloodstone and kaya don't stack. (Im not sure if this is a bug or not. Kaya variants seem a little weaker now). Like how SnY and Manta style yasha MS bonuses dont 

These buffs are very modest. The only time you will notice the damage buff is early game. Im fairly certain there was a time in dota where lina did more spell damage then she does in OAA.

DarkoniusXNG: Bloodstone and Kaya stack in OAA.

DarkoniusXNG: Dragon Slave damage values from 6.83 to 7.06e were 110/180/250/320. You can change it to that. For OAA values you can use 640/960 (1040) or keep the current values: 660/1060.

game/scripts/npc/abilities/talents/lina_talents.txt
```diff
@@ -69,7 +69,7 @@
 			"01"
 			{
 				"var_type"					"FIELD_INTEGER"
-				"value"				"140"
+				"value"				"500" //OAA
```
DarkoniusXNG: Since this is level 15 talent. 500 seems too much. 280 or 300 seems more fair.

POCKETMONEY99: ok

POCKETMONEY99:
> 
> 
> I will approve this PR if you revert changes to Laguna Blade and if you put 110/180/250/320/640/1280 for Dragon Slave damage (1280 is probably too high but you want a glass cannon that does million damage per second so lets have it).

Done

PR #2805, opened 2019-11-29
Radiance Adjustment

POCKETMONEY99:
Nerfed the level 2 Radiance Cheese
Increased the damage this item gives level 3-5. It now matches other items such as desolator and nullifier with 200 damage. This will help improve the items viability on heroes such as spectre (not by much but more damage is more damage).

### DIFF
--- a/game/scripts/npc/items/item_radiance.txt
+++ b/game/scripts/npc/items/item_radiance.txt
@@ -55,17 +55,17 @@
       "01"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "bonus_damage"                                    "65 80 100 125 160"
+        "bonus_damage"                                    "65 80 120 160 200"
       }
       "03" //has to be wrong in this case else dota will get confused
       {
         "var_type"                                        "FIELD_INTEGER"
-        "aura_damage"                                     "60 90 150 240 360" 
+        "aura_damage"                                     "60 75 150 240 360"
       }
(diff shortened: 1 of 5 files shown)


game/scripts/npc/items/item_radiance.txt
```diff
@@ -55,17 +55,17 @@
       "01"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "bonus_damage"                                    "65 80 100 125 160"
+        "bonus_damage"                                    "65 80 120 160 200"
```
DarkoniusXNG: I would put: 65/85/115/155/205

game/scripts/npc/items/item_radiance.txt
```diff
@@ -55,17 +55,17 @@
...
       }
       "03" //has to be wrong in this case else dota will get confused
       {
         "var_type"                                        "FIELD_INTEGER"
-        "aura_damage"                                     "60 90 150 240 360" 
+        "aura_damage"                                     "60 75 150 240 360"
```
DarkoniusXNG: I know radiance 2 is powerful but that's only because first capture point still exists. Also this number ruins the "aesthetics" of the scaling :(

POCKETMONEY99: If first cap gets removed it can be reverted.

PR #2810, opened 2019-11-30
Ogre Magi Nerfs

POCKETMONEY99:
This hero was once fairly balanced however dota changes to multicast have turned this hero into a permastunning tyrant. He can also multicast items which is extreemly strong considering your always 6 slotted. 
 Fireblast CD doesnt scale into OAA levels. 
Decreased slow percentage and slow duration on ignite at OAA levels. However the damage has been increased. 
Decreased 4x multicast chance at level 5. Now you have to outplay the enemies harder to get 4x multicast. I dont think ogre could lose late game teamfights with level 50 and a hex. 

This hero still can perma stun you, but he will need to be more lucky.


game/scripts/npc/abilities/ogre_magi_ignite.txt
```diff
@@ -37,17 +37,17 @@
...
       {
         "var_type"                                        "FIELD_FLOAT"
-        "duration"                                        "5 6 7 8 10 15"
+        "duration"                                        "5 6 7 8 9 10"
       }
       "02"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "burn_damage"                                     "26 34 42 50 90 130"
+        "burn_damage"                                     "26 34 42 50 150 250"
```
DarkoniusXNG: This is a buff actually damage-wise. Also you should not change this because of the new mechanic: "Ignite hitting an already affected unit now adds to the duration, rather than just refreshing it". In other words its infinite aids unless you dispel it.

game/scripts/npc/abilities/ogre_magi_fireblast.txt
```diff
@@ -27,7 +27,7 @@
 
     // Time
     //-------------------------------------------------------------------------------------------------------------
-    "AbilityCooldown"                                     "11 10 9 8 7 6"
+    "AbilityCooldown"                                     "11 10 9 8 8 8"
```
DarkoniusXNG: Old multicast reduced cd of this to 4 or something ridiculous fyi,

POCKETMONEY99: old multicast didnt extend stun duration. Also I think getting past level 30 on supports was unheard of at that time, so you never saw the horror.

game/scripts/npc/abilities/ogre_magi_ignite.txt
POCKETMONEY99: Reduced buff. Same damage level 5 and more level 6. (Pipe + Luciences will outheal the damage). Isn't it good that im nerfing the spell duration and slow due to the ignite change?.

DarkoniusXNG: If I don't comment on that then its good :D He definitely needed that. I remember when it was 20 seconds or something at max level.

POCKETMONEY99: yikes

PR #2813, opened 2019-12-06
Reduce Giant Form Bonus size

POCKETMONEY99:
Reducing the bonus size you gain from giant form. Heroes with large models become too large when giant form is activated, making it extremely hard to click other units near the target. (Mainly just so Tiny doesnt become the worldender. There are other heroes which would gain huge bonus size but giant form isn't purchased on them).


DarkoniusXNG:
I will close this because its included in #2837

PR #2818, opened 2019-12-10
Remove Charge Bkb

POCKETMONEY99:
New players don't even purchase this item.


lovesmuffins:
Honestly don't think this is right. Should've had it combine alongside the new holy locket.

pretty much, it'd be a burst of health and mana as well as a duration based on the charges (max of 20 still). plus the passive healing boost. I think that could pot. be a good item

chrisinajar:
that'd be dope, but until then the item is a noob trap and is bad. happy to re-add it as a good item!

PR #2821, opened 2019-12-11
First Capture Point changes

DarkoniusXNG:
* First capture point now spawns at ~~15~~ 10 minutes.
* First capture point not on highground anymore.
* ~~First capture point now gives tier 2 core.~~

First capture point kinda lost its purpose when we added sparks. You don't have a need to upgrade farming boots anymore. These days first capture points make some heroes overpowered and unbearable in the early game because of Radiance 2 rush (Ember Spirit, Sand King). Makes carries snowball harder because of Assault Cuirass 2 or Desolator 2 rush. On the other side supports can upgrade Ward Stack to lvl 2 or buy Far Sight with that core and other than that I don't see any benefit to have first capture point at 5 minutes in the game.

P.S. If we remove capture point at 5 min, people will stop upgrading Desolator into Battlefury (Cleaver) right away.


chrisinajar:
this is great although the boots all scale on the same constant, so that's gotta be changed at the same time..

chrisinajar:
Wait... this ruins the core economy... There's 2 fewer tier 1 cores on the map for the number of items teams need to build.

DarkoniusXNG:
I changed first capture point timing to 10 minutes and they give tier 1 cores again. But they are not safe capture points anymore - not on highground.

PR #2825, opened 2019-12-14
7.23d + 7.23e

DarkoniusXNG:


DarkoniusXNG:
http://www.dota2.com/patches/7.23d

DarkoniusXNG:
http://www.dota2.com/patches/7.23e

PR #2826, opened 2019-12-19
Grimstroke Buffs

POCKETMONEY99:
Reduced Cooldown on his W and E to improve his early game roaming. 
Increased the damage that his E does. However I reduced the movespeed bonus as currently at OAA levels you have little chance of not getting max duration stunned with +200 radius and all the extra movespeed.


game/scripts/npc/abilities/grimstroke_ink_creature.txt
```diff
@@ -28,7 +28,7 @@
 
     // Time
     //-------------------------------------------------------------------------------------------------------------
-    "AbilityCooldown"           "32 28 24 20 20 20"
+    "AbilityCooldown"           "20" //OAA
```
DarkoniusXNG: Cooldown refreshes when Phantom survives the whole duration. So cooldown could be 60 seconds and the ability would still be fine. If you trying to improve early game cooldown this won't change much because you want to max this ability anyway (reason: number of hits required and dmg increases per level). That's my reasoning at least.

PR #2853, opened 2020-03-04
Chatterjee buffs

DarkoniusXNG:
* Base Attack Speed increased from 90 to 100.
* Electric Shield now blocks damage after reductions (after armor and magic resistance).
* Electric Shield damage block changed from 10/20/30/40/50/60% to 40%.
* Electric Shield damage radius increased from 300 to 400.
* Electric Shield damage interval improved from 0.5 to 0.2
* Cleansing Shock now bounces without Aghanim Scepter. Number of bounces is 0/1/2/3/3.
* Cleansing Shock debuff duration is no longer affected by status resistance. Slow value is affected instead like with all other slow debuffs in the game.
* Cleansing Shock Aghanim Scepter reworked: Increases number of bounces to 1/2/3/4/5 and deals 1000 damage to illusions and summoned units that are not spell-immune.
* Changed talents:
- Level 15 talent changed from +15 STR to +90 Damage.
- Level 20 talent changed from +800 mana to Cleansing Shock always cast on yourself.
- Level 25 talent changed from Cleansing Shock always cast on yourself to +1000 mana.


chrisinajar:
`Electric Shield damage block changed from 10/20/30/40/50/60% to 40%.` this, surprisingly, is a massive nerf.

You want the damage block to be very low at early levels else a single hit pops it and it goes away, you don't have the mana pool to sustain it. Instead, early game it's used purely for damage while you build up your mana pool.

PR #2870, opened 2020-03-29
Nullifier buff

DarkoniusXNG:
Since it doesn't mute anymore, it can be an item that doesn't need a core.


lovesmuffins:
it constantly dispels

PR #2875, opened 2020-04-04
Fix AI of simple bosses during duels

DarkoniusXNG:
Simple bosses in some edge cases completely stop acting when duel ends, because boss was aggroed before the duel.


chrisinajar:
Are you sure about this? Seems like it'd allow the boss to still move around and stuff, where it really should just freeze still the whole time..

PR #2880, opened 2020-04-07
Crystal maiden Buffs

POCKETMONEY99:
Improved Arcane Aura Mana Regen, Is now a really good value point, offering your hero +6 mana regen with one skill point. This will make roaming alot easier. 

Heavily increased damage Crystal nova does. Especially at OAA levels. This is to compensate for frostbites poor scaling at OAA levels. (In theory she would have 2 1k damage nukes at max level, so I increased the damage on her Q to replicate this damage output).


DarkoniusXNG:
STAHP buffing this hero

lovesmuffins:
can u explain why you think cm needs buffs lol?

PR #2888, opened 2020-04-11
Neutral Items tweaks

DarkoniusXNG:
* Tier 1 drop chance reduced from 14% to 2%
* Other tiers drop chance reduced from 10% to 1%.
* Greater Faerie Fire is now tier 2.
* Titan Sliver removed.
* Added Minotaur Horn and Witless Shako to tier 3.


POCKETMONEY99:
Maybe remove elixer aswell? I feel that item is irrelevant with bottles dropping from netruals.

DarkoniusXNG:
Its like a super bottle. Consumable neutral items are tier 1s on purpose. I think its fine.

PR #2889, opened 2020-04-12
Luna Buffs

POCKETMONEY99:
Improvements aimed at improving luna as both a carry and caster. 

Reduced Mana Cost on Lucent Beam and Eclipse at OAA levels. The mana cost doesnt affect caster luna but its crippling for carry luna. This spell won't be dealing any damage on carry luna so a mana cost reduction is ok. 

Reverted the nerfs to aghs eclipse cast range and the talents mini-stun duration. What I wanted to achieve on esclipse (0.3 fire rate with reduced max beams per target) never got into the game so these nerfs are unessecary. Dagger of Moriah has also been "removed" so the use of that item + bkb is gone. 

Increased Moon Glaives bounces by 2 at each level. Lunas farming speed is very underwhelming because of how many creeps spawn in a camp. This change should help luna flash farm faster. It also allows luna to deal more damage in teamfights. 

Increased Luna Blessing armour bonus by 5% at all levels. (its back to dota values). The armour bonus is negilable until late game. Even late game only heroes with high agility get a decent amount of bonus armour.


game/scripts/npc/abilities/luna_moon_glaive.txt
```diff
@@ -24,7 +24,7 @@
       "02"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "bounces"                                         "1 2 3 4 5 6" //OAA
+        "bounces"                                         "3 4 5 6 7 8" //OAA
```
lovesmuffins, line 27: lmao this is pushing it too far.

DarkoniusXNG: Every Moon Glaive bounce can crit, remember?

POCKETMONEY99: Ive decreased the bounces by 1 at each level. It should start off higher then 1 bounce at level 1 though. This heroes farming speed is bad and I don't think moon glaives is a crazy carry ability vs a full team of 6 slotted heroes (crimson guard hard counters it).

lovesmuffins:
I guess I am ok with trying it.

PR #2892, opened 2020-04-16
Sumicenz

warpdragon:
Item icons for Darkonius


DarkoniusXNG:
Thanks for adding dots.
What are blue and purple skull for?
 I also wanted Horn of Sorcery but that ring looks really cool so I will rename the new item to Ring of Sorcery.
And what about Silver Staff?

PR #2896, opened 2020-04-17
Boss resistance changes

DarkoniusXNG:
* Bosses now use new armor formula (default dota fromula)
* Bosses now have minimum armor (their starting armor -20)
* Bosses now take bonus damage from abilities that give bonus dmg to buildings (Leshrac Edict, Tiny Tree Grab, Lone Druid Bear Demolish...)
* DP Spirit Siphon and Huskar Life Break dont do dmg to bosses anymore.


TheUnholyWrath:
you forgot pugna...

DarkoniusXNG:
Just in case people think you are not meme-ing: I will not nerf Pugna and Techies damage to bosses just because their abilities do less damage to buildings in vanilla. :)

DarkoniusXNG:
This is finished. I just need to test how it behaves on the client.

PR #2897, opened 2020-04-18
Techies buffs

lovesmuffins:


lovesmuffins:
> I dont care what people say anymore. Techies doesn't need buffs. He has 2 bad talents yeah but other than that he is perfectly fine. Revert those activation delays and cast points. Those things are the only things making him balanced.

remove mine movement speed talent > add these buffs = a fun hero that isn't completely BS to play against.

His mine movement speed talent is very similar to storm's old talent and other suck situations. It completely defines what the hero's playstyle is and it's complete BS to me. These Buffs allowed me to be active in fights rather than having to stack mines prior to them. Techies is complete ass in duels, these fixes are to make techies an overall better hero. The talent is the ONLY reason he is good right now.

POCKETMONEY99:
You should remove the all the buffs to his cast points and trap activation times. The Blast off damage talent buff is good and his CDR talent is probably fine. While I previously liked the idea of "battle techies" (I even hoped the dota 2 devs wold rework him into that) that isn't really techies anymore. Techies should remain as a high damage area denial hero. His spells also need to remain as traps and not just AOE spells, because all his spells are OP if they are easily usable AOE spells (no long cast points or long activation delay). 

His mines movespeed talent is too good but techies isnt great in OAA so it should be left unchanged for now.

lovesmuffins:
> You should remove the all the buffs to his cast points and trap activation times. The Blast off damage talent buff is good and his CDR talent is probably fine. While I previously liked the idea of "battle techies" (I even hoped the dota 2 devs wold rework him into that) that isn't really techies anymore. Techies should remain as a high damage area denial hero. His spells also need to remain as traps and not just AOE spells, because all his spells are OP if they are easily usable AOE spells (no long cast points or long activation delay).
> 
> His mines movespeed talent is too good but techies isnt great in OAA so it should be left unchanged for now.

wouldn't you rather remove the dumb talent and have the activation buffs?

DarkoniusXNG:
Since you are stubborn, I will close this.

PR #2898, opened 2020-04-19
Bloodstone HP and Mana nerf

DarkoniusXNG:


lovesmuffins:
I really don't think we should change this until we know whether or not spellcasters got buffed/nerfed from 7.26

LeighJoseph:
I'm pretty sure they have been buffed

DarkoniusXNG:
It needs a nerf because everyone buys it. If everyone buys it means its op.
It doesn't matter if Valve changes anything, amount of HP and mana this item gives should be equal to Octarine Core.

POCKETMONEY99:
This is good. The HP and Mana Increase was only relevant before bloodstone gave spell amp.

PR #2901, opened 2020-04-20
Tier 5 Temple Guardian Boss Nerf

POCKETMONEY99:
Temple Guardians are still by far the hardest boss to do. However I think tier 4 guardians are managable. Tier 5 guardians are very strong and most players actively avoid these bosses. They will often pick up multiple kills when attempted.

Reduced Tier 4 enrage hammer smash from 6400 to 6000. 

Reduced Tier 5 Health From 16000 to 15000. 
Reduced Tier 5 Hammer Throw Damage from 3000 to 2500. 
Reduced Tier 5 Purification heal/damage to 1500

### DIFF
--- a/game/scripts/npc/abilities/siltbreaker/temple_guardian_hammer_throw_tier5.txt
+++ b/game/scripts/npc/abilities/siltbreaker/temple_guardian_hammer_throw_tier5.txt
@@ -62,7 +62,7 @@
       "04"
       {
         "var_type"                  "FIELD_INTEGER"
-        "hammer_damage"             "3000"
+        "hammer_damage"             "2500"
       }
     }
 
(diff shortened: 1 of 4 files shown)


DarkoniusXNG:
I think damage numbers of abilities you nerf are fine actually. Problems appear when only 1 temple guardian remains: he heals during Omni Party with boss regen. I would solve this by disabling all regen (or just his passive boss regen) during Omni party. Second problem is bkb-piercing damage of Omni Party.

POCKETMONEY99:
> 
> 
> I think damage numbers of abilities you nerf are fine actually. Problems appear when only 1 temple guardian remains: he heals during Omni Party with boss regen. I would solve this by disabling all regen (or just his passive boss regen) during Omni party. Second problem is bkb-piercing damage of Omni Party.

Sounds like a good idea but I don't know how.

lovesmuffins:
guardians are not hard at all lol.

PR #2903, opened 2020-04-21
Ogre Boss: Ogre Seer Buffs

POCKETMONEY99:
Ogre Seers Die far too fast when doing the ogre boss. This basically removed a part of the boss because the ogre seers are dead within the first 10 seconds of the fight. Ogre seers can also be countered by silences, stuns and hexes which are quite common when doing tier 3+ bosses.

Gave Ogre Seers the boss resistance ability. I reduced their HP to compensate for this buff. Now the Ogre Seers are more durable. I also changed their armour and magic resistance to match the ogre boss. I gave their new HP by multiplying their old HP by [1-0.8 (the damage reduction)] = 0.2

Increased Ignite DPS on both tier 3 and tier 5 bosses. The damage currently isn't noticable and you can stand in the fire with no consequence. I reduced the duration on the tier 5 ignite damage duration to match the tier 3 boss. When you walk out of the fire the buff isn't supposed to linger for long.

### DIFF
--- a/game/scripts/npc/abilities/siltbreaker/ogre_seer_area_ignite.txt
+++ b/game/scripts/npc/abilities/siltbreaker/ogre_seer_area_ignite.txt
@@ -43,7 +43,7 @@
             "02"
             {
                 "var_type"          "FIELD_INTEGER"
-                "burn_damage"       "400"
+                "burn_damage"       "600"
             }
 
             "03"
--- a/game/scripts/npc/abilities/siltbreaker/ogre_seer_area_ignite.txt
+++ b/game/scripts/npc/abilities/siltbreaker/ogre_seer_area_ignite.txt
@@ -43,7 +43,7 @@
             "02"
             {
                 "var_type"          "FIELD_INTEGER"
-                "burn_damage"       "500"
+                "burn_damage"       "600"
             }
 
             "03"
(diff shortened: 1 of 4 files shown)


TheUnholyWrath:
if the boss resistance makes the way it acts the same as the actual boss, don't add it. its the only punish for the creep/helm abuse.

TheUnholyWrath:
the sweet spot is the biggest problem with this boss. if you can remove the sweet spot you can make the ogre seers harder to kill without changing them at all, i'd also rather you buff the movement speed slow over the fire damage

DarkoniusXNG:
Adding boss resistance and 40 armor to Ogres is definitely making them harder to kill. If the goal was to make this boss fight last longer you achieved it.

POCKETMONEY99:
> 
> 
> if the boss resistance makes the way it acts the same as the actual boss, don't add it. its the only punish for the creep/helm abuse.

POCKETMONEY99:
> 
> 
> the sweet spot is the biggest problem with this boss. if you can remove the sweet spot you can make the ogre seers harder to kill without changing them at all, i'd also rather you buff the movement speed slow over the fire damage

Movespeed slow is fine. Issue is that I overnerfed the fire damage. You can pretty much just stand in the fire with little consequences.

DarkoniusXNG:
Boss AI is not tied to boss resistance.

POCKETMONEY99:
Ogre Seers are pretty tanky after then armour change, and the ignite does better damage now that the default value for magic resistance is 25%. Mabye I will look at Ogres later.

PR #2904, opened 2020-04-21
Swiper Boss Damage Type Change

POCKETMONEY99:
Changed the damage type on the Swiper boss from pure to physical. I increased the damage this boss does to compensate for this change. Additionally I buffed the minimum damage for reapers rush. 

The reason for this change is that most heroes apart from high HP strength heroes with heart get punished hard by Swiper. Swiper gets player kills. Additionally, during boss contests swiper is one of the most dangerous bosses, with his ranged abilities one shotting most heroes. 

Another reason for this change is so the boss falls off in powerlevel more reasonably like all the other bosses in the game when you get higher tier items etc. (Just for reference some of sven bosses attacks hit harder then temple guardians.)

The boss still hits very hard, dealing more damage then tier 3 bosses however everyone has slightly more EHP vs this boss, allowing for a slightly higher marign for error. Swiper is also alot easier to kite then tier 3 bosses, so its high damage is acceptable.

### DIFF
--- a/game/scripts/npc/abilities/swiper/boss_swiper_backswipe.txt
+++ b/game/scripts/npc/abilities/swiper/boss_swiper_backswipe.txt
@@ -16,7 +16,7 @@
         "AbilityUnitTargetTeam"         "DOTA_UNIT_TARGET_TEAM_ENEMY"
         "AbilityUnitTargetType"         "DOTA_UNIT_TARGET_HERO"
         "AbilityUnitTargetFlags"        "DOTA_UNIT_TARGET_FLAG_MAGIC_IMMUNE_ENEMIES"
-        "AbilityUnitDamageType"         "DAMAGE_TYPE_PURE"
+        "AbilityUnitDamageType"         "DAMAGE_TYPE_PHYSICAL"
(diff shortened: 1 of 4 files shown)


TheUnholyWrath:
why are we specifically rewarding agi cores for being bad?

DarkoniusXNG:
Swiper boss was perfectly fine when you nerfed the damage. Swiper boss is also fine now but you want heroes with 100 armor to take 800 damage from it. This change supports laziness, bad/new players and bad agi cores. Thematically, damage type **should be physical** but damage should be a lot higher. 10000 for example. But you will say that's too much.

POCKETMONEY99:
The intention of this change is to make everyone lose a simillar amount of HP when then hit the bosses. Rather then heavily punishing players which can't afford to go full tank item builds. 

This change doesn't support lazyiness, you are being lazy if you decide to not dodge the bosses attacks to take a huge amount of damage. I just want it so everyone can make a few mistakes, rather then it only being tanks that can make mistakes around sven boss. If enemies decide to contest you, if your being lazy your nearly dead. 

Also the damage is high enough to the point where you can't just stand and tank the boss.

PR #2905, opened 2020-04-21
Manta Style Buffs (Updated to 7.26a)

POCKETMONEY99:
Increased the damage manta style illusions do. Manta style is a very weak item at the moment, with the illusions dealing no damage. 

Damage now scales on melee "33 38 43 48 53"
Damage now scales on ranged "28 33 38 43 48"

### DIFF
--- a/game/scripts/npc/items/item_manta.txt
+++ b/game/scripts/npc/items/item_manta.txt
@@ -78,12 +78,12 @@
       "08"
       {
         "var_type"                                        "FIELD_INTEGER"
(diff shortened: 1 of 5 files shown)


DarkoniusXNG:
You are over-buffing the damage numbers. I am also not a fan of making lvl 1 item better than vanilla.

POCKETMONEY99:
Reduced the scaling to +5% at each level and the level 1 damage matches dota values.

PR #2907, opened 2020-04-22
Tinker tweaks

DarkoniusXNG:
With 7.26 and removal of magic resistance from Strength, all dot spells got indirectly buffed. These are reasonable nerfs. It will not destroy a hero or anything.
* March of the Machines damage nerf
* Rearm mana cost nerf (I watched Tinkers spam spells and they were not losing mana at all - not cool)
* Fixed Rearm tooltip not showing percentage mana cost.


DarkoniusXNG:
After some backlash I decided to buff March of the Machines talent. Laser talent buff is also temporary until we add a new talent.

PR #2913, opened 2020-04-23
Clinkz Changes

lovesmuffins:
Late-game clinkz ult does not scale, this tends to fix him and reduce lag where necessary. Death Pact receives a small cd buff to help clinkz when he is from behind.


DarkoniusXNG:
ah no no no no. I am not falling for this again.

PR #2923, opened 2020-04-30
Windranger buffs

POCKETMONEY99:
Alot of things have changed in dota and OAA which have affected windranger. Windrangers 2 main issues were that you could permastun people and she was impossible to catch. She got nerfed a few months ago and hasn't been good since.

The things that have changed that affect WR:
- OAA Octarine Core CDR nerf.
- Her talents got shuffled. She can't get the windrun invis and CDR talent until level 50. 
- Her CDR talent got nerfed
- Nullifier changes (she can't windrun at all)
- She gets no free movespeed from agility so she is a bit slower. 

Buffs
- Reduced powershot damage reduction from 20% to 10%. I have reduced the damage penatly because the main reason the damage penalty exists is because in dota your wave clear would be too strong if there was no damage penatly. This will allow windranger to clear camps easier. A simillar change happened to ember spirit. 
- Changed windrun values back to dota values. If windrun is too good the bonus aghs movespeed values should be reduced instead of nerfing the cooldown (outside of the bonus movespeed this spell is not good). 
- Rescaled Focus fire cooldown to "50 40 30 30 30". (It now shares the same cooldown as its dota 2 level 3 ultimate instead of being 10 seconds higher).


game/scripts/npc/abilities/windrunner_powershot.txt
```diff
@@ -39,7 +39,7 @@
       "02"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "damage_reduction"                                "20"
+        "damage_reduction"                                "0" //OAA
```
DarkoniusXNG, line 42: I am not a fan of removing parts of spells that make those spells unique. I would scale it: 20/15/10/5/0/0.

POCKETMONEY99: I have made the damage reduction 10% at all levels. There is basically no point in making this change if its 20% at level 1. Tbh if the damage penatly is 0% is it a "better" version of KOTL illuminate without aghs for early game farming.

PR #2928, opened 2020-05-04
Make First Duel Start Later

POCKETMONEY99:
Changed the timing of first duel from 0 seconds to 7 minutes. First duel at 0 minutes has always been a questionable feature of OAA. At first it was insanely important, but now first duel is pointless. Arguably losing first duel is optimal because first wanderer will roam to your side if your behind on points.  

I have also made a few other minor changes. Increased the pregame timer by 10 seconds. You can't buy items during strategy time so I have given players a few more seconds to buy items and get into position before the game starts. Alot of the time first duel would start with players still choosing starting items in base. 

Increased base hero kill exp from 20 to 40. Now that the game doesnt start with a 5v5, early game kills can be made more vauable. Currently creeps are far more valuable then heroes.


DarkoniusXNG:
We discussed this multiple times about this. It seems you invalue our opinions about it.
> Changed the timing of first duel from 0 seconds to 7 minutes. 

I would change it to 5 minutes if I would ever change it. 
> First duel at 0 minutes has always been a questionable feature of OAA.

I think you are the only one that question it. Its a mechanic for new players to learn that this game is about fighting (it works for new players that have at least some IQ). Its also a fun factor in pubs. First duel can be delayed in captain's mode if you care that much.
> At first it was insanely important, but now first duel is pointless. 

Duel was never insanely important. ''Wow he got lvl 2 just because he was alone and killed everyone.'' You can still do that.
> losing first duel is optimal because first wanderer will roam to your side if your behind on points.

Its not an optimal thing, its a strat that can backfire easily.
> Increased the pregame timer by 10 seconds.

I don't how this affects base protection time (time during you are not allowed to leave the base highground). Chris needs to see this.

>Increased base hero kill exp from 20 to 40. 

I discussed hero kill exp with Pigeon. I gave him all the numbers to analyze. One idea came up: disable hero kill exp in duels and then increase these numbers. Also before changing anything mentioned in this pull request, DOUBLE FIRST CREEP SPAWN needs to be fixed first.

chrisinajar:
> Arguably losing first duel is optimal because first wanderer will roam to your side if your behind on points.
Wanderer doesn't favor a side until the score difference is at least 10, so you'd also need to feed at least 4 kills for that to work. I don't think that's optimal at all, especially because an enemy hero must get the reward for the score to change, plus playing around that strat is very very easy.

POCKETMONEY99:
> > Changed the timing of first duel from 0 seconds to 7 minutes.
> 
> I would change it to 5 minutes if I would ever change it.

I chose 7 minutes because this is near the time that second duel would start (which I think should be the first duel). It gives players time to get the items and levels they need to give them a fighting change to win duel. This also includes ultimates which is very important for 5v5 teamfights. Currently first duel is decided by whoever has the best level 1 skills and higher EHP. 

> I think you are the only one that question it. Its a mechanic for new players to learn that this game is about fighting (it works for new players that have at least some IQ). Its also a fun factor in pubs. First duel can be delayed in captain's mode if you care that much.

First duel at 0 seconds has failed as a mechanic if it is supposed to teach new players that the game is about fighting. Most new players will farm highground unless somebody explains to them how to play the game. Its not intuitive to roam around the map and fight in OAA, so most new players will farm highground. Usually these players have come from another AA game. I assume in these games farming highground is a viable strategy. 

> I don't how this affects base protection time (time during you are not allowed to leave the base highground). Chris needs to see this.

I have tested it. Base proctection time will always remain constant.

POCKETMONEY99:
> I discussed hero kill exp with Pigeon. I gave him all the numbers to analyze. One idea came up: disable hero kill exp in duels and then increase these numbers. Also before changing anything mentioned in this pull request, DOUBLE FIRST CREEP SPAWN needs to be fixed first.

I have been looking at double creep spawn and I think the issue is that INITIAL_CREEP_DELAY in the settings.lua file seems to not work. I tried increasing it but nothing changed. Creeps will spawn as soon as the game starts. Even increasing the pregame timer has no effect on when the creeps spawn. They will still spawn every 60 seconds even if the pregame timer is 2 minutes. It appears the game always starts at 0 seconds. 

image

Im assuming it has something to do with this line in the spawner.lua file but when I tinkered with it only changed the creep respawn interval not the initial creep spawn time.

PR #2930, opened 2020-05-08
Disables leaving penalties temporarily

DarkoniusXNG:
Until we are sure crashes are gone with Void Spirit.


POCKETMONEY99:
This needs commiting ASAP

PR #2937, opened 2020-05-11
Silver Edge buffs

DarkoniusXNG:
* Bonus attack speed increased from 30/33/36/40/45 to 30/40/50/60/70.
* Bonus all stats rescaled from 15/19/24/30/38 to 15/20/25/30/35.

Used same scaling rule that was used on majority of the items. This item was forgotten for some reason.

### DIFF
--- a/game/scripts/npc/items/item_silver_edge.txt
+++ b/game/scripts/npc/items/item_silver_edge.txt
@@ -61,7 +61,7 @@
       "02"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "bonus_attack_speed"                              "30 35 40 45 50"
+        "bonus_attack_speed"                              "30 40 50 60 70"
       }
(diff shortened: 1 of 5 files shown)


POCKETMONEY99:
Silver Edge is forgotton because the item stats are bad. Additionally invis is very easy to counter with a ward stack or farsight, which both make your hero more durable. 

These buffs arent enough. I would increase the attack speed bonus to "30 40 50 60 70" and the stat bonus to "15 25 35 45 55" then see how the item is after. Silver edge is supposed to be a fairly good stat item, which it isnt currently.

LeighJoseph:
2 attack speed buff omega lul

DarkoniusXNG:
Buffed attack speed. Didnt buff stats because of infamous Pigeon patch.

PR #2940, opened 2020-05-14
Capture Point timings tweaks and some code improvements

DarkoniusXNG:
* First capture point now spawns at 11 minutes (after max duration duel, so there are no conflicts).
* Reduced capture points spawn interval from 10 to 8 minutes better match boss killing times.
* DOTA_MAX_TEAM_PLAYERS fixes (Added -1)
* Replaced 19 in for loops with DOTA_MAX_TEAM_PLAYERS-1. 19 is not the max playerID since forever.
* Added IsValidPlayerID(playerId) checks. I am not sure if it should be added in every loop with playerIDs.

### DIFF
--- a/game/scripts/vscripts/components/bottlepass/server.lua
+++ b/game/scripts/vscripts/components/bottlepass/server.lua
@@ -58,7 +58,7 @@ function Bottlepass:SendWinner (winner)
   local abandonedPlayers = {}
 
   local playerBySteamid = {}
-  for playerID = 0, DOTA_MAX_TEAM_PLAYERS do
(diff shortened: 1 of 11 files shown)


DarkoniusXNG:
Capture Points and Duel timings should be 3 minutes apart. When Capture Points spawn after duel, there is less chance for them to overlap with Duel timing. Why I am doing this? I believe that if second duel lasts too long, first capture point causes a server crash. I know some people want a Capture Point back at 5 minutes, but that will make pubs unbearable again because of heroes that rush radiance 2.

PR #2941, opened 2020-05-19
Fixing what we can

DarkoniusXNG:
* Necrophos Heartstopper KV fix
* Antimage hero KV file fix
* Lifestealer hero KV file fix
* Changed file structure for portraits.txt so it can be updated more easily (one file for vanilla models and one for custom models)
* Disabled remnants of custom vector targeting


DarkoniusXNG:
* Changed addoninfo.txt (BattlePass update changed encoding of addoninfo.txt
* Replaced DOTAItems with DOTAAbilities

DarkoniusXNG:
Penalties are disabled in a new addoninfo.txt

PR #2942, opened 2020-05-28
Update outdated GetDedicatedServerKey

DarkoniusXNG:


chrisinajar:
This'll change the key right?

PR #2979, opened 2020-09-01
Fix panorama errors, hero item builds and some clientside code refactor

DarkoniusXNG:


DarkoniusXNG:
I think I finally fixed chat during the hero selection.

PR #2991, opened 2020-10-14
Pull Staff Fixes

DarkoniusXNG:
* Pull Staff cast range rescaled from 800 to 600/700/800/900.
* Pull Staff max distance increased from 500/600/700/800 to 600/700/800/900.
* Pull Staff pull speed reduced from 1500 to 1200.
* Pull Staff cant target units in Black Hole, Chronosphere and Kinetic Field anymore.
* Pull Staff can't target Dueled heroes and Lassoed units anymore.
* Pull Staff doesn't actually let you target mentioned units above unlike Force Staff.
* It now uses motion controller buff instead of linear projectile -> effect should be instant when you use the item instead of having a weird delay or not working sometimes.
* Fixed particles not going away if used on a target already affected by Pull Staff (it was possible with Refresher and multiple Pull Staves).
* Pulling animation should be smoother and above ground the whole time now.
* Fixed not being able to destroy Pull Staff item when dropped on the ground.


chrisinajar:
```
Pull Staff cant target units in Black Hole, Chronosphere and Kinetic Field anymore.
Pull Staff can't target Dueled heroes and Lassoed units anymore.
```
that's just mean

DarkoniusXNG:
Its a bug that became a feature. I will buff it next week to compensate.

PR #3007, opened 2020-11-04
Error fixes

DarkoniusXNG:
* Fixed an error with boss resistance true sight when the boss is killed too fast.
* Fixed GetBehavior errors introduced with Diretide patch.
* Reflection Shard: Preventing Anti-Mage's Counter Spell and Mirror Shield reflected spells to be reflected by Reflection Shard. It was causing an infinite loop and server crash.
* Reflection Shard: Preventing Morphling Morph to be reflected by Reflection Shard. It was causing a server crash.
* Reflection Shard: Fixed custom unit-target abilities not being reflected.
* Added more ability nil checks to:
1) bloodstone.lua
2) greater_guardian_greaves.lua
3) satanic_core.lua
4) vampire.lua
* Bounty Runes: Fixed Alchemist passive not working with Bounty Runes (not multiplying the gold value for the Alchemist).
* Bounty Runes: Fixed an error when an entity without playerID picks up the rune.
* Bounty Runes: Added overhead number notification for the experience gained.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -2728,6 +2728,7 @@ read_globals = { -- these globals can only be accessed.
 "DOTA_UNIT_ORDER_HOLD_POSITION",
 "DOTA_UNIT_ORDER_DROP_ITEM",
 "DOTA_ModifyGold_Buyback",
+"DOTA_ModifyGold_BountyRune",
 "DOTA_HEROPICK_STATE_AP_SELECT",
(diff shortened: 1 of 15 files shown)


DarkoniusXNG:
* Fixed Slark being revealed with Boss True Sight when using Shadow Dance (Shadow Dance is supposed to give TrueSight immunity).
* Added overhead particle that indicates if you are being revealed from invisibility.

PR #3009, opened 2020-11-15
Faceless Void Aghanim Scepter update

DarkoniusXNG:
* Aghanim Scepter on Faceless Void now upgrades **Time Walk and Chronosphere** instead of upgrading Time Lock.
* Time Walk (scepter) now applies Time Lock upon reaching destination - vanilla aghs upgrade.
- Differences from vanilla aghs:
1) Time Lock will not be applied if affected enemy is immune to attacks (ethereal for example).
2) Time Lock will not be applied if Faceless Void is disarmed.
3) Time Lock goes on cooldown if it was procced with Time Walk.
* Chronosphere (scepter) now disables all evasion, healing and regen, and freezes cooldowns (of enemies only).

P.S. Previous Aghs upgrade was a placeholder (don't scream at me for ''ruining aghs''!). I was brainstorming with people what aghs upgrade is best for the hero, what aghs upgrade is best for the game and whats code-able without rewriting whole abilities. With this we will have a fun aghs. I removed some bullshit vanilla interactions (mentioned above). We will see if it needs more changes.


chrisinajar:
dope

PR #3014, opened 2020-11-19
Wanderer Updates

DarkoniusXNG:
* Wanderers are now more distinguishable (Level 2 - Snow Owl; Level 3 - Manta Ray)
* Wanderers now have abilities:
1. **Wanderer's Net** - tracking projectile, applies root, break and disables regen for 4 seconds. Wanderer uses this only when below 75% of max hp.
2. **Cleanse** - AoE damage, knockback and dispel. Kills player controlled wards, mines etc. Wanderer uses this only when below 50% of max hp and when more than 1 enemy is around.
3. **Sticky Blood (passive)** - Applies a debuff similar to Sticky Napalm to the attacker every time it gets damaged. Has a cooldown and damage threshold.
4. **Point Aversion (passive)** - Deals more damage to the team with more points.

* Fixed ''pet Wanderer bug'', I hope I didnt break something else in the proces


DarkoniusXNG:
* Decreased Level 1 and 2 Wanderer's attack damage by 500.
* Decreased Level 3 Wanderer's attack damage by 1000.
* Wanderer's attack damage doesn't vary from value to value anymore. Min and max damage are the same.
* Added a notification to all players when Wanderer is killed.

chrisinajar:
`Point Aversion (passive) - Deals more damage to the team with more points.` YASSS

PR #3029, opened 2020-12-06
Re-adding Meepo

lovesmuffins:
With talking with darkonius we have come up with ideas of meepo changes that'll make him viable in the game. These are just fixes to his overall kit and the small changes. Will be continue to be tweaked than merge with darkonius ult changes when he gets around to it (if ever).

Thoughts and Edits are welcomed!

I also probably forgot some things.


DarkoniusXNG:
This is OK now if you ask me -> it can be merged. I fixed all the bugs with clones that I know of. Meepo clones not getting custom boots is still a thing. I will fix that issue next week.

PR #3030, opened 2020-12-14
Another Wanderer update

DarkoniusXNG:
* Cleanse damage changed from 5000 to 5000+50% of max hp. Reminder: Its physical damage that can be blocked with spell immunity. And for reference, Swiper's damage is 6500.
* Added Siltbreaker boss protection to all Wanderers.
* Sticky Blood cooldown reduced from 3 to 1 second.


lovesmuffins:
can it be current HP not max hp cuz uh that's really good.

DarkoniusXNG:
It should be fine.

PR #3031, opened 2020-12-14
Blade of Judecca and status resistance

DarkoniusXNG:
* Blade of Judecca debuff is no longer affected by status resistance.


DarkoniusXNG:
It will be awkward if Chris merges this pull request and the commit just says Status resist. Confusing everyone that reads patch notes on the arcade page lmao.

PR #3044, opened 2021-01-06
Sohei 2021 Update

DarkoniusXNG:
* Removed Scepter upgrade for Dash
* Added Shard upgrade for Dash: "Sohei heals allies he dashed through for 100 plus 10% of his current health and pulls them along the way. Sohei and his allies are invulnerable during Dash and disjoint projectiles."
* Dash speed is now based on Sohei's movement speed at the moment of the cast. Dash speed = 7 x movement speed.
* Dash casting behavior changed: Sohei will no longer walk into range to cast Dash, he will always use Dash towards target location.
* Dash (cast) range reduced from 600 to 575.
* Dash: Sohei's health bar is now visible during Dash if he is vulnerable.
* Dash: Now it looks better when going over cliffs.
* New Scepter Upgrade: Grants **Quivering Palm** (new ability) - When activated, last attacked hero by Sohei will be affected by lesser Momentum Strike and damaged based on the difference between the hero's strength and Sohei's. Can only affect heroes attacked in the last 10 seconds. If there is no valid target, cooldown is reduced to 1 second. Quivering Palm debuff is dispellable. Damage is magical and its 100 + 15 x strength difference.
* Level 10 talent changed from +50 Damage to +20 Strength
* Level 15 talent Dash Invulnerability replaced with +20% Dash Slow.
* Level 20 talent +2s Momentum Strike Stun replaced with +275 Dash Cast Range.
* Level 25 talent +200 Flurry of Blows AoE with +2s Momentum Strike Stun.
* Wholeness of Body mana cost increased from 85 to 80/85/90/95/100/105.
* Wholeness of Body bonus magic resistance increased from 90% to 100%
* Wholeness of Body base heal rescaled from 90/160/230/300/600/900 to 100/150/200/250/500/1000.
* Momentum Strike cast point increased by 0.05
* Momentum Strike mana cost increased by 5.
* Momentum Strike collision of a unit affected by knockback now looks better against trees, buildings, blockers and other units.
* Flurry of Blows bonus damage is now applied only on the delayed instant attack and not on normal attacks made by Sohei during Flurry of Blows duration (Sohei can still attack but it will not do bonus damage). This also means Momentum Strike will not do extra damage if used during Flurry of Blows.


DarkoniusXNG:
* Momentum Strike crit damage increased from 150/180/210/240/270/300% to 150/190/230/270/310/350%.
* Level 25 talent +120% Momentum Strike Crit Damage reduced to +100%.

PR #3098, opened 2021-04-18
7.29b

POCKETMONEY99:
The new patch which came out. 

I put //OAA for huskars bezerkers blood skill, because he isnt very good in this game mode. Lifebreak cooldown now also gets reduced at OAA levels. 

Additionally valve have forgot to update the spirit bears savage roar duration to match the patch description. I have changed it to the values to satisfy travis, but this should quickly get fixed.


DarkoniusXNG:
Yup Valve are monkeys

PR #3106, opened 2021-04-20
Sacred Skull buffs

DarkoniusXNG:
* A Sacred Skull now also provides bonus hp regen: 3/6/9/12.
* Damage and heal per missing hp increased from 10/20/30/40 to 15/20/30/45.
* A Sacred Skull Death Flare now provides vision again. 800 flying vision at death location for 10 seconds.
* Fixed item icons not having correct number of pips.
* Simplified tooltip.


lovesmuffins:
current hp should be reducable by magic resistance or something. otherwise this item will always kinda be bad

DarkoniusXNG:
Its reduceable with damage reductions like Ursa Enrage. You are also missing the point of the item, it deals dmg and heals based on your missing hp, if you are full hp you don't do anything.

lovesmuffins:
> Its reduceable with damage reductions like Ursa Enrage. You are also missing the point of the item, it deals dmg and heals based on your missing hp, if you are full hp you don't do anything.

oh ya than Nvermind. 

I guess than I have no idea whats the point of the item. If anything item should just do damage/heal on your death based on your max hp difference.

DarkoniusXNG:
It already heals on death for 300 + max hp amount, so if your hero has 12k hp, on death it will heal allies for 12300.

PR #3109, opened 2021-04-20
MIDAS

lovesmuffins:


chrisinajar:
this seems like a bad idea?

DarkoniusXNG:
Yes. But Loves is stubborn and doesn't listen. I want to remove midas upgrades and buff lvl 1 midas.

lovesmuffins:
> this seems like a bad idea?

not bad idea. Suggested by community members. Darkonius wants to just get rid of item this is essentially compromise

PR #3112, opened 2021-04-20
Earth Spirit buff

lovesmuffins:
+25 is seemingly a very icefrog kind of buff (timber q and axe call) and the dmg per second has been increased at all levels. Max oaa level is a lot seemingly because the thought process is simple: If you're level 49 and don't have a dispel you are a monkey.

### DIFF
--- a/game/scripts/npc/abilities/earth_spirit_magnetize.txt
+++ b/game/scripts/npc/abilities/earth_spirit_magnetize.txt
@@ -37,7 +37,7 @@
       "01"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "cast_radius"                                     "350"
+        "cast_radius"                                     "375" //OAA
       }
       "02"
       {
--- a/game/scripts/npc/abilities/earth_spirit_magnetize.txt
+++ b/game/scripts/npc/abilities/earth_spirit_magnetize.txt
@@ -19,7 +19,7 @@
 
     // Casting
     //-------------------------------------------------------------------------------------------------------------
-    "AbilityCastRange"                                    "350"
+    "AbilityCastRange"                                    "375" //OAA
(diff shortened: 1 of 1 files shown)


lovesmuffins:
THIS IS ALL I THINK HE NEEDS TO BE A VIABLE HERO. Haters be damned

DarkoniusXNG (commented):
You need to increase cast_radius too.

game/scripts/npc/abilities/earth_spirit_magnetize.txt
```diff
@@ -19,7 +19,7 @@
 
     // Casting
     //-------------------------------------------------------------------------------------------------------------
-    "AbilityCastRange"                                    "350"
+    "AbilityCastRange"                                    "375" //OAA
```
POCKETMONEY99, line 22: You should leave the AOE unchanged. People are grouped up enough to the point where you should quickly be able to spread magnetise.

lovesmuffins: ok so does it matter than? 25 is just a nice QOL fix

lovesmuffins: like you're literally asking for earth spirit buffs and now you don't want them?

POCKETMONEY99: Why make this change though? It serves no purpose. The issue with magnetise is that it deals no damage.

DarkoniusXNG: The issue with Magnetise is that its easily dispellable. So any change to Magnetise is useless so whatever.

lovesmuffins:
FULLY FUNCTIONAL AND READY TO PUSH!!!

PR #3129, opened 2021-05-13
New map and enabled dota arena

DarkoniusXNG:


DarkoniusXNG:
Also some creep camp adjustments. Chris don't forget after building the map to create a minimap (command is dota_minimap_create or something) .

PR #3133, opened 2021-05-16
Visage Buffs and Shard fix

lovesmuffins:
Small changes to Q to ease my eyes of the pain and too allow oaa values to be a bit higher, as it should be.

Gravekeeper cloak heavy nerfs reverted since the max dmg cap is added to ability.

Buffs to familiar to cosign with all the other player units that have been buffed. Familiars will now do bit more dmg and be a bit more tanky but still should be killable in all respective manners. Like to see what people think

### DIFF
--- a/game/scripts/npc/abilities/visage_grave_chill.txt
+++ b/game/scripts/npc/abilities/visage_grave_chill.txt
@@ -48,7 +48,7 @@
       "03"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "attackspeed_bonus"                               "34 46 58 70 82 94"
+        "attackspeed_bonus"                               "36 48 60 72 96 120" //OAA
       }
       "04"
       {
--- a/game/scripts/npc/abilities/visage_summon_familiars_oaa.txt
+++ b/game/scripts/npc/abilities/visage_summon_familiars_oaa.txt
@@ -70,6 +70,7 @@
         "var_type"                                        "FIELD_INTEGER"
         "scepter_total_familiars"                         "3"
         "LinkedSpecialBonus"                              "special_bonus_unique_visage_6"
+        "scepter_reqired"                                 "1"
       }
     }
   }
--- a/game/scripts/npc/abilities/visage_summon_familiars_oaa.txt
+++ b/game/scripts/npc/abilities/visage_summon_familiars_oaa.txt
@@ -70,7 +70,7 @@
         "var_type"                                        "FIELD_INTEGER"
         "scepter_total_familiars"                         "3"
         "LinkedSpecialBonus"                              "special_bonus_unique_visage_6"
-        "scepter_reqired"                                 "1"
+        "RequiresScepter"                                 "1"
       }
     }
   }
--- a/game/scripts/npc/abilities/visage_summon_familiars_oaa.txt
+++ b/game/scripts/npc/abilities/visage_summon_familiars_oaa.txt
@@ -69,6 +69,7 @@
       {
         "var_type"                                        "FIELD_INTEGER"
         "scepter_total_familiars"                         "3"
+        "LinkedSpecialBonus"                              "special_bonus_unique_visage_6"
       }
     }
   }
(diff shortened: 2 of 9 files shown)


game/scripts/npc/abilities/visage_grave_chill.txt
```diff
@@ -48,7 +48,7 @@
       "03"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "attackspeed_bonus"                               "34 46 58 70 82 94"
+        "attackspeed_bonus"                               "36 48 60 72 96 120" //OAA
```
DarkoniusXNG, line 51: You can put similar values to Shiva's Guard. Keep in mind that bonus AS affects familiars too.

game/scripts/npc/abilities/visage_summon_familiars_oaa.txt
```diff
@@ -40,12 +40,12 @@
       "01"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "familiar_hp"                                     "500 700 900 1500 2100"
+        "familiar_hp"                                     "500 750 1000 1750 2500" //oaa
```
DarkoniusXNG, line 43: //oaa is not needed for custom abilities.

game/scripts/npc/abilities/visage_summon_familiars_oaa.txt
```diff
@@ -69,8 +69,6 @@
       {
         "var_type"                                        "FIELD_INTEGER"
         "scepter_total_familiars"                         "3"
-        "LinkedSpecialBonus"                              "special_bonus_unique_visage_6"
```
DarkoniusXNG, line 72: Don't remove this. Visage's scepter in OAA also gives another familiar. I know its not like in vanilla. Its on purpose this way.

game/scripts/npc/abilities/visage_summon_familiars_oaa.txt
```diff
@@ -56,7 +56,7 @@
       "04"
       {
         "var_type"                                        "FIELD_INTEGER"
-        "familiar_attack_damage"                          "25 50 75 150 300"
+        "familiar_attack_damage"                          "40 80 120 200 360" //oaa
```
DarkoniusXNG, line 59: Familiar damage isn't the issue afaik but it can't hurt. //oaa not needed for custom ability.

DarkoniusXNG (commented):
Random Idea: In soul assumption you can increase stack limit at OAA levels from 11/16 to 12/24.

game/scripts/npc/abilities/visage_summon_familiars_oaa.txt
lovesmuffins: So you are ok with dmg buff and the 4 familiars? Fine with me in that case.

game/scripts/npc/abilities/visage_grave_chill.txt
lovesmuffins: shivas values seems a little too good, Do you dislike these current numbers

game/scripts/npc/abilities/visage_summon_familiars_oaa.txt
DarkoniusXNG: Yeah its fine.

game/scripts/npc/abilities/visage_grave_chill.txt
DarkoniusXNG: I don't dislike them, it just doesn't seem impactful. It doesn't seem like it will change something about the hero.

lovesmuffins: Purpose of these changes are more or less more impactful at oaa levels to counteract shivas a bit better, early levels is just to look prettier and scale correctly with oaa levels.

lovesmuffins:
> Random Idea: In soul assumption you can increase stack limit at OAA levels from 11/16 to 12/24.

At max level you'd need 2400 dmg to be done to heroes before you get max charges. Do you consider that easily gainable? Or would you rather just incease dmg per charge at oaa levels.

DarkoniusXNG:
* Fixed Gravekeeper's Cloak not being disabled with break.
* Fixed Visage Shard not working.
* Gravekeeper's Cloak max number of layers reduced from 6/10 to 5/7. New level 25 talent increases this by 5 and minimum damage now scales at OAA levels. It doesn't seem fun if max number of layers with the talent is 15 while layer recovery time is 3 seconds.
* Updated tooltips for buffs and shard.

PR #3143, opened 2021-05-28
nerfs

lovesmuffins:
atrophy aura nerfs, changed to vanilla values than buffed a tad. OAA levels does not need big improvements since pct scale well anyway.

Cast point to dark rift for QOL improvement, doesn't change all that much.

Makes it so underlord ulti doesn't go through magic immune enemies and significantly reduced dmg (took values from magnus ult than buffed accordingly)

Pit of malice scales better despite vanilla nerfs, this is because I both changed duration to 10 and made the cd not scale. The oaa levels needed to scale more as a result so overall changes to ability was required. Intention is to not have people perma rooted which was possible before. If we're anti-perma stun we're anti-perma root.

Nerfed familiar damage slightly in the mid levels, which is where visage got a huge powerspike usually.


lovesmuffins:
actually cast point change isn't pointless, In a hypothetical semi-competitive enviroment, the .5 seconds allows an ally not in the area to save an ally with spells or force, or stun UL stopping him from casting it. Think shadow demon shadow poison sub-spell.

Realistically no spell that does a significant action should have no cast point (earthshaker is the exception)

DarkoniusXNG:
Merged into #3145

PR #3148, opened 2021-06-16
Spark economy rebalance

DarkoniusXNG:
* GPM spark added to passive GPM. Everyone has GPM spark now.
* Default spark is no longer GPM for each hero. Each hero has a different default spark.
* Other sparks don't give bonus gold from killing creeps anymore.
* XP spark reworked into Support spark: Increases passive experience gain by 25%. One-shot a non-ancient creep for bonus experience (125%) every time the counter reaches 100.
* Power Spark bonus damage on illusions and summons increased from 1/10 to 1/8.
* Added some new spark buffs. They are too snowbally to use at the moment but maybe we can incorporate them somewhere else in the future.


lovesmuffins:
perhaps we should have sparks that revolve around current game mechanics. Sparks can't be changed with my ideas in mind

Bonus dmg to boss spark, Less dmg taken by boss spark, bonus dmg in duel spark, damage resistance in duels spark, stuff like that.

lovesmuffins:
make midas a neutral item!

PR #3153, opened 2021-07-21
Core points currency system (WIP)

DarkoniusXNG:
* Killing bosses and taking capture point now gives 2/4/8/16 core points instead of upgrade cores 1/2/3/4. Because of this there is no auto-combine glitch anymore.
* Upgrade Cores are now purchasable with core points. Because of this inventory space is easier to manage.
* Upgrade Core Core Fission ability removed.
* Upgrade Cores gold value set to 1 gold. When sold they are fully refunded in core point value (because you can accidentally buy an upgrade core).
* Item selling will now give you half the core point value if the item has the core point value. For example, maxed-out items are worth 30 core points, when you sell them you gain 15 core points, even during ''sell grace time''. This means changing item builds is less punishing because you don't lose too many upgrade cores/core points.
* Azazel's items in the Azazel's Shop no longer require Upgrade Cores to buy. They now require core points instead. This means that it is easier to buy them.
* Buying tier 2 and higher upgrade cores is blocked until you or the enemy kills a corresponding tier boss or when both capture points of the corresponding tier are captured/taken.
* When you try to buy an upgrade core or other item that requires core points, an error message will pop out if you don't have enough core points or if you didn't take an objective.
* Aghanim's Blessing no longer requires Upgrade Cores to buy.
* Aghanim's Blessing can no longer be upgraded from level 1 Aghanim Scepter. (now only from Aghanim Scepter 5)
* Aghanim's Blessing recipe cost changed from 20000 gold to 1600 gold and 10 core points.
* Divine Rapier no longer requires Upgrade Cores to buy.
* Divine Rapier recipe cost changed from 50000 gold to 50000 gold and 40 core points.
* Items can now be bought just by right-clicking on them in the shop if you have enough resources.
* Wanderer gives 6/12/16 core points instead of Upgrade Cores 1+2/2+3/4.
* Couriers and Spirit Bear can buy items normally and they will combine on the unit that buys them.

TO DO:
1) Change how shrines work and change what snake shrine does to corepoint-gold conversion shrine.
2) Improve the look (css) of core point UI (currently it's empty and shows a red number) otherwise it is fully functional.
3) Icon or the background of the core point resource panel needs to change color (4 colors) based on the number of core points: if number < 4 then color1; if 4 <= number < 8 then color2; if 8 <= number < 16 then color3; if number >= 16 then color4;
4) Check if core point modifier is applied too early. Some heroes may not have it. Hero death fixes it probably.


chrisinajar:
`Change how shrines work and change what snake shrine does to corepoint-gold conversion shrine.` one way right? only core -> gold to maintain the ability to "sell" cores?

PR #3158, opened 2021-07-30
Spiked Mail update

DarkoniusXNG:
* Returned damage from both components is now the same damage type as the original/received damage.
* Spiked return damage now scales with levels: 80/85/90/95/100%
* Spiked return damage now stacks with passive damage return of Spiked Mail. While Spiked Mail is active damage returned is: 100/105/110/115/120%. (2 damage instances)
* Spiked Mail mana cost reduced from 100 to 25.
* Fixed Blade Mail passive stacking with Spiked Mail passive and active. Now only 20 constant damage from Blade Mail's passive stacks.
* Fixed Veil of Discord debuff amplifying damage from Spiked Mail (both passive and active).

### DIFF
--- a/game/resource/English/ability/items/tooltip_spiked_mail.txt
+++ b/game/resource/English/ability/items/tooltip_spiked_mail.txt
@@ -3,10 +3,10 @@
 //==========================================================================
 "DOTA_Tooltip_Ability_item_spiked_mail_1"                                "Spiked Mail"
 "DOTA_Tooltip_Ability_item_recipe_spiked_mail_1"                         "Spiked Mail Recipe"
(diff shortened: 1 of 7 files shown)


lovesmuffins:
from what I saw the damage values were insane, why buff it?

TheUnholyWrath:
> from what I saw the damage values were insane, why buff it?

values were insane because it was pure damage, it is not anymore

DarkoniusXNG:
I thought that was clear from the first sentence.

lovesmuffins:
what's different from blademail than

TheUnholyWrath:
spiked mail pierces BKB

PR #3159, opened 2021-08-03
New Heart

DarkoniusXNG:
Heart needs to be different from Heart Transplant and more offensive-oriented. I added an active ability to the heart that works like old Havoc Hammer's active but without the slow (knockback, AoE dmg based on strength). When activated it also buffs the caster with bonus strength and bonus base damage. Illusions inherit this buff too but for a limited time (not like Armlet).
* Regen Crystal: Added activation sound (Essence Ring cast sound).
* Regen Crystal: Added more distinguishable particle effect to the Regen Crystal buff (Essence Ring particle effect).
* Heart Transplant: Updated tooltip and code.
* Updated item builds and shops.


lovesmuffins:
What if regen crystal heal was based on current mana percentage and than the active would give like a 1000% mana regen or 5 seconds or something. Could be a neat way to differentiate it from heart

PR #3160, opened 2021-08-04
New Butterfly

DarkoniusXNG:
Added an active ability to Butterfly: Grants 0.08% movement speed for each point of Agility and 100% evasion for 3 seconds. Undispellable. Cooldown: 25s.


lovesmuffins:
100% evasion doesn't add really anything to the game since MKB will always counter evasion. What if the active had a chance to disarm attackers instead?

AHGIJMKLKKZNPJKQR:
Disarm doesn't add really anything to the game since halberd exists.

lovesmuffins:
> Disarm doesn't add really anything to the game since halberd exists.

a disarm on all potential attackers would provide way more than 100% evasion while still being dependent on the enemy attacking you in order to get disarmed

DarkoniusXNG:
100% evasion works against bosses.

lovesmuffins:
> 100% evasion works against bosses.

3 second 100% evasion isn't going to be any more effective against bosses than the 45% evasion it has normally

PR #3162, opened 2021-08-09
Blade of Judecca another update

DarkoniusXNG:
* Blade of Judecca changed damage type from Pure to Magical.
* Blade of Judecca damage can no longer spell lifesteal. (It's still affected by spell amp.)
* Blade of Judecca debuff cannot be applied with item damage anymore.
* Blade of Judecca debuff can no longer be applied with damage that has HP removal flag (Heartstopper Aura for example).

All you need to know: 
1) Rightclick-damage heroes without -magic resist or spell amp are better with Eye of Skadi. 
2) Casters and rightclick heroes with -magic resist or spell amp (Viper for example) are better with Judecca.
3) Judecca damage can now be mitigated with magic resistance and magic block.
4) Blademail and Spiked Mail actually return Judecca damage now without spell-lifesteal bs.
5) Radiance, Dagon, Spirit Vessel etc. no longer apply Judecca debuff.


chrisinajar:
aww, i liked radiance judecca

PR #3165, opened 2021-08-19
7.30 (WIP)

DarkoniusXNG:
* Finished items (some tooltips remaining)
* Fixed crashes caused by CreateModifierThinker when caster or ability is nil.
* Created custom thinker in those cases.


DarkoniusXNG:
Finished. 
1) I didn't do Meepo Shard because it's buggy as hell. So Meepo doesn't have a shard now. 
2) I also didn't bother to fully remake Beastmaster Shard in OAA. You still need to manually control the Hawk. But I hope I at least fixed the ''Hawk not dying bug''.
3) I ignored all Valve's undocumented nerfs. (nerfs that are not in the patch notes and that don't make sense)
4) I fixed most of the tooltip errors but there is probably more.
5) Updated some custom abilities.

PR #3181, opened 2021-09-15
Invoker Orb buffs

TheUnholyWrath:
they're now all double their vanilla values so the cutesy scaling is preserved.

if you think invoker having 6 free regen in first duel is too broken lemme know


lovesmuffins:
you have to double exort then

PR #3187, opened 2021-09-24
Wanderer timings changes

DarkoniusXNG:
* Wanderer respawn time increased from 3-6 minutes to 4-6 minutes.
* Wanderer buff duration reduced from 3 minutes to 2 minutes and 30 seconds.


DarkoniusXNG:
With this change minimum downtime of wanderer buff is 1 min and 30 seconds.

PR #3198, opened 2021-09-29
Octarine Buffs

lovesmuffins:
With the recent changes to how CDR works, it seems many players are no longer favoring Octarine. Octarine CDR doesn't stack with other CDR items or abilities and thus many players opt for something different. This used to scale even when it could be stacked with others and the change to 25 was when cast range was added, but now with the recent changes one can argue we could change it back.


DarkoniusXNG:
Uhm no

PR #3210, opened 2021-10-13
Luna buffs

TheUnholyWrath:
-Ult cd 120->100s, no hit_count limit without aghs
-25 Luna blessing talent 35 -> 70 damage

I was going to change the talent for eclipse cd but honestly I think its fine at 100s. Hit count without aghs gets increasingly stupid as it scales into oaa levels, I don't think it needs to exist but the case can be made for it scaling under beam count. Luna blessing is a shit ability and giving her 105 damage at level 25 seems fine (yes even with Visage Darkonius), not even good I don't think just fine.


lovesmuffins:
have your playerbase keep dying that'll certainly help

PR #3218, opened 2021-10-28
Arcane Blink rebalance and Rubick talent buff

TheUnholyWrath:
- Nerf cast point scaling on Arcane blink from 30-70% to 35-55%
- Buff Rubick stolen spells cd 25% -> 30%

I'm shocked 70% cast point reduction survived as long as it did. 

If you have a better idea for what to do with this talent I am open to anything. Its made pretty much completely pointless by not stacking with octarine, I was considering removing it entirely but this seems like a better option. Should be relevant without being overpowered as long as it doesn't stack with other sources of cooldown reduction.


lovesmuffins:
new idea for talent:

each soul gives 1% cdr

PR #3219, opened 2021-10-29
7.30e (OAA version)

TheUnholyWrath:
https://www.dota2.com/patches/7.30e

Idk if I had to do shards there were some of them with values but I couldn't see others. Missed lightning bolt stun duration, the non upgradable item changes idk where they are, probably some other stuff... Save me Darkonius


DarkoniusXNG:
Saved.

PR #3229, opened 2021-11-10
Hero Balance november 2021 - Chapter 1

DarkoniusXNG:


DarkoniusXNG:
I will do other heroes some other time. PR is becoming too big.

PR #3241, opened 2021-11-30
Ward stack update

DarkoniusXNG:
* Changed Ward Stack empty icon from Iron Branch to black-and-white ward dispenser icon.
* Ward Stack will now show an error when trying to use it while empty.
* Ward Stack will now generate wards when in the backpack, on the ground or when destroyed. Generating wards will start with the first Ward Stack purchase.


chrisinajar:
> Ward Stack will now generate wards when in the backpack, on the ground or when destroyed. Generating wards will start with the first Ward Stack purchase.

literally overpowered

PR #3248, opened 2021-12-06
Rebalance Fiery Soul

TheUnholyWrath:
-2 charges instead of 3
-same overall ms, nerfed as 

fiery soul is a stupid spell. bottles make it free as + ms. I wanted to nerf the as values but its hard with 3 stacks to not dumpster it. 2 stacks also makes it a bit stronger to compensate for being nerfed


lovesmuffins:
this is some of the oddest balancing I've seen in a long while here.

PR #3249, opened 2021-12-06
Nerf cleave spark

TheUnholyWrath:
0 thought went into the specifics gimme feedback 

-cleave spark is better than power spark
-cleave spark is better than power spark on illusion heroes
-i don't want to buff power spark


DarkoniusXNG:
When I nerfed from 50% to 35% I thought it would be enough. I did increase number of targets from 4 to 5, because 4 felt bad to me.

PR #3256, opened 2021-12-10
Final duel comeback fix for real this time

DarkoniusXNG:


game/scripts/vscripts/components/duels/final-duel.lua
```diff
@@ -124,7 +124,7 @@ function FinalDuel:EndDuelHandler (currentDuel)
 
   -- Give points to winners as a comeback compensation
   local playerCount = PlayerResource:SafeGetTeamPlayerCount()
-  local pointAward = math.ceil(playerCount / 20) * self.pointDifference
+  local pointAward = math.ceil(playerCount * self.pointDifference / 20)
```
chrisinajar, line 127: hahaha 1 times point difference lets go xD xD

Thinking about this though, why include player count? Including player count just increases the number of finals duels needed before coming back in lower player count games, really I think we just want something like `self.pointDifference / 4` so it takes 4 duels to come back, or something like that. I can't really think of a good reason to include player count here, especially in the nominator as opposed to the denominator, if anyhting I would expect the comeback mechanic to be stronger in 1v1's compared to 5v5's since there's lots of other ways to gain points proportional to player count so the more platers in the game the easier it is to come back today...

chrisinajar (approved):

PR #3258, opened 2021-12-10
Capture points rework

DarkoniusXNG:
Work in progress


DarkoniusXNG:
Let's merge this so I can see how it's in the game. It shouldn't feel worse than it already is.

PR #3260, opened 2021-12-16
Fixing Valve's latest new addition to modding community

DarkoniusXNG:
* Fixed panorama crashes
* Fixed surrender panel being wack.
* Changed required score difference for surrender from 50 to 35.
* Fixed Ogre Magi and Lycan materials. Idk why compiled files aren't ignored. They are in .gitignore and yet they are here. /shrug


chrisinajar:
if they were added to the repo before they were in the gitingore then they stay, we probably want `git rm --cached` the files :+1:

PR #3266, opened 2021-12-21
Top bar fix during duels

DarkoniusXNG:
Found the cause, coming up with the solution.


DarkoniusXNG:
* Fixed ''perma-dead'' top bar bug by disabling SetRespawnDisabled lines.
* Heroes now have a respawn time (equal to max duel timeout time) when they die during duels.
* Reduced Final Duel timeout from 5 min to 3 min.
* You are now able to ping enemies and target allies on the top bar.
* Updated .gitignore
* Minor improvements

PR #3276, opened 2022-01-04
Duel and capture point timings settings changes

DarkoniusXNG:
* Added CAPTURE_POINTS_AND_DUELS_NO_OVERLAP to settings.lua. If you want to try to overlap Duel and Capture point timings, set this to false.
* Changed duel timeout for every duel in 1v1 mode from 80/90/180 to 70 seconds.
* Changed duel interval in 1v1 mode from 480 seconds to 360 seconds.
* Changed initial capture point delay from 660 seconds (11 min) to **FIRST_DUEL_TIMEOUT + DUEL_INTERVAL + DUEL_TIMEOUT + 10** seconds (11 min normal, 8.5 min in 1v1 mode). Changing INITIAL_CAPTURE_POINT_DELAY or ONE_V_ONE_INITIAL_CAPTURE_POINT_DELAY will have no effect if CAPTURE_POINTS_AND_DUELS_NO_OVERLAP is true.
* Changed capture point interval from 480 seconds (8 min) to **DUEL_INTERVAL** (8 min normal, 6 min in 1v1 mode). Changing CAPTURE_INTERVAL or ONE_V_ONE_CAPTURE_INTERVAL will have no effect if CAPTURE_POINTS_AND_DUELS_NO_OVERLAP is true.

### DIFF
--- a/game/scripts/vscripts/components/capturepoints/capturepoints.lua
+++ b/game/scripts/vscripts/components/capturepoints/capturepoints.lua
@@ -59,8 +59,8 @@ function CapturePoints:Init ()
   self.currentCapture = nil
   self.NumCaptures = 0
 
-  CapturePoints.nextCaptureTime = INITIAL_CAPTURE_POINT_DELAY
-  HudTimer:At(INITIAL_CAPTURE_POINT_DELAY - 60, function ()
+  self.nextCaptureTime = self:GetInitialDelay()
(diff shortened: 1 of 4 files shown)


chrisinajar:
settings = fuel for modifier modes :D :D 

objective based mode could be low cooldown on duels and capture points with massively buffed rewards for each

PR #3279, opened 2022-01-06
Sacred Skull nerfs

DarkoniusXNG:
* Damage per missing hp percent reduced from 15/20/30/45 to 15/20/25/30.
Min damage reduced from 735/980/1470/2205 to 735/980/1225/1470.
Max damage reduced from 1485/1980/2970/4455 to 1485/1980/2475/2970.
* Damage no longer pierces spell immunity.


chrisinajar:
why do you hate bloodstone even more than me

PR #3285, opened 2022-01-09
Scale butterfly active duration

TheUnholyWrath:
I think it lasts too long at level 1.

This will make the tooltip a lot messier, very ready to close this at the slightest pushback

### DIFF
--- a/game/scripts/npc/items/item_butterfly.txt
+++ b/game/scripts/npc/items/item_butterfly.txt
@@ -98,7 +98,7 @@
       "07"
       {
         "var_type"                                        "FIELD_FLOAT"
-        "buff_duration"                                   "3.0"
+        "buff_duration"                                   "2.6 2.8 3.0 3.2 3.4"
       }
     }
 
--- a/game/scripts/npc/items/item_butterfly.txt
+++ b/game/scripts/npc/items/item_butterfly.txt
@@ -98,7 +98,7 @@
       "07"
       {
         "var_type"                                        "FIELD_FLOAT"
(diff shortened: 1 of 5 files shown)


lovesmuffins:
i'm ngl this scaling doesn't seem like it'll change anything but maybe I'm wrong.

TheUnholyWrath:
yeah I do not feel good about any of this. if this is still bad/meaningless I'm just gonna close it and leave it alone. butterfly is just in such a weird balance state where this one thing about it is pretty annoying and everything else doesn't seem overtuned, and as soon as you get an mkb that thing isn't even really a problem

lovesmuffins:
@TheUnholyWrath I know this is closed but I am wondering what your thoughts on the butterfly item w/o the active. Like is it in a good state, does the active make it OP, does the active make it just good, etc.

TheUnholyWrath:
without the active its really hard to balance/ ever make good. it was good historically because you could get a diffu and a core and rush level 2 bfly. things like me deleting the evasion and the 3 other stats is has being ones that are pretty controlled in oaa meant it was destined to be shit or op. with valve adding more effects to items this was pretty much the only direction to go with butterfly. the active has a lot of variance but can be quite powerful so I like it. in general it shouldn't be super overbearing since you can get mkb and counter it but I think its slightly too good against tier 1 bosses and the ancient camp, and before people can get mkb to counter it. seems like darko is messing around with giving bosses true strike in the other pr so i think once its not 3s of immunity against damage tanks at (what I think is) the most important time to take bosses, butterfly should be in a really good place

PR #3293, opened 2022-01-31
Leshrac Shard nerf

DarkoniusXNG:
Instructions for Chris: Find "particles/leshrac/leshrac_split_earth_aoe_indicator.vpcf" and compile it. Best way is to pick Leshrac in game, do -item shard, and use Q. If recompiling through the asset browser works then even better.


game/scripts/npc/abilities/leshrac_split_earth_oaa.txt
```diff
@@ -0,0 +1,82 @@
...
+      }
+      "05"
+      {
+        "var_type"                                        "FIELD_INTEGER"
+        "shard_extra_instances"                           "1"
+      }
+      "06"
+      {
+        "var_type"                                        "FIELD_INTEGER"
+        "shard_extra_radius_per_instance"                 "75"
```
lovesmuffins, line 73: I think the nerf should be the radius just doesn't increase and triggers only 3 extra instances. I genuinely think we're overrating this shard. I know it's good for capture points and such but still very useless for fights.

PR #3296, opened 2022-02-01
Add Blink Daggers to no rearm list

TheUnholyWrath:
Theres probably a better way of phrasing the tooltip. I also don't know how to test tooltips, should be red.


lovesmuffins:
tinker's bootiecheeks in OAA ain't he with the removal of march of machines?

PR #3299, opened 2022-02-05
Force minimap on the left and shop/gold on the right.

DarkoniusXNG:


lovesmuffins:
baumi will never play this game again XD

PR #3301, opened 2022-02-09
Splash Cannon rework

DarkoniusXNG:
* Splash Cannon now passively provides 100/150 damage.
* Splash Cannon now passively provides 40% splash in 250 radius. Doesn't work on melee heroes.
* Splash Cannon active reworked: **Cannon Fire** - Launches a projectile (cannon ball) at target location that damages and knocks back enemies in 500 radius upon landing. Damage is 700/1000 damage + 150% of your attack damage. Using Cannon Fire will cause a powerful recoil, pushing you backwards 250 units. 
* Notes:
1) Cast range: 1100; 
2) Projectile speed: 1800;
3) Knockback distance: 250; 
4) Knockback doesn't work against spell-immune and attack-immune enemies.
5) Mana cost: 200
6) Cooldown: 18 seconds.
7) Knockback duration is 0.5s and recoil duration is 0.2s
8) Active does work on melee heroes but the damage is reduced. It only applies 700/1000 physical damage.
9) Passive splash damage doesn't ignore enemy Damage Block. Cannon Fire damage ignores enemy Damage Block.

### DIFF
--- a/content/soundevents/items/oaa_items_sfx.vsndevts
+++ b/content/soundevents/items/oaa_items_sfx.vsndevts
@@ -16,5 +16,37 @@
     volume_fade_out = "0.600000"
     spread_radius = "300.000000"
   }
+  
+  Splash_Cannon.Launch = {
+    type = "dota_src1_3d"
+    volume = 1.000000
+    pitch_rand_min = -0.050000
+    pitch_rand_max = 0.050000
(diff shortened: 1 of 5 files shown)


lovesmuffins:
does the active use attack modifiers (like will tidebringer proc/time lock proc/slardar stun etc)?

PR #3304, opened 2022-02-10
Another modifier update

DarkoniusXNG:
* Added Rend modifier
* Echo Strike modifier now uses True Random instead of Pseudo Random.
* Echo Strike modifier now has 0.5s cooldown.
* Global and boss modifiers will now be removed from units after death if they can't respawn.
* Fixed icon for Blood Magic modifier.
* Fixed icon for Hyper Active modifier.
* Resolved Hyper Active modifer cdr conflict with Pro-Active modifier cdr.
* Added icon for Hyper Experience modifier.
* Added IsDebuff check to modifiers that didn't have it.


DarkoniusXNG:
Fixed diarrhetic and hyper experience being dispellable.

PR #3317, opened 2022-03-07
Revenants Brooch 

TheUnholyWrath:
Attacking through ethereal is a cool mecahnic, 5 attacks is shit. Theres still several knobs to adjust on it if its still shit or op now, should be a significantly better direction


lovesmuffins:
...

PR #3331, opened 2022-03-30
New Hero / Tinker Rework

DarkoniusXNG:
* Renamed Tinker to Tinkerer
* Replaced Heat-Seeking Missiles with Smart Missiles
* Replaced Laser with Tar Spill
* Replaced Rearm with Laser Contraption
* Defense Matrix buffed.
* All talents changed.

Tinkerer is similar to Chipper from Hon, so it should be a fun hero. I reworked Tinker because I think he is a lost cause. And also this hero seems better for the game. Numbers are probably overtuned, but we need to play-test the hero in the actual game to see if that is the case.


lovesmuffins:
Numbers are probably overtuned, but we need to play-test the hero in the actual game to see if that is the case.

this is my argument everytime i make a pr :D and darkonius says no

DarkoniusXNG:
Except this is a new hero and not related to your PRs.

warpdragon:
now now, ladies

PR #3333, opened 2022-03-31
Tinkerer buffs

DarkoniusXNG:
* Smart Missiles width increased from 100 to 115.
* Tar Spill attack speed slow rescaled from 15/20/25/30/35/40% to 10/20/30/40/45/50%.
* Tar Spill will now ignite on any spell or item damage, not just from Smart Missiles and not just from Tinkerer.
* Laser Contraption initial damage increased from 200/300/400/700/1200 to 200/400/600/1200/1800.
* Laser Contraption damage per second increased from 75/150/225/375/675 to 75/175/275/575/875.
* Level 10 Talent +40 Smart Missiles Damage increased to +50.
* Level 25 Talent Tar Spills magic resistance reduction improved from 25% to 45%.

### DIFF
--- a/game/resource/English/npc/tooltip_tinkerer.txt
+++ b/game/resource/English/npc/tooltip_tinkerer.txt
@@ -60,7 +60,7 @@
 "DOTA_Tooltip_ability_tinkerer_smart_missiles_stun_duration"                    "STUN DURATION:"
 
 "DOTA_Tooltip_ability_tinkerer_oil_spill"                                       "Tar Spill"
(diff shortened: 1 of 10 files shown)


DarkoniusXNG:
* Laser Contraption with scepter now also reduces healing by 25%.
* Added sound to Tar Spill.
* Smart Missiles: Improved collision radius. (Collision radius is bigger on launch. This means it's really hard to miss point-blank.)
* Smart Missiles: Improved direction of the projectile. (It's still weird sometimes if you click really near the Tinkerer because missiles come out from his missile launcher which is on his side.)
* Smart Missiles: Improved missile projectile when stolen by Rubick.
* Laser Contraption Keen Node collision radius reduced. This means that destroying 1 Keen Node is enough to leave the Laser Contraption.
* Fixed Smart Missiles doing 0 bonus damage to bosses instead of reduced damage.

PR #3334, opened 2022-04-05
Silver Staff tweaks

DarkoniusXNG:
* Cooldown reduced from 22 to 20 seconds.
* Cast Range increased from 900 to 950. 
* Mana Cost increased from 50 to 75.
* Bonus mana regen increased from 3/4 to 3.25/4.25
* Silver Staff will now do reduced damage to bosses instead of 0. (Only percentage dmg is reduced)

### DIFF
--- a/game/scripts/vscripts/items/silver_staff.lua
+++ b/game/scripts/vscripts/items/silver_staff.lua
@@ -142,7 +142,7 @@ function modifier_item_silver_staff_debuff:OnCreated()
 
     -- Do reduced damage to bosses
     if parent:IsOAABoss() then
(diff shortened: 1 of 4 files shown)


game/scripts/vscripts/items/silver_staff.lua
```diff
@@ -140,11 +140,13 @@ function modifier_item_silver_staff_debuff:OnCreated()
       ability = ability,
     }
 
-    -- Don't do damage to bosses
-    if not parent:IsOAABoss() then
-      ApplyDamage(damage_table)
+    -- Do reduced damage to bosses
+    if parent:IsOAABoss() then
```
chrisinajar, line 144: bosses already take reduced percentage damage, that's the point of boss resistance... this is double scaling.

PR #3337, opened 2022-04-12
Fix the shops in the dota arena

TheUnholyWrath:
-Fix dota arena shops
-Add gem to Sideshop


chrisinajar (approved):
@DarkoniusXNG squash merge if this is good to go, i gotta do some work stuff right up till the wire today, will ship whatever is merged when the time comes 🙇 🥇

PR #3364, opened 2022-05-30
Aeon Disk stat 'fix'

TheUnholyWrath:
my greatest hater npm test is telling me this is a moral and ethical violation of OAA law. both solutions I have break the law, is there some way that I can bribe test to shut up or is this the end of the road?

Item looks worse but functions better. Every level looks like it only has 1 level, it only shows the currents levels stats 1 time (it works if you change 400 400 400 400 400 to just 400 but I liked this better idk sue me). 

You can also scale the number of levels between the files but it still requires you to drop it and pick it up for it to work. Upgrading it could do that for you but its still really messy.


TheUnholyWrath:
you win this round test....

PR #3373, opened 2022-07-04
Shard Scaling

TheUnholyWrath:
The stupid discussion about how QOP is unrivalled at level 1 with shards was stupid. 

- Some shards that I think are mediocre now scale. I was too lazy to make files for Flamethrower or Fireball (fuck DK he doesn't need any more help). 
- Some shards that other people thought were OP now scale.
- Axe lost some armor. Permanent scaling is just a counter, I wanted to make it 2 at all levels but it seemed too strong early.
- Axe W change 
- Satanic and Pull Staff don't have crit damage scaling. I think we all hate Satanic. If Pull Staff is too bad then it can be buffed, but I don't think it is.
- Armlet attack speed buff. I need to offload all this armlet stock I bought.


game/scripts/npc/abilities/queenofpain_blink.txt
```diff
@@ -33,7 +33,7 @@
       "shard_damage" //OAA
       {
         "RequiresShard"				"1"
-        "value"						"250"
+        "value"						"100 150 200 250 500 750"
```
lovesmuffins, line 36: i feel like thats too much dmg at oaa levels. Like this is a buff for QOP outside of the stupid shard modifier, which shouldn't influence balancing eitherway.

PR #3374, opened 2022-07-06
Tiny Aghs buff

lovesmuffins:
happy birthday


lovesmuffins:
i know it does 200% dmg now .

PR #3377, opened 2022-07-10
Ta nerfs 

TheUnholyWrath:
- Base camps can only stack up to 6 creeps 
- Devastator has a slightly longer cooldown, costs more mana and costs 50 more gold.


lovesmuffins:
why not just nerf ta lol

PR #3386, opened 2022-07-31
Underlord nerfs

DarkoniusXNG:
Q: Underlord's only damage spell. In some cases it's very bad, I wanted to buff or rework this spell so many times. Like every dot damage spell in the game it's countered with Pipe or Heart or Regen Crystal or with just walking out of range. Max hp damage pct scales because it scales in vanilla as well, and trust me that's not the problem here with the Underlord.

W: I nerfed the duration of the pit. I would also like to nerf the scepter slow to not be 40% at all levels but I need to test if it's possible. Duration of the root itself is fine imo. Permaroot isn't possible because of the how the spell works (you are immune to all pits at least 1 second). Spell becomes bullshit with scepter because it slows you and you can't get out of the pit with just walking. Can be countered with any dispel, bkb, push or pull item/ability.
* Pit of Malice duration reduced from 12 to 7/8/9/10/11/12 seconds. (Pit duration looks like bullshit to me, area denial spell for 12 seconds.)

E: Atrophy Aura isn't OP. Here is why: At lvl 4 dmg reduction is the same as vanilla value at lvl 4. OAA lvls only improve this by 3. So max attack damage reduction is 53% with the talent. But that's only base damage reduction. In OAA physical damage dealers usually have more green (additional raw) damage than white (base) damage. It is strong against some bosses and illusion heroes but overall not as much as you think. Main reason why this looks op is because aghanim scepter gives a damage reduction buff to Underlord. In combination with Atrophy Aura it's ofc bonkers. If needed we can nerf the talent. Can be countered with buying raw dmg items, helm of the overlord, wraith pact, break items, bursting Underlord with spells...

Ultimate: I don't think Dark Portal (current vanilla ultimate) is good for OAA. It's an Io Relocate for the whole team without downsides. It gives damage reduction and movement speed for the whole team that passes the portal too. Dark Rift is also more fun to use than Dark Portal. If some of you think that current version of OAA Dark Rift isn't good for OAA, I will keep an open mind and listen to suggestions. In the meantime, I will nerf the bullshit aspects of the spell and make cast time slightly worse.
* Dark Rift cast range reduced from 1350/1400/1450/1500/1550 to 1400.
* Dark Rift cast point increased from 1.6 to 1.65 seconds. (Shadow Fiend Requiem cast time is 1.67 for reference)
* Dark Rift scepter cooldown increased from 80/75/70/65/60 to 90/85/80/75/70 seconds. (I can remove this part of scepter entirely if needed.)
* Dark Rift scepter damage reduction decreased from 10/20/30/35/40% to 10/15/20/25/30%. (Dark Portal buff damage reduction is 30% at max level.)
* Dark Rift scepter bonus movement speed reduced from 10/20/30/40/50% to 10/15/20/25/30%. (Dark Portal buff move speed bonus is 30% at max level.)
* Dark Rift scepter buff duration rescaled from 6 to 5/5.5/6/6.5/7 seconds. (Dark Portal buff duration is 7 seconds at max level.)


lovesmuffins:
I just do not think his current ult fits his character at all. His vanilla ult may be more "powerful" in a sense but it'd be far less annoying considering all of his other abilities are annoying to deal with.

PR #3396, opened 2022-08-10
Solar Crest Buff + rapier/magic lamp Buffs

lovesmuffins:
Solar Crest buffs because nobody fucking buys the item and I want to again provide alternative item choices. Rapier and Lamp buffed because again... nobody buys those items and chaos is fun imho


DarkoniusXNG (changes requested):
* Magic Lamp buff is fine.
* Pipe nerf - I don't think it's needed. I nerfed it when it was bonkers. You probably noticed that's it worse than vanilla during early levels. We like scaling, so don't remove scaling.
* Rapier buff - instead of buffing core point cost, buff the gold cost. I don't think it's needed but whatever.
* Shivas nerf - revert this, Shivas needs rework to have separate values for ranged and melee units.
* Solar Crest buff - either buff to be the same as Shivas or dont change it at all. I don't like touching bad vanilla items before the patch either.

lovesmuffins:
final changes pog!

PR #3412, opened 2022-09-15
Magic Lamp v2.0 buffs

lovesmuffins:
Me and darkonius glorious attempt to make this item worthwhile.

Last buff, if this fails we remove it :(


DarkoniusXNG:
I will try to rework the item.

PR #3453, opened 2023-01-11
Disable bottle drop after 8 min

DarkoniusXNG:
Creeps that spawn after 8 min will no longer have a chance to drop a bottle.


chrisinajar:
never

DarkoniusXNG:
Bruh

PR #3474, opened 2023-03-09
Bigmode

warpdragon:
Welcome to Big Mode.

Intended features:
- Bigger map
- Organic map
- Intelligent map


DarkoniusXNG:
I will check out what you did over the weekend. 
If you need suggestions for how many Arenas, Bosspits and Camps we need, I can tell you that:
* Number of arenas should be the same: 6 (2 small, and 4 big). Oaa seasonal map doesn't have 8! Bottom left and bottom right 'arenas' are unused empty space. One more thing: I am not sure but the code will probably return an error if there is less than 3 arenas (2 small and 1 big). So if your new map needs to have less than 6, make at least 4 (2 small and 2 big) and if there is an error consult me.
* Number of bosspits should also be the same: 10. I don't know how you got 8. Each side has 2 in base and 1 in each corner ((2+2)x2 = 8). And 2 middle bosses. We experimented in the past with less bosses, it resulted in 'snowballier' games.
* Number of camps can be the same or less. For example 1 of each type on each side of the map and 2 of each type in various places. Camp types are 'small', 'medium', 'hard', 'ancient', 'corner ancient dragon' and 'solo ancient'. So at least 36 camps, we can modify this later.

If your map doesn't have 'map border entities/triggers' you will need to edit this line: https://github.com/OpenAngelArena/oaa/blob/master/game/scripts/vscripts/components/zonecontrol/map.lua#L5

For offside zones if there is an error, don't worry about it.

warpdragon:
I know it won't merge like this, but I'll leave the PR open. It's only 2 files. I'll redo the PR when it's ready.

PR #3602, opened 2024-01-03
Tiny mode map spawns closer to center. map offsides closer to center. arena spa…

warpdragon:
…wns placed further apart. minor prop changes.


warpdragon:
^Changes to tiny mode.

PR #3614, opened 2024-02-11
Marci arcana

warpdragon:
Model and animations imported into mod. Needs implementation and testing. I have no idea if it works or not as I ran into a lot of weird issues in Modeldoc.


DarkoniusXNG:
I will test it out when I finish with bug fixing and if Valve doesn't release 7.35c in the meantime.

DarkoniusXNG:
Just letting you know that I didn't forgot about this. I just don't have a clear idea how to implement this.

PR #3732, opened 2024-10-03
Alt fixes

warpdragon:


warpdragon:
map edge control "fixed". tree traps "fixed". terrain work, plants. ship it.

PR #3747, opened 2024-11-25
Chatterjee rework 2024

DarkoniusXNG:
Nerfed across the board
Innate reworked
Facets reworked
Electric Shield reworked


DarkoniusXNG:
1 - picking screen
2 - facets
3 - electric shield dc facet
4 - electric shield ac facet
5 - electric shield nc facet
6 - talents
7 - shard and scepter

PR #3792, opened 2025-03-04
Map fixes March 2025

warpdragon:
Compiled a list of reports from community discord and corrected the stuff.


warpdragon:
Mergable PR. Seasonal, Alternate, Bigmode, and  Tinymode should be rebuilt/compiled for changes to take effect.

